PR #11, opened 2016-02-17
Fix hardcoded thrift path

juanmaneo:
Hi,

I'm starting to play around with the code at McKay-Brothers.
Just had to fix the hardcoded thrift path in gradle build file.

Can this be changed, thrift shall be in the PATH anyway, no ?


hbs:
Thanks for the PR

PR #12, opened 2016-02-22
Fix GTSHelper.quicksortByLocation

juanmaneo:
Hi,

Is GTSHelper.quicksortByLocation still used it doesn't seem to be called ?


hbs:
Thanks for the PR.

It's not (yet) called :-)

PR #16, opened 2016-02-26
Add ->ITERABLE, ITERABLE-> and CONCAT operations

juanmaneo:
- allow to handle iterable in WarpScript
   by adding ->ITERABLE (TOITERABLE) and ITERABLE-> (ITERABLETO) conversion operations.
 - add CONCAT function that concatenate in constant time two ITERABLE converted list.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/WarpScriptLib.java
+++ b/warp10/src/main/java/io/warp10/script/WarpScriptLib.java
@@ -305,6 +305,9 @@ public class WarpScriptLib {
     functions.put("STU", new STU("STU"));
     functions.put("UNIXTIMEEND", new UNIXTIMEEND("UNIXTIMEEND"));
     functions.put("UNIXTIMEALIGN", new UNIXTIMEALIGN("UNIXTIMEALIGN"));
+    functions.put("CONCAT", new CONCAT("CONCAT"));
+    functions.put("->ITERABLE", new TOITERABLE("->ITERABLE"));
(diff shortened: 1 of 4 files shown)


hbs:
What is the purpose of ->ITERABLE? simply casting the input to an Iterable has no effect whatsoever since it can only be done if the input implements Iterable

juanmaneo:
Yes you are right a List implement Iterable.

It's only for "symmetry" with ITERABLE-> only a syntactic sugar.

I use it like that:
[ 1 ] ->ITERABLE
[ 2 3 4 5 6 ] ->ITERABLE
CONCAT 
ITERABLE->

to have a stack with:
[1,2,3,4,5,6]

Of course it only moves the copy to the end in the ITERABLE-> operation.

juanmaneo:
A new Iterable type will only make sense if other functions on List type are also able have there counter part for Iterable ...
Removing the need to transform back to an ArrayList and a copy of the elements.

hbs:
The semantics of Iterables are kind of misleading since they only manipulate references to the underlying collections, so doing a CONCAT on two lists will produce an iterable but later modifying one of those two lists will also modify the iterable.

I suggest this functionnality be added through a WarpScriptRawJavaFunction rather than as a built-in in WarpScript.

juanmaneo:
Thanks, I will look into this then.
Do I close/abort this pull request ?

hbs:
I'm closing it as won't fix. Thanks for your time.

PR #19, opened 2016-03-01
Fixing bug while reading configuration

starsway:
System configuration was read instead of the Warp configuration
when assessing needed support for foreign languages in WarpScript.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/WarpScriptLib.java
+++ b/warp10/src/main/java/io/warp10/script/WarpScriptLib.java
@@ -203,6 +203,7 @@ import io.warp10.script.unary.TOLONG;
 import io.warp10.script.unary.TOSTRING;
 import io.warp10.script.unary.TOTIMESTAMP;
 import io.warp10.script.unary.UNIT;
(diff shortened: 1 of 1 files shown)


starsway:
By the way, a fully working example of a call to e.g. 'PYTHON' would be most welcome - we could not manage to make it work despite this minor fix.
Thanks

hbs:
Using third party languages is actually not encouraged as those languages cannot be secured and thus could lead to security issues (networking access, memory allocations, local disk access, cryptographic keys leakage, etc).

Please open an issue in cityzendata/www.warp10.io so we document 'PYTHON' with a higher priority.

starsway:
Thank you for these explanations.

I understand that the suggested fix is not acceptable performance-wise.
However, I confirm that calls to 'PYTHON' function resulted into an error message complaining about this function being unknown before this fix was applied...

Regarding the documentation request : understood, I'll create a separate issue.

hbs:
I closed this PR, a similar change appears in commit b62c59f.

PR #40, opened 2016-04-04
Added fine grained metrics to track the number of points deleted per owner+application

swallez:
For accounting purposes, I need to have a fine grained visibility on the number of points deleted by a particular user. This change add the write token to Kafka delete messages so that we can report in Sensision the count of deleted points associated to the owner (and app if present)

### DIFF
--- a/warp10/src/main/thrift/io_warp10_continuum_store_thrift_data.thrift
+++ b/warp10/src/main/thrift/io_warp10_continuum_store_thrift_data.thrift
@@ -1,5 +1,6 @@
-namespace java io.warp10.continuum.store.thrift.data
+include "../../../../token/src/main/thrift/io_warp10_quasar_token_thrift_data.thrift"
 
+namespace java io.warp10.continuum.store.thrift.data
 /**
  * Metadata describing a Geo Time Serie
  */ 
@@ -95,6 +96,11 @@ struct KafkaDataMessage {
    * Minimum age of cells to delete (in ms)
    */
   8: optional i64 deletionMinAge,
+
+  /**
+   * Optional write token to log metrics in deletion requests
+   */
+  9: optional io_warp10_quasar_token_thrift_data.WriteToken writeToken,
 }
 
 /**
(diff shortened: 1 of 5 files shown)


warp10/src/main/thrift/io_warp10_continuum_store_thrift_data.thrift
```diff
@@ -95,6 +96,11 @@ struct KafkaDataMessage {
    * Minimum age of cells to delete (in ms)
    */
   8: optional i64 deletionMinAge,
+
+  /**
+   * Optional write token to log metrics in deletion requests
+   */
+  9: optional io_warp10_quasar_token_thrift_data.WriteToken writeToken,
```
hbs: Instead of adding a field with the write token, add a field with the Metadata.

swallez: WDYM with "add a field"? A new property in Metadata (seems a heavy change) or using the "attributes" map?

hbs:
Nope, I meant in the KafkaDataMessage structure, instead of adding the writeToken, add the Metadata, they contain all the information you need.

swallez:
You mean the Metadata objects returned by `directoryClient.iterator()` contain `Constants.OWNER_LABEL` and `Constants.APPLICATION_LABEL` in their labels?
Then no change is needed in Ingress nor the thrift schema, right?

hbs:
The Metadata object is not part of the KafkaDataMessage used for deletion, so you need to add a field which contains it so it will be available when deletion is actually performed. Then indeed the labels (including owner and application) will be available for you to update the deletion metrics.

Given the additional space the Metadata take, also add a configuration parameter so Metadata can be omitted from the KafkaDataMessage if fine grained deletion statistics are not needed.

swallez:
Ok, got it. Now since not all labels are needed by the delete handler, is it ok that the Metadata added to KafkaDataMessage only contains owner and application labels in the case of DELETE messages, or do you prefer that the whole unmodified metadata are sent?

hbs:
It's better to have the whole unmodified Metadata instance.

swallez:
Closing this PR, will create a new one with a clean commit history (my git-fu is not good enough, messed things up when cleaning history)

PR #43, opened 2016-04-07
Added detailed deletion metrics

swallez:
For accounting purposes, I need to have a fine grained visibility on the number of points deleted by a particular user. This change reports in Sensision the count of deleted points associated to the owner (and app if present).
This is optional (and disabled by default) for the HBase version, and always active for the standalone version.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
+++ b/warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
@@ -65,8 +65,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.UUID;
-import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.LockSupport;
@@ -250,6 +248,8 @@ public class Ingress extends AbstractHandler implements Runnable {
   private final byte[] aesDataKey;
   private final long[] siphashDataKey;
 
+  private final boolean sendMetadataOnDelete;
+
   private final KafkaSynchronizedConsumerPool  pool;
   
   public Ingress(KeyStore keystore, Properties props) {
@@ -309,7 +309,9 @@ public class Ingress extends AbstractHandler implements Runnable {
     
     this.aesDataKey = this.keystore.getKey(KeyStore.AES_KAFKA_DATA);
     this.siphashDataKey = SipHashInline.getKey(this.keystore.getKey(KeyStore.SIPHASH_KAFKA_DATA));    
-    
+
+    this.sendMetadataOnDelete = Boolean.parseBoolean(props.getProperty(Configuration.INGRESS_DELETE_METADATA_INCLUDE, "true"));
+
     //
     // Prepare meta, data and delete producers
     //
@@ -1463,6 +1465,10 @@ public class Ingress extends AbstractHandler implements Runnable {
       msg.setDeletionMinAge(minage);
       msg.setClassId(metadata.getClassId());
       msg.setLabelsId(metadata.getLabelsId());
+
+      if (this.sendMetadataOnDelete) {
+        msg.setMetadata(metadata);
+      }
 
       sendDataMessage(msg);
     } else {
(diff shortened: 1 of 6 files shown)


warp10/src/main/java/io/warp10/continuum/store/Store.java
```diff
@@ -16,12 +16,14 @@
 
 package io.warp10.continuum.store;
 
+import com.sun.xml.bind.v2.runtime.reflect.opt.Const;
```
hbs: This should not be there I suppose!

warp10/src/main/java/io/warp10/continuum/store/Store.java
```diff
@@ -1011,6 +1013,25 @@ public BulkDeleteResponse call(BulkDeleteService service) throws IOException {
       Sensision.update(SensisionConstants.SENSISION_CLASS_CONTINUUM_STORE_HBASE_DELETE_OPS, Sensision.EMPTY_LABELS, 1);
       Sensision.update(SensisionConstants.SENSISION_CLASS_CONTINUUM_STORE_HBASE_DELETE_REGIONS, Sensision.EMPTY_LABELS, noOfRegions);
       Sensision.update(SensisionConstants.SENSISION_CLASS_CONTINUUM_STORE_HBASE_DELETE_DATAPOINTS, Sensision.EMPTY_LABELS, noOfDeletedVersions);
+
+      Metadata meta = msg.getMetadata();
+      if (null != meta) {
+        Map<String, String> labels = new HashMap<>();
+
+        String owner = meta.getLabels().get(Constants.OWNER_LABEL);
```
hbs: owner and application cannot be absent from metadata, so these tests are superfluous and can be removed.

warp10/src/main/java/io/warp10/standalone/StandaloneDeleteHandler.java
```diff
@@ -288,6 +290,23 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
         metas.append(sb);
         metas.append("\n");
         gts++;
+
+        // Log detailed metrics for this GTS owner and app
+        Map<String, String> labels = new HashMap<>();
+
+        String metaOwner = metadata.getLabels().get(Constants.OWNER_LABEL);
```
hbs: Same here, owner and application are always set so those tests can be removed.

warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
```diff
@@ -309,7 +309,9 @@ public Ingress(KeyStore keystore, Properties props) {
     
     this.aesDataKey = this.keystore.getKey(KeyStore.AES_KAFKA_DATA);
     this.siphashDataKey = SipHashInline.getKey(this.keystore.getKey(KeyStore.SIPHASH_KAFKA_DATA));    
-    
+
+    this.sendMetadataOnDelete = Boolean.parseBoolean(props.getProperty(Configuration.INGRESS_DELETE_METADATA_INCLUDE, "true"));
```
hbs: I'd rather have this set to false by default since it adds an overhead in the Kafka message being sent.

warp10/src/main/java/io/warp10/continuum/store/Store.java
swallez: Good to know, fixed!

warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
swallez: Doh! Obviously...

warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
```diff
@@ -316,6 +316,11 @@
   public static final String SENSISION_CLASS_CONTINUUM_STANDALONE_DELETE_DATAPOINTS = "warp.standalone.delete.datapoints";
 
   /**
+   * Number of datapoints deleted by 'standalone' delete per owner and application
+   */
+  public static final String SENSISION_CLASS_CONTINUUM_STANDALONE_DELETE_DATAPOINTS_PEROWNER = "warp.standalone.delete.datapoints.perowner";
```
hbs: I've been having second thoughts, it's probably better to have a perowner AND a perapp set of classes rather than a combo app/owner, this would be more in sync with the rest of the exposed metrics.

Do you need the combo?

swallez: I actually just need the owner, but added app for completeness. I'll make it a separate metric tomorrow :-)

hbs: Then simply rename the metric to 'perownerapp', You can then REDUCE by 'owner' or 'app'.
If cardinality ever becomes a problem we can always add a TTL to the metric.

swallez: Done!

hbs:
Thanks for the PR

PR #50, opened 2016-04-29
Adds Hbase filter task and publication

slambour:
The hbase filter jar contains only 'org/apache/hadoop/hbase/filter/**'  classes from warp10 without any dependencies.


hbs:
I merge it as is, but I wonder if publications should not be done as a simple jar rather than as a maven artefact which no project will ever use.

PR #54, opened 2016-05-04
Add support for 'hbase.zookeeper.property.clientport' hbase option.

julien-girard:
Add support for 'hbase.zookeeper.property.clientport' hbase option.


hbs:
Thanks for the PR.

PR #56, opened 2016-05-30
move configuration templates to a dedicated directory

slambour:
- template moved into a dedicated directory
- hard path into standalone.path variable
- configuration tag for standalone version upgrade with docker

### DIFF
--- a/etc/conf-standalone.template
+++ b/etc/conf-standalone.template
@@ -1,7 +1,7 @@
 /////////////////////////////////////////////////////////////////////////////////////////
 //
 // S T A N D A L O N E
-//
+// REVISION_TAG=1.0
 /////////////////////////////////////////////////////////////////////////////////////////
 
 //
@@ -93,7 +93,7 @@ warpscript.maxloop.hard = 10000
 //
 // Macro Repository
 //
-warpscript.repository.directory = /opt/warp10-@VERSION@/macros
+warpscript.repository.directory = ${standalone.home}/macros
 warpscript.repository.refresh = 60000
 
 //
@@ -137,7 +137,7 @@ warp.timeunits = us
 //
 // Path of the 'bootstrap' warpscript code for Egress
 //
-warpscript.bootstrap.path = /opt/warp10-@VERSION@/etc/bootstrap/egress.mc2
+warpscript.bootstrap.path = ${standalone.home}/etc/bootstrap/egress.mc2
 
 //
 // How often to reload the bootstrap code (in ms) for Egress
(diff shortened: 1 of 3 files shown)


etc/conf-standalone.template
```diff
@@ -216,6 +216,11 @@ standalone.host = 127.0.0.1
 standalone.port = 8080
 
 //
+// Directory of Warp10 standalone install
```
hbs: It would be better to move this before the actual uses of 'standalone.home'

warp10/src/main/sh/warp10-standalone.bootstrap
```diff
@@ -36,4 +37,4 @@ chown -R warp10:warp10 ${WARP10_HOME}
 
 # Generate the configuration file with Worf
 # Generate read/write tokens valid for a period of 100 years. We use 'io.warp10.bootstrap' as application name.
-su warp10 -c "${JAVA_HOME}/bin/java -cp ${WARP10_HOME}/bin/warp10-@VERSION@.jar io.warp10.worf.Worf -a io.warp10.bootstrap -puidg -t -ttl 3153600000000 ${WARP10_HOME}/etc/conf-standalone.template" >> ${WARP10_HOME}/etc/initial.tokens
+su warp10 -c "${JAVA_HOME}/bin/java -cp ${WARP10_HOME}/bin/warp10-@VERSION@.jar io.warp10.worf.Worf -a io.warp10.bootstrap -puidg -t -ttl 3153600000000 ${WARP10_HOME}/templates/conf-standalone.template -o ${WARP10_HOME}/etc/conf-standalone.conf" >> ${WARP10_HOME}/etc/initial.tokens
```
hbs: What is the '-o' for?

etc/conf-standalone.template
```diff
@@ -1,7 +1,7 @@
 /////////////////////////////////////////////////////////////////////////////////////////
 //
 // S T A N D A L O N E
-//
+// REVISION_TAG=1.0
```
hbs: What is this for?

warp10/src/main/sh/warp10-standalone.bootstrap
slambour: -o is the outpout file when written configuration file in not in the same directory than the template

etc/conf-standalone.template
slambour: done, I hesitate between before the first use and keep standalone.xxx grouped

etc/conf-standalone.template
slambour: manage configuration upgrades with docker -> put the rigth information on the container startups when the actual configuration file saved in the volume dosent match with the latest version of warp10 platform

warp10/src/main/sh/warp10-standalone.bootstrap
hbs: Seems weird to have an option AFTER the name of the input file which is not an option parameter.

-i <INPUT_FILE> would be more understandable I think, maybe a TODO for Worf

PR #80, opened 2016-08-22
Change agg AND and OR managing null values and locations

aurrelhebert:


warp10/src/main/java/io/warp10/script/aggregator/And.java
```diff
@@ -60,8 +64,11 @@ public Object apply(Object[] args) throws WarpScriptException {
         continue;
       } else {
         and = and && Boolean.TRUE.equals(values[i]);
-        if (!and) {
-          break;
+        
+        if (ticks[i] > timestamp) {
```
hbs: What happens if the GTS has duplicate timestamps? What happen if one timestamp is MIN_VALUE?

PR #109, opened 2016-11-14
Add functions MD5 SHA1 SHA1HMAC SHA256 SHA256HMAC

slambour:
Adds cryptographic hashes and hashmac.
The code of MD5 SHA1 & SHA256 is not factorized.
It could be with a generic function DIGEST with the Digest instance given in the constructor.
by ex : functions.put("SHA1", new DIGEST("SHA1", new SHA1Digest()));


hbs:
Can't you factor those functions already?

But don't create DIGEST with an instance of Digest, instead pass it a class and in 'apply' create a new instance, otherwise you would be sharing the same Digest instance between all current executions of the DIGEST functions.

so DIGEST(String name, Class<? extends GeneralDigest>)

warp10/src/main/java/io/warp10/script/functions/DIGEST.java
```diff
@@ -0,0 +1,61 @@
...
+
+      digest.update(bytes, 0, bytes.length);
+
+      digest.doFinal(digestOctets, 0);
+
+      stack.push(digestOctets);
+
+      return stack;
+    } catch (Exception exp) {
+      throw new WarpScriptException("Unable to instantiate digest " + getName());
```
hbs: Can you change the WarpScriptException to start with getName() so as to retain the overall structure.

throw new WarpScriptException(getName() + " unable to instantiate digest").

Same for HMAC.

PR #125, opened 2016-11-28
Made Timezone optional for ADDDAYS, ADDMONTHS, ADDYEARS while modifying timestamp

mehranshakeri:
Based on documentation timezone parameter should be optional for timestamp input but else-if condition in code makes it mandatory. With this pull request those conditions expect also Long on top of stack as timestamp.


mehranshakeri:
ADDMONTHS documentation example also includes wrong function (ADDDAYS).

http://www.warp10.io/reference/functions/function_ADDMONTHS/#example

hbs:
Thanks for the PR.

The documentation has been corrected also.

PR #126, opened 2016-11-30
Bugfix: FROMTSELEMENTS expects a LIST not LONG

mehranshakeri:
Fixed else if condition to expect a LIST
Changed exception message to include expected arg types


hbs:
Thanks for the PR

PR #127, opened 2016-12-07
Fixed standalone config template

mehranshakeri:
Fixed typo for parameters:
#datalog.forwarder.deleteforwarded = true
#datalog.forwarder.deleteignored = true


hbs:
Thanks for the fix Mehran.

PR #129, opened 2016-12-08
Bugfix in REPLACE and REPLACEALL

mehranshakeri:
REPLACE and REPLACEALL were called with wrong condition


hbs:
Thanks for the fix.

PR #142, opened 2017-01-10
Redefined protection

hbs:


aurrelhebert:
In your opinion wouldn't in be better if this new REDEFS function operate on an authenticate stack, As other functions that impact of the configuration ?

hbs:
Nope, as a stack can only be authenticated once and as it would force to use AUTHENTICATE in secure scripts thus including expiring tokens in them.

PR #157, opened 2017-02-18
Added labels to Worf CLI arguments

StevenLeRoux:
Allow to pass labels as arguments to create WRITE tokens that carry fixed labels inside the token.

### DIFF
--- a/warp10/src/main/java/io/warp10/worf/WorfCLI.java
+++ b/warp10/src/main/java/io/warp10/worf/WorfCLI.java
@@ -17,6 +17,7 @@ package io.warp10.worf;
 
 import com.google.common.base.Strings;
 import io.warp10.Revision;
+import io.warp10.continuum.gts.GTSHelper;
 import org.apache.commons.cli.*;
 import org.codehaus.jettison.json.JSONObject;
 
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/worf/WorfCLI.java
```diff
@@ -161,6 +168,21 @@ public int execute(String[] args) {
           appName = cmd.getOptionValue(APPNAME);
         }
 
+        if (cmd.hasOption(LABELS)) try {
```
hbs: Systematically use { } after an if

warp10/src/main/java/io/warp10/worf/WorfCLI.java
```diff
@@ -161,6 +168,21 @@ public int execute(String[] args) {
...
+            lbs = cmd.getOptionValue(LABELS);
+            String[] array = lbs.split(" ");
+            if (array.length % 2 != 0) {
+                throw new WorfException("Odd number of values provided as labels ");
+            }
+            for (int i = 0; i <= array.length - 1; i = +2) {
```
hbs: use i < array.length

warp10/src/main/java/io/warp10/worf/WorfCLI.java
```diff
@@ -101,9 +104,12 @@ public int execute(String[] args) {
       String producerUID = null;
```
hbs: use two spaces as indent

warp10/src/main/java/io/warp10/worf/WorfCLI.java
slambour: Infinite loop when labels > 2

slambour: Uses the ingress labels input format for Worf (xbeeId=XBee_40670F0D,moteId=53)
Spaces are allowed has value

StevenLeRoux:
PR updated using ingress input format

d33d33:
Ping

PR #175, opened 2017-04-25
feat: Allow application limit without producer one

d33d33:


d33d33:
I am working on a multi-tenant application where throttling is set by app (aka: not by user). Working at the app level is much more easy and clear for me than using the producer indirection.

hbs:
I still don't understand your concern, you don't want to have per producer limits at all?

d33d33:
Your right, i don't need those ones at all :)
Why producer limits are so important?

hbs:
There is a default producer limit which applies to producers which do not have a specific limit set in the throttling file.

You can use the special value -1 to indicate that no producer throttling should be done by default.

As it turns out there seems to be a bug in the ThrottlingManager code which will prevent the per application limit to be enforced when producer limit is set to -1. I'll fix that and the result should fit your needs.

d33d33:
Yes -1 is the good way to remove producer throttling.

It is the purpose of this PR to fix this bug no ? :)

hbs:
You had marked your PR with 'feat', but it was a bugfix. I'm closing it, I created issue #176 with an associated fixing commit (a060a71), please check that for your use case.

PR #184, opened 2017-05-12
Split directory metadatas by any label

d33d33:
Directory split classnames by `.owner` label. This work well on multi-tenant instance but is useless on instance with only one producer.

This PR add a configuration to choose on which label perform the split.


hbs:
What is the problem this PR solves?

d33d33:
This PR solve performances problem on find - as all the series are under the same owner.
Applying this patch allow us to reduce the query time from 33s to 4ms ;)

hbs:
What was the particular modeling? How many GTS, how many classes? I'm worried this PR will only solve a very specific problem and not offer a generic framework.

d33d33:
We have 20M GTS on 29 classes all under the same `.owner`

hbs:
Closing this PR as this type of indexing is specific to a use case and should be handled using a DirectoryPlugin.

PR #185, opened 2017-05-15
Undisclose embedded token labels in TOKENINFO

StevenLeRoux:
This patch adds tokens'extra labels in the output of TOKENINFO.


warp10/src/main/java/io/warp10/script/functions/TOKENINFO.java
```diff
@@ -76,7 +79,9 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
         if (wtoken.getLabelsSize() > 0) {
           tokenParams.put("labels", wtoken.getLabels());
         }
-        
+        if (wtoken.getLabelsSize() > 0) {
```
hbs: This code is not needed!

PR #191, opened 2017-05-29
Fixes #180 - Set a more meaningful class name in REDUCE

### LINKED ISSUE #180: Set a more meaningful class name in 'REDUCE'.
Right now we set the empty string as the name of the result GTS. It would be better to retain the original GTS name if it is common to all GTS we reduce.

PierreZ:
This PR is fixing [Issues 180](https://github.com/cityzendata/warp10-platform/issues/180). Now during REDUCE, classname is retained if identical across all GTS.


hbs:
Thanks for the PR, this was fixed independently.

PR #195, opened 2017-07-19
Fixes #194 - Add comments to TRL

### LINKED ISSUE #194: Add comments to TRL
Right now, we can put two types of format in the TRL:

* a token identifier
* an application name

One feature present in throttling files is comments, which allows us to add additional information next to a throttling. Can we do the same things for TRL? This would allow us to display additional information next to token identifiers like this:

```
# app.abcd
awes0mesiphash1
awes0mesiphash2
awes0mesiphash3
```

PierreZ:


hbs:
Thanks for the PR.

PR #200, opened 2017-08-04
Initial commit of last activity support

hbs:


StevenLeRoux:
Initials tests are looking good for now.

Adding "last activity" support on DELETE too would avoid a FOREACH cycle to META.

We continue testing with more loads now.

hbs:
The reason why _activeAfter_ / _quietAfter_ support is not present for neither DELETE nor the /delete endpoint is because the GTS selection done using those attributes cannot guarantee that the returned GTS were indeed active/quiet after the given timestamps. So to avoid potential deletion of GTS which do not match the given criteria support for deletion has not been implemented.

I understand this could still be a useful feature for those willing to accept the uncertainty of the GTS selection process, so what we could implement is the support of active/quiet after parameters only when a specific configuration is set to true (i.e. warp.activity.period.deletionsupport = true).

StevenLeRoux:
I think there are too much static configuration for now :) 

What about an activity mode : 

- WarpScript : AUTHENTICATE the stack, and enabling a flag (like  T ACTIVITY)
- HTTP : passing a &activity=yes parameter in the URL

hbs:
The purpose of the configuration parameters is to allow the platform
administrator to make choices which cannot be overriden.

StevenLeRoux:
This choice should be taken in the scope of an application, not for the entire platform.

If a customer wants to use it (I'm only taking about deletion integration), it doesn't have any impact for other customers.

I agree that the last activity feature has to be enabled by configuration. But deletion is to be observed on a per app basis to me.

cityzendata:
Deletion can be done using a wildcard token which spans multiple
applications, so this is not feasible.

StevenLeRoux:
I don't get it.

If a wildcard token is emitted, it's on behalf of an administrator. So no need for a specific hard config setting right ?

Or maybe you could detail the right escalation you have in mind?

hbs:
You should not consider that a token is for a single application, a token
could very well have been issued which spans multiple apps.

StevenLeRoux:
Yes but the "activity mode" would have an effect only for the scope of the write token.

Why making it a platform setting?

StevenLeRoux:
Latest tests have proved the PR to be working well.

Approved to be merged for us.

d33d33:
Hello!
Is it possible to sync this one with master?

This one is in prod on our infras since almost a year now 🎂 
Do you have plan to merge on master?
We will be glad help you if needed 😃

hbs:
This is planned for the next major release due this summer.

aurrelhebert:
Hi quick issue with FIND queries on an inmemory instance: lastactivity isn't propagated in result meta in [Standalone directory](https://github.com/cityzendata/warp10-platform/blob/master/warp10/src/main/java/io/warp10/standalone/StandaloneDirectoryClient.java#L518). 
Missing 
`meta.setLastActivity(metadata.getLastActivity());`

hbs:
StandaloneDirectoryClient has been corrected in 6e516d92f1ab56588f93f60cd55f85cb724fa415

PR #201, opened 2017-08-10
Add support for consumerid in Kafka ConsumerConnector

StevenLeRoux:
Add support for consumerid infix in Kafka ConsumerConnector to identify consumer instances running on the same host

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/Configuration.java
+++ b/warp10/src/main/java/io/warp10/continuum/Configuration.java
@@ -795,6 +795,11 @@ public class Configuration {
   public static final String STORE_KAFKA_DATA_GROUPID = "store.kafka.data.groupid";
 
   /**
+   * A prefix prepended to the Kafka ConsumerId
+   */
+  public static final String STORE_KAFKA_DATA_CONSUMERID_PREFIX = "store.kafka.data.consumerid.prefix";
+
+  /**
    * Client id to use to consume the data topic
    */
   public static final String STORE_KAFKA_DATA_CONSUMER_CLIENTID = "store.kafka.data.consumer.clientid";
--- a/warp10/src/main/java/io/warp10/continuum/store/Store.java
+++ b/warp10/src/main/java/io/warp10/continuum/store/Store.java
@@ -28,6 +28,8 @@ import io.warp10.crypto.CryptoUtils;
 import io.warp10.crypto.KeyStore;
 import io.warp10.sensision.Sensision;
 
+import java.net.InetAddress;
+import java.util.UUID;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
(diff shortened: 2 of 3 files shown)


hbs:
Why only allow this in Store?

StevenLeRoux:
Because multiple instances is mostly interesting for Store only. The current threading model (induced by Kafka < 0.9 and the lack of Coordinator to deal with the partition assignment) make it hard to provision both : 

- free resources for HA and recovering
- equal consumer instances repartition across hosts

Ingress & Directory don't really suffers from this, mainly due to the difference of load between metadata & data. Stores are being more pressured so we need this special option to manage multiples instances per host, multiplying the number of ConsumerConnector, but keeping a 1/1 ratio threads/kafka_threads.

warp10/src/main/java/io/warp10/continuum/store/Store.java
```diff
@@ -360,6 +362,16 @@ public void run() {
...
               props.setProperty("client.id", properties.getProperty(io.warp10.continuum.Configuration.STORE_KAFKA_DATA_CONSUMER_CLIENTID));
             }
+            if (null != properties.getProperty(io.warp10.continuum.Configuration.STORE_KAFKA_DATA_CONSUMERID_INFIX)) {
+              // If a consumerId inflix is provided, the consumerId is built the same way than inside the Kafka ConsumerConnector with the inflix part being suffixed to the hostname
+              UUID uuid = UUID.randomUUID();
+              String consumerUuid = String.format("%s_%s_%d-%s",
+                      InetAddress.getLocalHost().getHostName(),
+                      properties.getProperty(io.warp10.continuum.Configuration.STORE_KAFKA_DATA_CONSUMERID_INFIX),
+                      System.currentTimeMillis(),
+                      Long.toHexString(uuid.getMostSignificantBits()).substring(0,8));
```
hbs: Why only keep 32 bits of the 128bits UUID? What's the purpose of this?

warp10/src/main/java/io/warp10/continuum/Configuration.java
```diff
@@ -795,6 +795,11 @@
   public static final String STORE_KAFKA_DATA_GROUPID = "store.kafka.data.groupid";
 
   /**
+   * ConsumerId infix will be affixed inside the Kafka consumerId
+   */
+  public static final String STORE_KAFKA_DATA_CONSUMERID_INFIX = "store.kafka.data.consumerid.infix";
```
hbs: Please also modify etc/conf-distributed.template to explain the purpose of this configuration

StevenLeRoux: Are you ok with this : 
```

//
// Infix to be appended after hostname while consuming the data topic
//
#store.kafka.data.consumerid.infix =
```

warp10/src/main/java/io/warp10/continuum/store/Store.java
StevenLeRoux: I kept it ISO with the Kafka ConsumerConnector implementation which implement it this way.

hbs:
Could you rename 'infix' to 'prefix' since you moved it at the front of the consumer id string?

StevenLeRoux:
Moved infix to prefix the hostname, fixed the comments and squashed commits.

hbs:
Thanks for the PR.

PR #233, opened 2017-12-05
Add support for attributes via UPDATE endpoint

StevenLeRoux:
This PR allows for pushing attributes directly from a update endpoint.

This way, a source push metrics with the form "class{labels}{atttibutes} value"  and at cache expiration, the new meta will be taken into account.

This has to be considered as a best effort meta update in contrast to what /meta provides.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
+++ b/warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
@@ -269,8 +269,7 @@ public class Ingress extends AbstractHandler implements Runnable {
    */
   private final boolean rejectDelete;
 
-
-  private final boolean acceptAttributes;
+  final boolean acceptAttributes;
 
   public Ingress(KeyStore keystore, Properties props) {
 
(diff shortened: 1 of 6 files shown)


warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
```diff
@@ -742,7 +748,11 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
           }
           
           try {
-            encoder = GTSHelper.parse(lastencoder, line, extraLabels, now, maxValueSize, false);
+            if (this.acceptAttributes) {
+              encoder = GTSHelper.parse(lastencoder, line, extraLabels, now, maxValueSize, true);
```
miton18: encoder = GTSHelper.parse(lastencoder, line, extraLabels, now, maxValueSize, this.acceptAttributes);

// with that you can remove the if/else

StevenLeRoux: good idea 👍  

fixed and squashed

hbs:
This PR is not exhaustive, it does not provide the suggested feature in the standalone version neither in the streaming update endpoint.

Its behavior is also not guaranteed as the source used for the Metadata is not the one from the /meta endpoint and thus attributes might not be handled as expected, if handled at all.

Please open an enhancement issue to track progress around this feature request.

PR #237, opened 2018-01-03
Add new metric when an estimator is reset 

PierreZ:
This PR is adding a new metric called *warp.estimators.resets*. Thanks to it, Warp10 administrators can keep track of estimators resets per producer and per app. 

This is also fixing a bad metric name when resetting a per-app estimator.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
+++ b/warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
@@ -1026,6 +1026,16 @@ public class SensisionConstants {
   public static final String SENSISION_CLASS_CONTINUUM_ESTIMATORS_CACHED_PER_APP = "warp.estimators.cached.perapp";
 
   /**
+   * Number of times an estimator was reset
+   */
+  public static final String SENSISION_CLASS_CONTINUUM_ESTIMATOR_RESETS = "warp.estimator.resets";
+
+  /**
+   * Number of times an estimator was reset per app
+   */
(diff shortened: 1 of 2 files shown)


warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
```diff
@@ -1025,6 +1025,16 @@
    */
   public static final String SENSISION_CLASS_CONTINUUM_ESTIMATORS_CACHED_PER_APP = "warp.estimators.cached.perapp";
 
+  /**
+   * Number of times the estimators was reset
+   */
+  public static final String SENSISION_CLASS_CONTINUUM_ESTIMATORS_RESETS = "warp.estimators.resets";
```
hbs: Can you rename this metric to warp.estimator.resets as it concerns a single estimator.

warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
```diff
@@ -1025,6 +1025,16 @@
...
+
+  /**
+   * Number of times the estimators was reset per app
+   */
+  public static final String SENSISION_CLASS_CONTINUUM_ESTIMATORS_RESETS_PER_APP = "warp.estimators.resets.perapp";
```
hbs: Same remark, please rename class to warp.estimator.resets.perapp

PierreZ:
PR updated 😄

hbs:
Thanks for the PR.

PR #241, opened 2018-01-22
Fixed MAXRECURSION long/int casts.

ftence:
MAXRECURSION could not be used because of incorrect/missing casts, fixed that by forcing casts to int.
Same issue as https://github.com/cityzendata/warp10-platform/issues/21


warp10/src/main/java/io/warp10/script/functions/MAXRECURSION.java
```diff
@@ -45,11 +45,11 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
     
     long limit = ((Number) top).longValue();
```
hbs: Change it to int as you are only using int comparisons in the following code.

PR #273, opened 2018-04-04
Creation of WARP10FUNCTIONS

pi-r-p:
Usefull to check doc consistency
Usefull for extensions developpers, allow to check functions are
registered.


pi-r-p:
replaced by inventory extension

PR #289, opened 2018-04-12
RTFM now pointing to exact function documentation

pi-r-p:
Ready for next doc release. WARP10_FUNCTION_DOC_URL may evolve in the meantime.

RTFM  //returns "http://www.warp10.io/" (empty stack)
54 RTFM //returns "http://www.warp10.io/" (no string on top)
'BINOUZE' RTFM //returns "Unknown function name, please check http://www.warp10.io/"
'DROP' RTFM //returns "http://www.warp10.io/doc/DROP"
'**' RTFM  //returns "http://www.warp10.io/doc/9Xc"


hbs:
This breaks the behavior of RTFM which does not have parameters.

You can add another function which pushes the URL for a given function.

Call it RTFD (read the function doc)?

PierreZ:
What about MAN?

hbs:
MAN is probably more meaningful indeed :-)

PR #291, opened 2018-04-24
TYPEOF evol for PIMAGE type

pi-r-p:
Make a clear distinction between Pgraphics and Pimage objects


warp10/src/main/java/io/warp10/script/functions/TYPEOF.java
```diff
@@ -60,6 +61,7 @@
   public static final String TYPE_VECTOR = "VLIST";
   public static final String TYPE_REALVECTOR = "VECTOR";
   public static final String TYPE_REALMATRIX = "MATRIX";
+  public static final String TYPE_PGRAPHICSRAWIMAGE = "PIMAGE";
```
hbs: Why not call the constant TYPE_PIMAGE?

PR #306, opened 2018-05-31
Reducer rms

pi-r-p:
Returns the RMS of buckets + last position


warp10/src/main/java/io/warp10/script/WarpScriptLib.java
```diff
@@ -1286,6 +1287,8 @@
     functions.put("bucketizer.count.nonnull", new Count("bucketizer.count.nonnull", true));
     functions.put("bucketizer.mean.circular", new CircularMean.Builder("bucketizer.mean.circular", true));
     functions.put("bucketizer.mean.circular.exclude-nulls", new CircularMean.Builder("bucketizer.mean.circular.exclude-nulls", false));
+    functions.put("bucketizer.rms", new RMS("bucketizer.rms", false));
```
hbs: Can you also declare them as mappers and reducers since they implement those interfaces too.

PR #310, opened 2018-06-14
Automagically guess JAVA_HOME.

ftence:


warp10/src/main/sh/warp10-standalone.sh
```diff
@@ -22,7 +22,12 @@ OS=$(uname -s)
 
 if [ -z "$JAVA_HOME" ]; then
   echo "JAVA_HOME not set";
-  exit 1
+  JAVA_HOME=$(readlink -f /usr/bin/java | sed "s:\(/jre\)\{0,1\}/bin/java::")
```
hbs: readlink -f does not work on OSX, so this is not advised.

PR #311, opened 2018-06-14
Cleaner WARP10_HOME, resolving relative path.

ftence:


warp10/src/main/sh/warp10-standalone.sh
```diff
@@ -33,7 +33,7 @@ if [ -z "${WARP10_HOME}" ]; then
     WARP10_HOME=`pwd`
     popd > /dev/null 2>&1
   else
-    WARP10_HOME=$(dirname $(readlink -f $0))/..
+    WARP10_HOME=$(readlink -f $(dirname $0)/..)
```
hbs: readlink -f is not supported in OSX.

ftence: There's already a check for Darwin OS just above so this will only execute on non-OSX.

Maybe this could be simplified as the pushd/pwd/popd method works on Linux too, so it could be used without any check?
Is there any reason `pwd` is used instead of `pwd -P`. Should symlinks be kept or not? Is it working on OSX?

steven-gueguen:
After your modifications the OS variable remains unused. It could be removed

PR #314, opened 2018-06-15
Functions based on MATH2 can use (Long,Long) methods if available.

ftence:
Idea to make reflection calls faster.


warp10/src/main/java/io/warp10/script/functions/MATH2.java
```diff
@@ -22,18 +22,37 @@
...
+    WarpScriptException warpExceptionDouble = null;
+
+    try {
+      this.methodDouble = Math.class.getMethod(methodName, double.class, double.class);
+    } catch (Exception e) {
+      warpExceptionDouble = new WarpScriptException(e);
+    }
+
+    // No method suitable, throw
+    if (null == warpExceptionLong && null == warpExceptionDouble) {
```
hbs: This should probably be different.

PR #320, opened 2018-06-21
Correction of Pblend signature

pi-r-p:
Mode is now a string, as in Pfilter

### DIFF
--- a/warp10/src/main/java/io/warp10/script/processing/image/Pblend.java
+++ b/warp10/src/main/java/io/warp10/script/processing/image/Pblend.java
@@ -42,7 +42,48 @@ public class Pblend extends NamedWarpScriptFunction implements WarpScriptStackFu
     List<Object> params = ProcessingUtil.parseParams(stack, 9, 10);
         
     PGraphics pg = (PGraphics) params.get(0);
-    
+
+    String mode = "";
+    int imode = 0;
+    if (10 == params.size()) {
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/processing/image/Pblend.java
```diff
@@ -42,7 +42,48 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+    } else if ("HARD_LIGHT".equals(mode)) {
+      imode = PGraphics.HARD_LIGHT;
+    } else if ("SOFT_LIGHT".equals(mode)) {
+      imode = PGraphics.SOFT_LIGHT;
+    } else if ("DODGE".equals(mode)) {
+      imode = PGraphics.DODGE;
+    } else if ("BURN".equals(mode)) {
+      imode = PGraphics.BURN;
+    } else {
+      throw new WarpScriptException(getName() + ": invalid mode, should be 'BLEND', 'ADD', 'SUBTRACT', 'LIGHTEST', 'DARKEST', 'DIFFERENCE', 'EXCLUSION', 'MULTIPLY', 'SCREEN', 'OVERLAY', 'HARD_LIGHT', 'SOFT_LIGHT', 'DODGE', 'BURN'.");
```
hbs: should be -> should be one of

PR #322, opened 2018-06-22
Add PSHAPESVG to TYPEOF

pi-r-p:


warp10/src/main/java/io/warp10/script/functions/TYPEOF.java
```diff
@@ -64,6 +66,8 @@
   public static final String TYPE_REALMATRIX = "MATRIX";
   public static final String TYPE_PIMAGE = "PIMAGE";
   public static final String TYPE_PFONT = "PFONT";
+  public static final String TYPE_PSHAPESVG = "PSHAPESVG";
```
hbs: Could you change this to simply PSHAPE as it really is only a PShape created from an SVG, but there is nothing specific to its origin once it is created.

pi-r-p: Done. Documentation updated.

PR #335, opened 2018-08-03
Change process detection for docker compatibility

steven-gueguen:
Factorise user check and process check

### DIFF
--- a/warp10/src/main/sh/warp10-standalone.sh
+++ b/warp10/src/main/sh/warp10-standalone.sh
@@ -147,7 +147,7 @@ isUser() {
 #
 isStarted() {
   # Don't use 'ps -p' for docker compatibility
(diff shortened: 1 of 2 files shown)


warp10/src/main/sh/warp10-standalone.sh
```diff
@@ -132,17 +132,34 @@ moveDir() {
...
+    exit 1
+  fi
+}
+
+#
+# Return 0 if a Warp 10 instance is started
+#
+isStarted() {
+  # Don't use 'ps -p' for docker compatibility
+  if [ -e ${PID_FILE} ] && ps -eo pid |  grep "^\s*$(cat ${PID_FILE})$" > /dev/null; then
```
hbs: ps -eo does not work under Mac OS X

steven-gueguen: OK, fix for Mac OS X

PR #339, opened 2018-08-20
Switching from reflection to direct call for Math methods

ftence:
... to increase performances.
Import cleaning.

Performances increase are measured to be between 2% and 5% less time to execute.


warp10/src/main/java/io/warp10/script/functions/math/ACOS.java
```diff
@@ -14,45 +14,39 @@
...
+    if (!(op0 instanceof Number)) {
+      throw new WarpScriptException(getName() + " can only operate on numerical values.");
     }
-    
-    try {
-      stack.push(this.method.invoke(null, ((Number) o).doubleValue()));
-    } catch (Exception e) {
-      throw new WarpScriptException(e);
+
+    if ((op0 instanceof Double || op0 instanceof Long || op0 instanceof Float || op0 instanceof Integer || op0 instanceof BigDecimal)) {
```
hbs: What Number instances cannot be handled?

ftence: By definition: BigInteger, Byte, and Short and on top of that, only for Java 8, AtomicInteger, AtomicLong,  DoubleAccumulator, DoubleAdder, LongAccumulator, LongAdder.

Should we allow them too?

I think only AtomicInteger could be on the stack with the current code but others could be in the future or in extensions.

hbs:
I think supporting all Numbers probably makes more sense. The only thing to ensure is that we correctly chose between long or double conversion if the Math function accepts both.

warp10/src/main/java/io/warp10/script/functions/math/ADDEXACT.java
```diff
@@ -0,0 +1,57 @@
...
+
+    Object op1 = stack.pop();
+
+    if (!(op1 instanceof Number)) {
+      throw new WarpScriptException(getName() + " can only operate on numerical values.");
+    }
+
+    if ((op1 instanceof Long || op1 instanceof Integer)
+        && (op0 instanceof Long || op0 instanceof Integer)) {
+      stack.push(Math.addExact(((Number) op1).longValue(), ((Number) op0).longValue()));
```
hbs: If parameters are both Integer, calling intValue makes more sense

warp10/src/main/java/io/warp10/script/functions/math/DECREMENTEXACT.java
```diff
@@ -0,0 +1,50 @@
...
+  @Override
+  public Object apply(WarpScriptStack stack) throws WarpScriptException {
+    Object op0 = stack.pop();
+
+    if (!(op0 instanceof Number)) {
+      throw new WarpScriptException(getName() + " can only operate on numerical values.");
+    }
+
+    if ((op0 instanceof Long || op0 instanceof Integer)) {
+      stack.push(Math.decrementExact(((Number) op0).longValue()));
```
hbs: If parameter is an Integer calling intValue makes more sense

warp10/src/main/java/io/warp10/script/functions/math/FLOORDIV.java
```diff
@@ -0,0 +1,57 @@
...
+
+    Object op1 = stack.pop();
+
+    if (!(op1 instanceof Number)) {
+      throw new WarpScriptException(getName() + " can only operate on numerical values.");
+    }
+
+    if ((op1 instanceof Long || op1 instanceof Integer)
+        && (op0 instanceof Long || op0 instanceof Integer)) {
+      stack.push(Math.floorDiv(((Number) op1).longValue(), ((Number) op0).longValue()));
```
hbs: If parameters are Integer, call intValue

warp10/src/main/java/io/warp10/script/functions/math/FLOORMOD.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    Object op1 = stack.pop();
+
+    if (!(op1 instanceof Number)) {
+      throw new WarpScriptException(getName() + " can only operate on numerical values.");
+    }
+
+    if ((op1 instanceof Long || op1 instanceof Integer)
+        && (op0 instanceof Long || op0 instanceof Integer)) {
```
hbs: If parameters are Integer, call intValue

warp10/src/main/java/io/warp10/script/functions/math/INCREMENTEXACT.java
```diff
@@ -0,0 +1,50 @@
...
+  @Override
+  public Object apply(WarpScriptStack stack) throws WarpScriptException {
+    Object op0 = stack.pop();
+
+    if (!(op0 instanceof Number)) {
+      throw new WarpScriptException(getName() + " can only operate on numerical values.");
+    }
+
+    if ((op0 instanceof Long || op0 instanceof Integer)) {
+      stack.push(Math.incrementExact(((Number) op0).longValue()));
```
hbs: call intValue if parameters are both Integer

warp10/src/main/java/io/warp10/script/functions/math/MULTIPLYEXACT.java
```diff
@@ -0,0 +1,57 @@
...
+
+    Object op1 = stack.pop();
+
+    if (!(op1 instanceof Number)) {
+      throw new WarpScriptException(getName() + " can only operate on numerical values.");
+    }
+
+    if ((op1 instanceof Long || op1 instanceof Integer)
+        && (op0 instanceof Long || op0 instanceof Integer)) {
+      stack.push(Math.multiplyExact(((Number) op1).longValue(), ((Number) op0).longValue()));
```
hbs: call intValue if parameters are Integer

warp10/src/main/java/io/warp10/script/functions/math/NEGATEEXACT.java
```diff
@@ -0,0 +1,50 @@
...
+  @Override
+  public Object apply(WarpScriptStack stack) throws WarpScriptException {
+    Object op0 = stack.pop();
+
+    if (!(op0 instanceof Number)) {
+      throw new WarpScriptException(getName() + " can only operate on numerical values.");
+    }
+
+    if ((op0 instanceof Long || op0 instanceof Integer)) {
+      stack.push(Math.negateExact(((Number) op0).longValue()));
```
hbs: call intValue if parameter is Integer

warp10/src/main/java/io/warp10/script/functions/math/SUBTRACTEXACT.java
```diff
@@ -0,0 +1,57 @@
...
+
+    Object op1 = stack.pop();
+
+    if (!(op1 instanceof Number)) {
+      throw new WarpScriptException(getName() + " can only operate on numerical values.");
+    }
+
+    if ((op1 instanceof Long || op1 instanceof Integer)
+        && (op0 instanceof Long || op0 instanceof Integer)) {
+      stack.push(Math.subtractExact(((Number) op1).longValue(), ((Number) op0).longValue()));
```
hbs: call intValue if parameters are Integer

warp10/src/main/java/io/warp10/script/functions/math/TOINTEXACT.java
```diff
@@ -0,0 +1,50 @@
...
+  @Override
+  public Object apply(WarpScriptStack stack) throws WarpScriptException {
+    Object op0 = stack.pop();
+
+    if (!(op0 instanceof Number)) {
+      throw new WarpScriptException(getName() + " can only operate on numerical values.");
+    }
+
+    if ((op0 instanceof Long || op0 instanceof Integer)) {
+      stack.push(Math.toIntExact(((Number) op0).longValue()));
```
hbs: Should cast the result to a long

ftence: Agreed.

warp10/src/main/java/io/warp10/script/functions/math/ADDEXACT.java
hbs: The purpose of the *Exact methods is to check for overflow given the type of their parameters, by casting Integer to Long you get rid of the overflow check for Integer.

We could chose to return an int if both parameters were int, this would make it coherent with the other functions and will not be a problem if you never generate Integers on the stack.

PR #342, opened 2018-08-23
Allow runners to really use runner.nthreads, code cleanup.

ftence:
For a detailled explanation on ThreadPoolExecutor, see:
http://www.bigsoft.co.uk/blog/2009/11/27/rules-of-a-threadpoolexecutor-pool-size

### DIFF
--- a/warp10/src/main/java/io/warp10/script/ScriptRunner.java
+++ b/warp10/src/main/java/io/warp10/script/ScriptRunner.java
@@ -81,6 +81,15 @@ import java.util.zip.GZIPOutputStream;
  */
 public class ScriptRunner extends Thread {
 
+  private class NamedThreadFactory implements ThreadFactory {
+    private AtomicLong threadNumber = new AtomicLong();
+
+    @Override
+    public Thread newThread(Runnable r) {
+      return new Thread(r, "[Warp ScriptRunner Executor #" + threadNumber.getAndIncrement() + "]");
+    }
+  }
+
   protected static final byte[] CLEAR = "\nCLEAR\n".getBytes(Charsets.UTF_8);
 
   protected ExecutorService executor;
@@ -209,14 +218,7 @@ public class ScriptRunner extends Thread {
       this.minperiod = Long.parseLong(config.getProperty(Configuration.RUNNER_MINPERIOD));
       this.endpoint = config.getProperty(Configuration.RUNNER_ENDPOINT);
 
-      ThreadPoolExecutor runnersExecutor = new ThreadPoolExecutor(nthreads, nthreads, 30000L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(nthreads * 256), new ThreadFactory() {
-        private AtomicLong threadNumber = new AtomicLong();
-
-        @Override
-        public Thread newThread(Runnable r) {
-          return new Thread(r, "[Warp ScriptRunner Executor #" + threadNumber.getAndIncrement() + "]");
-        }
-      });
+      ThreadPoolExecutor runnersExecutor = new ThreadPoolExecutor(nthreads, nthreads, 30000L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(nthreads * 256), new NamedThreadFactory());
       runnersExecutor.allowCoreThreadTimeOut(true);
 
       this.executor = runnersExecutor;
@@ -252,7 +254,11 @@ public class ScriptRunner extends Thread {
       String strategy = config.getProperty(Configuration.RUNNER_KAFKA_CONSUMER_PARTITION_ASSIGNMENT_STRATEGY);
       int nthreads = Integer.parseInt(config.getProperty(Configuration.RUNNER_KAFKA_NTHREADS));
       long commitPeriod = Long.parseLong(config.getProperty(Configuration.RUNNER_KAFKA_COMMITPERIOD));
-      this.executor = new ThreadPoolExecutor(1, nthreads, 30000L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(nthreads * 256));
+
+      ThreadPoolExecutor runnersExecutor = new ThreadPoolExecutor(nthreads, nthreads, 30000L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(nthreads * 256), new NamedThreadFactory());
+      runnersExecutor.allowCoreThreadTimeOut(true);
+
+      this.executor = runnersExecutor;
 
       new KafkaSynchronizedConsumerPool(zkconnect, topic, clientid, groupid, strategy, nthreads, commitPeriod, new ScriptRunnerConsumerFactory(this));
     }
--- a/warp10/src/main/java/io/warp10/script/ScriptRunner.java
+++ b/warp10/src/main/java/io/warp10/script/ScriptRunner.java
@@ -87,18 +87,24 @@ public class ScriptRunner extends Thread {
 
     NamedThreadFactory() {
       SecurityManager s = System.getSecurityManager();
-      group = (s != null) ? s.getThreadGroup() :
-          Thread.currentThread().getThreadGroup();
+      if (null == s) {
+        group = Thread.currentThread().getThreadGroup();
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/ScriptRunner.java
```diff
@@ -39,265 +68,226 @@
...
-      this.nthreads = Integer.parseInt(config.getProperty(Configuration.RUNNER_NTHREADS));
+      int nthreads = Integer.parseInt(config.getProperty(Configuration.RUNNER_NTHREADS));
       this.scanperiod = Long.parseLong(config.getProperty(Configuration.RUNNER_SCANPERIOD));
       this.minperiod = Long.parseLong(config.getProperty(Configuration.RUNNER_MINPERIOD));
       this.endpoint = config.getProperty(Configuration.RUNNER_ENDPOINT);
-      
-      this.executor = new ThreadPoolExecutor(1, nthreads, 30000L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(nthreads * 256));
-  
+
+      ThreadPoolExecutor runnersExecutor = new ThreadPoolExecutor(nthreads, nthreads, 30000L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(nthreads * 256), new ThreadFactory() {
```
ftence: Here is the important change.

warp10/src/main/java/io/warp10/script/ScriptRunner.java
```diff
@@ -252,7 +254,11 @@ public Thread newThread(Runnable r) {
       String strategy = config.getProperty(Configuration.RUNNER_KAFKA_CONSUMER_PARTITION_ASSIGNMENT_STRATEGY);
       int nthreads = Integer.parseInt(config.getProperty(Configuration.RUNNER_KAFKA_NTHREADS));
       long commitPeriod = Long.parseLong(config.getProperty(Configuration.RUNNER_KAFKA_COMMITPERIOD));
-      this.executor = new ThreadPoolExecutor(1, nthreads, 30000L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(nthreads * 256));
+
+      ThreadPoolExecutor runnersExecutor = new ThreadPoolExecutor(nthreads, nthreads, 30000L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(nthreads * 256), new NamedThreadFactory());
```
ftence: Here is the important change.

warp10/src/main/java/io/warp10/script/ScriptRunner.java
```diff
@@ -39,265 +68,232 @@
...
-    } else {
-      isWorker = false;
-    }
-  
-    if (isStandalone || isScheduler) {      
+    isStandalone = configuredRoles.contains("standalone");
+
+    isScheduler = configuredRoles.contains("scheduler");
+
+    boolean isWorker = configuredRoles.contains("worker");
```
hbs: Why redefining isWorker here?

warp10/src/main/java/io/warp10/script/ScriptRunner.java
```diff
@@ -39,265 +68,232 @@
...
-      this.executor = new ThreadPoolExecutor(1, nthreads, 30000L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(nthreads * 256));
-      
-      this.consumerPool = new KafkaSynchronizedConsumerPool(zkconnect, topic, clientid, groupid, strategy, nthreads, commitPeriod, new ScriptRunnerConsumerFactory(this));
+
+      ThreadPoolExecutor runnersExecutor = new ThreadPoolExecutor(nthreads, nthreads, 30000L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(nthreads * 256), new NamedThreadFactory());
+      runnersExecutor.allowCoreThreadTimeOut(true);
+
+      this.executor = runnersExecutor;
+
+      new KafkaSynchronizedConsumerPool(zkconnect, topic, clientid, groupid, strategy, nthreads, commitPeriod, new ScriptRunnerConsumerFactory(this));
```
hbs: Code is clearer if you keep a reference to the created consumer pool

warp10/src/main/java/io/warp10/script/ScriptRunner.java
```diff
@@ -39,265 +68,232 @@
...
  */
 public class ScriptRunner extends Thread {
-  
+
+  private class NamedThreadFactory implements ThreadFactory {
+    private AtomicLong threadNumber = new AtomicLong();
+
+    @Override
+    public Thread newThread(Runnable r) {
+      return new Thread(r, "[Warp ScriptRunner Executor #" + threadNumber.getAndIncrement() + "]");
```
hbs: Thread needs to be a daemon Thread.
Group needs to be set.
Priority needs to be set.

Check the DefaultThreadFactory from Executor.

warp10/src/main/java/io/warp10/script/ScriptRunner.java
ftence: isWorker was a member of ScriptRunner but was only used in its constructor so I made it a local variable.

warp10/src/main/java/io/warp10/script/ScriptRunner.java
ftence: I do agree this line looks strange without keeping a reference but not storing it in a member of the class makes it clear that it is not used/modified elsewhere.
Can a local variable do? What about the GC, could it destroy the instance if there are no ref to this instance?

warp10/src/main/java/io/warp10/script/ScriptRunner.java
ftence: Thanks for the pointer, will fix.

warp10/src/main/java/io/warp10/script/ScriptRunner.java
hbs: The overall understanding of the code is probably better if isWorker is defined with the other two, even if it is not used (yet?).

warp10/src/main/java/io/warp10/script/ScriptRunner.java
hbs: The GC issue would have to be investigated, a local variable would not work if the GC can discard the object, so for now it's probably safer to stick with the setting of the field.

warp10/src/main/java/io/warp10/script/ScriptRunner.java
```diff
@@ -39,265 +68,248 @@
...
 public class ScriptRunner extends Thread {
-  
+
+  static class NamedThreadFactory implements ThreadFactory {
+    final ThreadGroup group;
+    final AtomicInteger threadNumber = new AtomicInteger(1);
+
+    NamedThreadFactory() {
+      SecurityManager s = System.getSecurityManager();
+      group = (s != null) ? s.getThreadGroup() :
```
hbs: leave on a single line.

warp10/src/main/java/io/warp10/script/ScriptRunner.java
```diff
@@ -39,265 +68,248 @@
...
+          Thread.currentThread().getThreadGroup();
+    }
+
+    public Thread newThread(Runnable r) {
+      Thread t = new Thread(group, r,
```
hbs: Leave on a single line

warp10/src/main/java/io/warp10/script/ScriptRunner.java
```diff
@@ -39,265 +68,248 @@
...
+          "[Warp ScriptRunner Thread #" + threadNumber.getAndIncrement() + "]",
+          0);
+      if (t.isDaemon())
```
hbs: Use braces {}

warp10/src/main/java/io/warp10/script/ScriptRunner.java
```diff
@@ -39,265 +68,248 @@
...
+        t.setDaemon(false);
+      if (t.getPriority() != Thread.NORM_PRIORITY)
```
hbs: Use braces {}

PR #343, opened 2018-08-24
Allow the List pushed back by ELAPSED to be modified.

ftence:


hbs:
Why would you want to modify the actual elapsed list?

ftence:
As the first element does not give the same kind of information as the other elements one could use REMOVE to separate them. REMOVE applied to ELAPSED does, of course, not work and the error message is hard to understand.

hbs:
The reason why it was exposed as an unmodifiable list was because ELAPSED may be called before the end of the script, if the list is modified, the content will not be coherent as one would expect.

But we can consider that the modification is at the risk of the modifier...

ftence:
To clarify the title of this PR, the modification is that ELAPSED returns a copy of the list.

This, however, changes the behavior of the function more than I first thought as:
```
TIMINGS
ELAPSED
NOOP
```
will not give the same result as the timing of NOOP will be included if `Collections.unmodifiableList(elapsed)` is used and will NOT be included if `new ArrayList<Long>(elapsed)` is used.

I have the feeling that the current behavior of ELAPSE is rather hard to grasp without going through the source code. To makes it easier to understand, we could modify TIMINGS, NOTIMINGS and ELAPSED so that ELAPSED gives a modifiable list of the elapsed time for each line between the previous TIMINGS and NOTIMINGS or ELAPSED. This may need some thinking about how nested calls could work.

That would allow
```
TIMINGS
@myMacro
NOTIMINGS
ELAPSED

// Some code we don't want to time

TIMINGS
@myOtherMacro
NOTIMINGS
ELAPSED

// On the stack: two modifiable lists with elapsed time only for their "scope".
```

PR #347, opened 2018-08-30
Add the REVATLEAST function which can test revision numbers.

ftence:
Usage:
'1.2.22' REVATLEAST
> true

'1.3' REVATLEAST
> false

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/REVATLEAST.java
+++ b/warp10/src/main/java/io/warp10/script/functions/REVATLEAST.java
@@ -1,0 +1,106 @@
+//
+//   Copyright 2018  Cityzen Data
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


warp10/src/main/java/io/warp10/script/functions/REVATLEAST.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /**
+   * Parse strings of the format \d+(\.\d+)*-.* and returns an array of integers corresponding to the revision.
+   * @param revision The revision string to be parsed.
+   * @param errorMessage The error message to be included in the WarpScriptException in case a problem arises.
+   * @return An array or integers corresponding to each part of the revision numbering.
+   * @throws WarpScriptException In case the input string is malformed.
+   */
+  private static Integer[] SplitRev(String revision, String errorMessage) throws WarpScriptException {
```
hbs: return int[]

warp10/src/main/java/io/warp10/script/functions/REVATLEAST.java
```diff
@@ -0,0 +1,106 @@
...
+import io.warp10.Revision;
+import io.warp10.script.NamedWarpScriptFunction;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.script.WarpScriptStackFunction;
+
+/**
+ * Return true if the current revision tag is newer or equal to the given revision tag.
+ */
+public class REVATLEAST extends NamedWarpScriptFunction implements WarpScriptStackFunction {
```
hbs: I think MINREV would be a better name than REVATLEAST.

PR #349, opened 2018-08-31
Try to keep sorted and reversed flags when adding a new tick.

ftence:


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -924,8 +924,21 @@ public static final int setValue(GeoTimeSerie gts, long timestamp, long geoxppoi
     //
     
     if (gts.values == idx) {
-      // Reset 'sorted' flag as we add a value
-      gts.sorted = false;
+      // Try to keep 'sorted' flag if possible
```
hbs: What is the performance impact of all those tests when adding lots of datapoints?

ftence: When adding 10,000,000 ticks in increasing order, thus keeping the sorted flag, the proposed version in this PR takes between 0.91% and 1.83% more time to execute.

When adding 10,000,000 ticks in random order, thus not keeping the sorted flag, the proposed version in this PR takes between 0.00% and 0.34% more time to execute.


Here are the raw JMH results, results between rand and incr are not to be compared:

| Benchmark | Mode | Cnt | Score | Error | Units |
| --------- | ---- | --- | ----- | ----- |-------|
| AddValueSorted.setValueKeepSortIncr | avgt | 1000 | 1.773 | ± 0.007 | s/op |
| AddValueSorted.setValueNoSortIncr | avgt | 1000|  1.749 | ± 0.001 | s/op |
| AddValueSorted.setValueKeepSortRand | avgt | 1000 | 1.759 | ± 0.002 | s/op |
| AddValueSorted.setValueNoSortRand | avgt | 1000 | 1.756 | ± 0.001 | s/op |


Here is the graph, careful, the Y axis does not start from 0. This is because otherwise we wouldn't be able to see the CIs.
setvaluebench

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -134,6 +134,71 @@ public static final GeoTimeSerie sort(GeoTimeSerie gts, boolean reversed) {
   public static final GeoTimeSerie sort(GeoTimeSerie gts) {
     return sort(gts, false);
   }
+
+  public enum BinarySearchTickChoice {
+    ARBITRARY, FIRST, LAST
```
hbs: Could you comment the various values, explaining each strategy

PR #357, opened 2018-09-19
FETCH to work with MINLONG datapoints.

ftence:


ftence:
`[ $token 'example' {} MAXLONG MINLONG ] FETCH` can work with this PR.

PR #365, opened 2018-10-29
Fix the time limit issue in FORSTEP

pi-r-p:
It was fixed to 5000ms, whatever the MAXLOOP value.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/FORSTEP.java
+++ b/warp10/src/main/java/io/warp10/script/functions/FORSTEP.java
@@ -111,7 +111,7 @@ public class FORSTEP extends NamedWarpScriptFunction implements WarpScriptStackF
 
     while (true) {
       if (System.currentTimeMillis() - now > maxtime) {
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/functions/FORSTEP.java
```diff
@@ -106,9 +106,11 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
     
     long now = System.currentTimeMillis();
-    
+
+    long maxtime = this.maxtime > 0 ? this.maxtime : (long) stack.getAttribute(WarpScriptStack.ATTRIBUTE_LOOP_MAXDURATION);
+
     while (true) {
-      if (System.currentTimeMillis() - now > this.maxtime) {
+      if (System.currentTimeMillis() - now > maxtime) {
         throw new WarpScriptException(getName() + " executed for too long (> " + this.maxtime + " ms).");
```
hbs: Error message should use maxtime, not this.maxtime

pi-r-p: Indeed. Fixed.

PR #368, opened 2018-11-22
Fix for retrocompatibility of FIND signature without LIST or MAP.

pi-r-p:
Somewhere in the wild, there is some warpscripts which REXEC "readtoken classname {} FIND", without the LIST nor the MAP syntax. This was working in Warp 10 1.x.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/FIND.java
+++ b/warp10/src/main/java/io/warp10/script/functions/FIND.java
@@ -155,8 +155,11 @@ public class FIND extends NamedWarpScriptFunction implements WarpScriptStackFunc
       if (this.metaset) {
         throw new WarpScriptException(getName() + " expects a list of parameters on top of the stack.");
       }
-      mapparams = ((Map)top).containsKey(FETCH.PARAM_TOKEN);
       //syntax error tolerance for older systems : $readtoken 'classselector' { labels selector } FIND
+      mapparams = ((Map)top).containsKey(FETCH.PARAM_TOKEN) &&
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/functions/FIND.java
```diff
@@ -155,7 +155,8 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       if (this.metaset) {
         throw new WarpScriptException(getName() + " expects a list of parameters on top of the stack.");
       }
-      mapparams = true;
+      mapparams = ((Map)top).containsKey(FETCH.PARAM_TOKEN);
```
hbs: What happens if the object in 'top' is not a Map?

pi-r-p: The check is done three lines upper, the cast is safe.
The functional problem is someone using this unsupported syntax with a label called "token" as a label selector. 

Unit tests : 
`[ $READTOKEN '~.*speed.*' { 'id' 'xaaacgz' } ] FIND` //supported syntax with list input: OK
`$READTOKEN '~.*speed.*' { 'id' 'xaaacgz' } FIND` //unsupported syntax: OK with this PR
`$READTOKEN '~.*speed.*' { 'token' 'xx' } FIND` //unsupported syntax, but will be seen as the new map input syntax. Will fail with the message  `missing parameters 'class', 'labels', 'selector' or 'selectors'`

hbs: The check should be more specific to reduce the possible errors. Beyond checking 'token', check for either 'class'/'labels' or 'selector'/'selectors'

pi-r-p:
Not the easiest boolean expression to read, but here it is.

PR #369, opened 2018-11-22
Add an explicit error message when asking impossible geometry

pi-r-p:
Previously, unexpected null pointer exception.


hbs:
The bump to GeoXPLib 1.0.0-rc14 has the shape creation methods return null when no cover could be computed given the parameters. This was already the case for uniform shapes, it is now the case for regular shapes also.

PR #372, opened 2018-11-23
Handler now handles only REQUEST to avoid being called twice when there is an error.

ftence:
When a macro creates an exception on the initial REQUEST, the handler is called again to handle the ERROR, thus calling the macro twice.
This PR avoid the call of the macro for the ERROR handling.

### DIFF
--- a/warp10/src/main/java/io/warp10/plugins/http/WarpScriptHandler.java
+++ b/warp10/src/main/java/io/warp10/plugins/http/WarpScriptHandler.java
@@ -24,6 +24,7 @@ import org.eclipse.jetty.http.MimeTypes;
 import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.handler.AbstractHandler;
 
+import javax.servlet.DispatcherType;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/plugins/http/WarpScriptHandler.java
```diff
@@ -50,6 +51,12 @@ public WarpScriptHandler(HTTPWarp10Plugin plugin) {
 
   @Override
   public void handle(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Only handle REQUEST
+    if(DispatcherType.REQUEST != baseRequest.getDispatcherType()){
+      baseRequest.setHandled(true);
```
hbs: should it be true or false? Is there a default error handler?

ftence: It should be true.

When the request is done, the handler is first called with DispatcherType.REQUEST. If this handler throws an exception, the response status is automatically set to 500.
In case of an exception, the handler is then called again with DispatcherType.ERROR. It allows error management and modification of the response. If the handler does not set the request as handled, the request is considered as not handled, as there are no other handler, and the status is changed to 404. So it should be set to true to keep the 500 status.

If one want exceptions of the macro to be handled, the macro should do it itself using a TRY and setting the response as wanted.

PR #377, opened 2018-11-27
Add integrated plugins, extension and their conf to conf templates

ftence:


ftence:
One default value is only valid with #376

PR #379, opened 2018-11-30
Better handling of negative timescale & bucketized typos

ftence:


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -7709,19 +7709,34 @@ public static GeoTimeSerie fuse(Collection<GeoTimeSerie> chunks) throws WarpScri
...
+
+  /**
+   * Multiply the ticks of a GTS instance by a factor.
+   * @param gts The GTS instance whose ticks are to be scaled.
+   * @param scale The factor to apply to the ticks, cannot be negative for bucketized GTSs.
+   * @return A copy of the given GTS instance with its tick scaled.
+   * @throws WarpScriptException If scale is negative and the GTS instance is bucketized or if the bucketspan of the given bucketized GTS would be 0 after scaling.
+   */
   public static GeoTimeSerie timescale(GeoTimeSerie gts, double scale) throws WarpScriptException {
+    if (0 > scale && isBucketized(gts)) {
```
hbs: Use 0.0D instead of 0 and use scale < 0 which is clearer

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -7709,19 +7709,34 @@ public static GeoTimeSerie fuse(Collection<GeoTimeSerie> chunks) throws WarpScri
...
     }
-    
-    scaled.lastbucket = (long) (scaled.lastbucket * scale);        
+
+    scaled.lastbucket = (long) (scaled.lastbucket * scale);
     scaled.bucketspan = (long) (scaled.bucketspan * scale);
 
-    if (gts.bucketspan != 0 && 0 == scaled.bucketspan) {
-      throw new WarpScriptException("Down scaling is too agressive, bucketspan would end up 0.");
+    if (0 > scale) {
```
hbs: Use 0.0D instead of 0 and use scale < 0 which is clearer

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -7709,19 +7709,34 @@ public static GeoTimeSerie fuse(Collection<GeoTimeSerie> chunks) throws WarpScri
...
+      scaled.reversed = !scaled.reversed;
```
hbs: Shouldn't this be applied only if sorted is true?

PR #386, opened 2018-12-12
Get payload with URLFETCH even if status is 4XX or 5XX

ftence:


warp10/src/main/java/io/warp10/script/ext/urlfetch/URLFETCH.java
```diff
@@ -151,7 +151,14 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
 
-        InputStream in = conn.getInputStream();
+        InputStream in;
+        // When there is an error (response code is 404 for instance), body is in the error stream.
+        try {
+          in = conn.getInputStream();
+        }
```
hbs: Please use } catch (...

warp10/src/main/java/io/warp10/script/ext/urlfetch/URLFETCH.java
```diff
@@ -151,7 +151,14 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+        catch(IOException ioe){
```
hbs: Use a space before '{'

warp10/src/main/java/io/warp10/script/ext/urlfetch/URLFETCH.java
```diff
@@ -151,7 +150,13 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
 
-        InputStream in = conn.getInputStream();
+        InputStream in;
```
hbs: in should be initialized to null

PR #389, opened 2018-12-19
Bring forward plugin load errors

pi-r-p:
After Warp10 2.0.1 update with an old config file, the quantum loading error is really not obvious.
These small changes will facilitate the admin job.


pi-r-p:
moved plug in and extension errors to a logger, add stdout appender to rootlogger.

pi-r-p:
double appender lead to double logs as we also log stdout. corrected.

hbs:
Is this PR still to be considered for merging?

pi-r-p:
Yes, but not complete. Need a return value in case of service launch error. I must end the job.

PR #396, opened 2019-01-24
Allow plugins or extensions to control the output of TYPEOF on classes they would introduce

randomboolean:


hbs:
Allowing this without maintaining a registry of external types would lead to a zoo. Can you modify handling of Typeofables by prefixing the output type with 'X-' so external types will be easily recognizable

hbs:
Don't forget to update the doc of TYPEOF to mention this interface and the fact that the types will be ouput with a 'X-' prefix

PR #402, opened 2019-03-07
Add type check to */- operations involving a GTS and a value.

ftence:
Makes error message clearer.

Example:
```NEWGTS 'a' /```

- without PR returns:
`java.lang.String cannot be cast to java.lang.Number`
- with this PR returns:
`/ can only operate on numeric values or Geo Time Series .`

### DIFF
--- a/warp10/src/main/java/io/warp10/script/binary/DIV.java
+++ b/warp10/src/main/java/io/warp10/script/binary/DIV.java
@@ -143,7 +143,7 @@ public class DIV extends NamedWarpScriptFunction implements WarpScriptStackFunct
       }
 
       stack.push(result);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: warp10/src/main/java/io/warp10/script/binary/DIV.java [hunk 1]
```diff
@@ -143,7 +143,7 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       }
 
       stack.push(result);
-    } else if (op1 instanceof GeoTimeSerie || op2 instanceof GeoTimeSerie) {
+    } else if ((op1 instanceof GeoTimeSerie && op2 instanceof Number )|| (op1 instanceof Number && op2 instanceof GeoTimeSerie)) {
```
hbs: Can you insert a space before ||, ditto for the other two.

PR #412, opened 2019-03-13
fix(REXEC): allow deactivate REXEC in Warp 10 2.0

### LINKED ISSUE #411: Deactivate REXEC
REXEC since Warp10 2.0 is now a Warp 10 extension. Migrating to Warp 10 2.0 break the behavior of some scripts as the REXEC function isn't allowed. 
To resolve this issue we can simply keep a Warp 10 properties to allow the administrator to disable the REXEC extension?

aurrelhebert:
For example this PR would fix #411

### DIFF
--- a/warp10/src/main/java/io/warp10/WarpRun.java
+++ b/warp10/src/main/java/io/warp10/WarpRun.java
@@ -35,7 +35,10 @@ public class WarpRun {
   public static void main(String[] args) throws Exception {
     try {
       System.setProperty(Configuration.WARP10_QUIET, "true");
-      System.setProperty(Configuration.WARPSCRIPT_REXEC_ENABLE, "true");
+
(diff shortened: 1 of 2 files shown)


hbs:
The REXEC function is now part of the RexecWarpScriptExtension and thus is only available if the extension is loaded, i.e. if an explicit configuration is present in the Warp 10 configuration file.

If no extension loading configuration is present, the extension will not be loaded.

hbs:
To address #411, we have an upcoming extension called ShadowWarpScriptExtension which will allow you to configure which functions should be:

- undefined
- replaced by a FAIL or MSGFAIL
- replaced by a NOOP
- replaced by a macro

This will give you the flexibility you want for making REXEC known and an alias for FAIL.

The Shadow extension is planned to be included in release 2.1 which will be out before the end of March.

PR #417, opened 2019-03-18
Byte array concatenation

pi-r-p:
ArrayUtils already used elsewhere in the project. Not a new dependency.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/binary/ADD.java
+++ b/warp10/src/main/java/io/warp10/script/binary/ADD.java
@@ -27,7 +27,10 @@ import io.warp10.script.WarpScriptException;
 import io.warp10.script.WarpScriptStack;
 import io.warp10.script.WarpScriptStack.Macro;
 
-import java.util.*;
+import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/binary/ADD.java
```diff
@@ -213,8 +214,10 @@ public Object op(GeoTimeSerie gtsa, GeoTimeSerie gtsb, int idxa, int idxb) {
       }      
 
       stack.push(result);          
+    } else if (op1 instanceof byte[] && op2 instanceof byte[]) {
+      stack.push(ArrayUtils.addAll((byte[])op1,(byte[])op2));
     } else {
-      throw new WarpScriptException(getName() + " can only operate on numeric, string, lists, matrices, vectors, Geo Time Series and macro values.");
+      throw new WarpScriptException(getName() + " can only operate on numeric, string, lists, matrices, vectors, Geo Time Series, byte array and macro values.");
```
hbs: Use plural form, i.e. byte arrays

PR #419, opened 2019-03-20
Formatted

randomboolean:
The FormattedWarpScriptFunction abstract class takes care of how its implemented functions consume their argument off the top of the stack. It raises meaningful error messages when the argument parsing fails. It also has an automatic doc generation method.

FormattedWarpScriptFunctionTest is an example of implementation with unit tests. 

The typeof() method of TYPEOF has been made static and can take a class as argument (before it could only take an instance).


warp10/src/test/java/io/warp10/script/formatted/FormattedWarpScriptFunctionTest.java
```diff
@@ -0,0 +1,159 @@
...
+    StringBuilder props = new StringBuilder();
+
+    props.append("warp.timeunits=us");
+    WarpConfig.setProperties(new StringReader(props.toString()));
+
+    WarpScriptLib.addNamedWarpScriptFunction(new FormattedWarpScriptFunctionTest());
+  }
+
+  @Test
+  public void test1() throws Exception {
```
hbs: Give meaningful names to tests so we know what is being tested

warp10/src/main/java/io/warp10/script/functions/TYPEOF.java
```diff
@@ -85,57 +85,69 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
     return stack;
   }
 
-  public String typeof(Object o) {
+  public static String typeof(Object o) {
     if (null == o) {
       return (TYPE_NULL);
-    } else if (o instanceof String) {
+    } else if (o instanceof Typeofable) {
```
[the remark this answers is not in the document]
randomboolean: Since the typeof() method of a typeofable is non-static, it needs an instance to operate.
For a Typeofable, typeof(Class c) tries to instanciate an object and can fail, in which case we return the canonical classname.

warp10/src/main/java/io/warp10/script/formatted/FormattedWarpScriptFunction.java
```diff
@@ -0,0 +1,305 @@
...
+      args = new ArrayList<>();
+      optArgs = new ArrayList<>();
+    }
+
+    public ArgumentsBuilder addArgument(Class<?> clazz, String name, String doc) {
+      args.add(new ArgumentSpecification(clazz, name, doc));
+      return this;
+    }
+
+    public ArgumentsBuilder addOptionalArgument(Class<?> clazz, String name, Object defaultValue, String doc) {
```
hbs: Can you refactor this so the first 3 parameters are identical to the other signature, this will make it easier to add a default value afterwards.

warp10/src/main/java/io/warp10/script/formatted/FormattedWarpScriptFunction.java
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    if (null == optArgs) {
+      throw new WarpScriptException(getClass().getCanonicalName() + "'s method getOptionalArguments() returned null." +
+        " If no argument is expected, it should return an empty array instead.");
+    }
+
+    for (ArgumentSpecification arg: args) {
+      if (arg.isOptional()) {
+        throw new IllegalStateException("Output of " + getClass().getCanonicalName() + "'s method getArguments() must" +
```
hbs: Can you throw WarpScriptException

warp10/src/main/java/io/warp10/script/formatted/FormattedWarpScriptFunction.java
```diff
@@ -0,0 +1,305 @@
...
+          " only contain arguments without a default value.");
+      }
+    }
+
+    for (ArgumentSpecification arg: optArgs) {
+      if (!arg.isOptional()) {
+        throw new IllegalStateException("Output of " + getClass().getCanonicalName() + "'s method getArguments() must" +
```
hbs: Throw WarpScriptException

warp10/src/main/java/io/warp10/script/formatted/FormattedWarpScriptFunction.java
```diff
@@ -0,0 +1,305 @@
...
+      }
+    }
+
+    //
+    // If there are possible optional arguments, check that last mandatory argument is not a Map so that there is no
+    // confusion
+    //
+
+    if (args.size() > 1 && args.get(args.size() - 1) instanceof Map && optArgs.size() > 0) {
+      throw new IllegalStateException("In this case, the last non-optional argument can not be a Map so as to avoid any confusion when using optional arguments.");
```
hbs: Throw WarpScriptException

hbs: The message starting with 'In this case' is disturbing, it looks more like a comment in the code, can you make it clearer for the user receiving it out of context.

warp10/src/main/java/io/warp10/script/formatted/FormattedWarpScriptFunction.java
```diff
@@ -0,0 +1,305 @@
...
+
+    Map<String, Object> formattedArgs;
+
+    if (stack.peek() instanceof Map) {
+
+      //
+      // Case 1: A map is on top of the stack (some optional arguments may be given)
+      //
+
+      Map<String, Object> map = (Map) stack.peek();
```
hbs: This cast will throw an exception if stack.peek() is not a map with the expected types. You might want to perform a check first that all keys and values are of the expected types and throw an informative message if not.

warp10/src/main/java/io/warp10/script/formatted/FormattedWarpScriptFunction.java
```diff
@@ -0,0 +1,305 @@
...
+
+      //
+      // Check that non-optional args are contained in the map and that they have the correct type
+      //
+
+      for (ArgumentSpecification arg: args) {
+        if (!map.containsKey(arg.getName())) {
```
hbs: It is better to call map.get() and check that the returned value is not null, containsKey is more costly, and in case it is not null, you are using the returned value afterwards, you could save one call.

warp10/src/main/java/io/warp10/script/formatted/FormattedWarpScriptFunction.java
```diff
@@ -0,0 +1,305 @@
...
+            " be a " + arg.WarpScriptType() + ".");
+        }
+      }
+
+      //
+      // Check that optional args that are in the map are of the correct type
+      //
+
+      for (ArgumentSpecification arg: optArgs) {
+        if (map.containsKey(arg.getName()) && !arg.getClazz().isInstance(map.get(arg.getName()))) {
```
hbs: Ditto above, use map.get

warp10/src/main/java/io/warp10/script/formatted/FormattedWarpScriptFunction.java
```diff
@@ -0,0 +1,305 @@
...
+    if (i < 0) {
+      throw new IllegalStateException("Can not be negative");
+    }
+
+    if (0 == i) {
+      return "on top of the stack.";
+    } else if (1 == i) {
+      return "below the top of the stack.";
+    } else if (2 == i) {
+      return "on 3rd position counting from the top of the stack.";
```
hbs: We call those positions 'levels',  the top of the stack is also level 1, so if i == 2, you are at level 3, please use this terminology, i.e. stack level #

warp10/src/main/java/io/warp10/script/formatted/FormattedWarpScriptFunction.java
```diff
@@ -0,0 +1,305 @@
...
+    } else {
+      return "on " + (i + 1) + "th position counting from the top of the stack.";
```
hbs: Ditto

PR #439, opened 2019-05-07
Minor optimisations following 9240b90fd91dcc8a47727d879863182524924f3e

ftence:


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -1435,9 +1435,8 @@ public static final GeoTimeSerie subSerie(GeoTimeSerie gts, long starttimestamp,
     // Create sub GTS
     //
     
-    // The size hint impacts performance, choose it wisely...
     if (null == subgts) {
-      subgts = new GeoTimeSerie(128);
+      subgts = new GeoTimeSerie();
```
hbs: The argument to the constructor only sets a hint for growing the GTS, it is probably better if this is still the case.
The constructor does not allocate the arrays.

ftence: I am aware that it does not allocate the arrays. The reason I removed this is because it is the only place sizehint is set without any knowledge of how many data points could be added.

cityzendata:
The problem is that subserie is now used by timeclip and therefore the GTS
can be further augmented with other values, but now the hint will be unset
thus potentially leading to extraneous calls to provision, where the
arbitrary 128 value ensured this minimum

PR #443, opened 2019-05-20
Fix JSON formatting issue in EgressFetchHandler

KaseiFR:


warp10/src/main/java/io/warp10/continuum/egress/EgressFetchHandler.java
```diff
@@ -1385,7 +1385,7 @@ private static void jsonDump(PrintWriter pw, GTSDecoderIterator iter, long now,
           sb.append("}");
           sb.append(",\"i\":\"");
           sb.append(decoder.getLabelsId() & mask);
-          sb.append("\",\"la\":");
+          sb.append("\",\"la\":\"");
```
hbs: The last activity is a number, the trailing double quote was a left-over from the "i" field. Can you update your PR accordingly.

KaseiFR: Since `i` is also a `long` but was formatted a a string, I thought this was a future-proofing thing (some opaques ids)

hbs: 'i' is an id and the use of string was indeed to future proof things. 'la' is the last activity, it is a long and will remain a long.
The JSON returned by EgressExecHandler returns a long for la, therefore it is best to stick with that type for EgressFetchHandler.

hbs:
Thanks for the fix.

PR #444, opened 2019-05-23
bugfix subserie

randomboolean:
Commit 55be372 removed a sanity check l.1521 of GTSHelper in method subSerie().

This sanity check did prevent two cases of ticks being added to a subgts:
-  when stoptimestamp < starttimestamp
- if stoptimestamp is not present in gts, lastidx was wrong by +1

We fix those bugs


hbs:
Can you check that this is actually already fixed.

randomboolean:
Second point was already fixed

PR #449, opened 2019-06-10
Binary support

hbs:
Add support for storing byte arrays natively as values of GTS in the Warp 10 Storage Engine.

Added support for byte arrays in GTSEncoder.

Note that when using GTS in WarpScript, no byte arrays will be manipulated, instead byte arrays will be converted to STRING using ISO-8859-1


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -2363,6 +2369,10 @@ public static Object parseValue(String valuestr) throws ParseException {
         }
         
         value = TOQUATERNION.toQuaternion(q[0], q[1], q[2], q[3]);
+      } else if ('b' == firstChar && valuestr.startsWith("b64:")) {
+        value = Base64.decodeBase64(valuestr.substring(4));
+      } else if ('h' == firstChar && valuestr.startsWith("hex:")) {
+        value = Hex.decodeHex(valuestr.substring(4).toCharArray());
```
ftence: Remove FIXME line 2334.

warp10/src/main/java/io/warp10/continuum/ThrottlingConfigGenerator.java
```diff
@@ -305,6 +305,7 @@ public static void genConfiguration(String baseUrl, String warpEndPoint, String
       GTSDecoder decoder = encoder.getDecoder();
 
       if (decoder.next()) {        
+        // We don't call getBinaryValue because we know the value was stored as OPB64
```
ftence: Same comment for line 428.

warp10/src/main/java/io/warp10/udf/TUNWRAP.java
```diff
@@ -150,6 +150,7 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
           
           long location = decoder.getLocation();
           long elevation = decoder.getElevation();
+          // We are populating a GTS, so we do not call getBinaryValue
```
ftence: Add same comment in continuum/gts/GTSDecoder, line 544.

warp10/src/test/java/io/warp10/continuum/gts/GTSEncoderTest.java
```diff
@@ -613,4 +616,122 @@ public void testResetSTRING() throws Exception {
...
+    encoder.reset(decoder.getEncoder());
+    
+    decoder = encoder.getDecoder();
+    
+    decoder.next();
+    
+    Assert.assertEquals(1.0D, decoder.getTimestamp(), 0.000000000001D);
+    Object value = decoder.getBinaryValue();
+    Assert.assertTrue(value instanceof byte[]);    
+    Assert.assertTrue(1 == ((byte[]) value).length);
```
ftence: Use assertEquals.

warp10/src/test/java/io/warp10/continuum/gts/GTSEncoderTest.java
```diff
@@ -613,4 +616,122 @@ public void testResetSTRING() throws Exception {
...
+    
+    bytes = encoder.getBytes();
+    Assert.assertEquals(3, bytes.length);
+
+    GTSDecoder decoder = enc2.getDecoder();
+    
+    Assert.assertTrue(decoder.next());
+    Assert.assertTrue(0L == decoder.getTimestamp());
+    Assert.assertTrue(decoder.getBinaryValue() instanceof byte[]);
+    Assert.assertTrue("é".equals(decoder.getValue()));
```
ftence: Use assertEquals, also 2 lines above.

warp10/src/main/java/io/warp10/script/StackUtils.java
```diff
@@ -212,6 +212,7 @@ public static void objectToJSON(JsonSerializer serializer, PrintWriter out, Obje
         long ts = decoder.getTimestamp();
         long location = decoder.getLocation();
         long elevation = decoder.getElevation();
+        // We do not call getBinaryValue because JSON cannot represent byte arrays
```
ftence: May result in an invalid JSON, see this WarpScript:
```
NEWENCODER
1 1 1 1 '00' HEX-> ADDVALUE
```
Problem is that for GTSs, see line 167, we don't know that if a string represents a byte[] so it won't be straightforward to find a solution.

warp10/src/main/java/io/warp10/continuum/egress/EgressFetchHandler.java
```diff
@@ -1441,6 +1446,8 @@ static void jsonDump(PrintWriter pw, Iterator<GTSDecoder> iter, long now, long t
             pw.print(decoder.getElevation());
           }
           pw.print(",");
+          // For JSON representation we do not extract the binary value as byte[] cannot be
```
ftence: May result in invalid JSON, see next comment for more details.

warp10/src/main/java/io/warp10/standalone/StandalonePlasmaHandler.java
```diff
@@ -689,6 +689,7 @@ protected void dispatch(GTSEncoder encoder) throws IOException {
                 json.put("l", labels);              
                 
                 json.put("t", decoder.getTimestamp());
+                // Requested format is JSON so we do not use getBinaryValue as JSON cannot represent byte arrays
```
ftence: Ditto.

hbs:
The JSON representation will be perfectly valid as the binary content will be converted to a String. It is part of the contract that the user must know which values were stored as binary so the correct charset (ISO8859-1) can be used for converting the string value to a byte array.

PR #452, opened 2019-06-11
Abstract class to be inherited for WS function which can be applied on lists or element

ftence:
This class does not use a Map for parameters unlike GTSStackFunction so it is a tad faster.
This is also more flexible because it can be used on any type.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/ElementOrListStackFunction.java
+++ b/warp10/src/main/java/io/warp10/script/ElementOrListStackFunction.java
@@ -1,0 +1,42 @@
+package io.warp10.script;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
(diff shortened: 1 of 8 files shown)


warp10/src/main/java/io/warp10/script/ElementOrListStackFunction.java
```diff
@@ -0,0 +1,79 @@
...
+public abstract class ElementOrListStackFunction extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+
+  /**
+   * Interface defining the function to be generated which is applied to each element.
+   * This method should check the type of the given element and throw a WarpScriptException if the element is of
+   * unexpected type.
+   */
+  @FunctionalInterface
+  public interface ElementStackFunction {
+    Object applyOnElement(Object element) throws WarpScriptException;
```
hbs: Can you make it public, this is the modifier you use elsewhere, better use it in the interface definition too.

PR #456, opened 2019-06-11
boolean, bitwise and comparison operators on GTS

pi-r-p:
Fix the fate of boolean GTS operation in Warp 10 2.0 : allow not, and, or, xor on boolean GTS.
Tried to optimize NOT with flip operation, but I didn't test performances.

I choose to implement and or xor bitwise operators, not condition ones... I can change.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -10270,4 +10270,8 @@ public class GTSHelper {
   public static double skewness(GeoTimeSerie gts, boolean bessel) throws WarpScriptException {
     return standardizedMoment(3, gts, bessel);
   }
+
+  public static void booleanNot(GeoTimeSerie gts) {
+    gts.booleanValues.flip(0, gts.booleanValues.length());
+  }
 }
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -10271,7 +10271,11 @@ public class GTSHelper {
     return standardizedMoment(3, gts, bessel);
   }
 
-  public static void booleanNot(GeoTimeSerie gts) {
-    gts.booleanValues.flip(0, gts.booleanValues.length());
+  public static void booleanNot(GeoTimeSerie gts) throws WarpScriptException {
+    if (GeoTimeSerie.TYPE.BOOLEAN == gts.getType()) {
+      gts.booleanValues.flip(0, gts.booleanValues.length());
+    } else {
+      throw new WarpScriptException("Non boolean Geo Time Series.");
+    }
   }
 }
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSOpsHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSOpsHelper.java
@@ -20,8 +20,10 @@ public class GTSOpsHelper {
   public static interface GTSBinaryOp {
     public Object op(GeoTimeSerie gtsa, GeoTimeSerie gtsb, int idxa, int idxb);
   }
-    
   public static void applyBinaryOp(GeoTimeSerie result, GeoTimeSerie gts1, GeoTimeSerie gts2, GTSBinaryOp op) {
+    applyBinaryOp(result, gts1, gts2, op,false);
+  }
+  public static void applyBinaryOp(GeoTimeSerie result, GeoTimeSerie gts1, GeoTimeSerie gts2, GTSBinaryOp op, boolean copyGts1Location) {
     //
     // Determine if result should be bucketized or not
     //
@@ -49,7 +51,7 @@ public class GTSOpsHelper {
     // Sweeping line over the timestamps
     int idxa = 0;
     int idxb = 0;
-             
+    
     int na = GTSHelper.nvalues(gts1);
     int nb = GTSHelper.nvalues(gts2);
     
@@ -62,7 +64,9 @@ public class GTSOpsHelper {
     if (idxb < na) {
       tsb = GTSHelper.tickAtIndex(gts2, idxb);
     }
-
+  
+    Object value;
+  
     while(idxa < na || idxb < nb) {
       if (idxa >= na) {
         tsa = null;
@@ -74,7 +78,14 @@ public class GTSOpsHelper {
         // We have values at the current index for both GTS
         if (0 == tsa.compareTo(tsb)) {
           // Both indices indicate the same timestamp
-          GTSHelper.setValue(result, tsa, op.op(gts1, gts2, idxa, idxb));
+          if (copyGts1Location) {
+            value = op.op(gts1, gts2, idxa, idxb);
+            if (null != value) {
+              GTSHelper.setValue(result, tsa,GTSHelper.locationAtIndex(gts1,idxa),GTSHelper.elevationAtIndex(gts1,idxa),value,false);
+            }
+          } else {
+            GTSHelper.setValue(result, tsa, op.op(gts1, gts2, idxa, idxb));
+          }
           // Advance both indices
           idxa++;
           idxb++;
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSOpsHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSOpsHelper.java
@@ -64,9 +64,7 @@ public class GTSOpsHelper {
     if (idxb < na) {
       tsb = GTSHelper.tickAtIndex(gts2, idxb);
     }
-  
-    Object value;
-  
+    
     while(idxa < na || idxb < nb) {
       if (idxa >= na) {
         tsa = null;
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSOpsHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSOpsHelper.java
@@ -79,10 +79,7 @@ public class GTSOpsHelper {
         if (0 == tsa.compareTo(tsb)) {
           // Both indices indicate the same timestamp
           if (copyGts1Location) {
-            value = op.op(gts1, gts2, idxa, idxb);
-            if (null != value) {
-              GTSHelper.setValue(result, tsa,GTSHelper.locationAtIndex(gts1,idxa),GTSHelper.elevationAtIndex(gts1,idxa),value,false);
-            }
+            GTSHelper.setValue(result, tsa,GTSHelper.locationAtIndex(gts1,idxa),GTSHelper.elevationAtIndex(gts1,idxa),op.op(gts1, gts2, idxa, idxb),false);
           } else {
             GTSHelper.setValue(result, tsa, op.op(gts1, gts2, idxa, idxb));
           }
--- a/warp10/src/main/java/io/warp10/script/binary/BitwiseAND.java
+++ b/warp10/src/main/java/io/warp10/script/binary/BitwiseAND.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018  SenX S.A.S.
+//   Copyright 2019  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
--- a/warp10/src/main/java/io/warp10/script/binary/BitwiseAND.java
+++ b/warp10/src/main/java/io/warp10/script/binary/BitwiseAND.java
@@ -27,59 +27,15 @@ import io.warp10.script.WarpScriptStack;
 /**
  * Bitwise AND of the two operands on top of the stack
  */
-public class BitwiseAND extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+public class BitwiseAND extends BitwiseOperation {
 
   public BitwiseAND(String name) {
     super(name);
   }
-  
+
   @Override
-  public Object apply(WarpScriptStack stack) throws WarpScriptException {
-    Object op2 = stack.pop();
-    Object op1 = stack.pop();
-    
-    if (op2 instanceof Long && op1 instanceof Long) {
-      stack.push(((Long) op1).longValue() & ((Long) op2).longValue());
-    } else if (op1 instanceof GeoTimeSerie && op2 instanceof GeoTimeSerie) {
-      GeoTimeSerie gts1 = (GeoTimeSerie) op1;
-      GeoTimeSerie gts2 = (GeoTimeSerie) op2;
-      if (GeoTimeSerie.TYPE.LONG == gts1.getType() && GeoTimeSerie.TYPE.LONG == gts2.getType()) {
-        GeoTimeSerie result = new GeoTimeSerie(Math.max(GTSHelper.nvalues(gts1), GTSHelper.nvalues(gts2)));
-        result.setType(GeoTimeSerie.TYPE.LONG);
-        GTSOpsHelper.GTSBinaryOp op = new GTSOpsHelper.GTSBinaryOp() {
-          @Override
-          public Object op(GeoTimeSerie gtsa, GeoTimeSerie gtsb, int idxa, int idxb) {
-            return ((Long) GTSHelper.valueAtIndex(gtsa, idxa)) & ((Long) GTSHelper.valueAtIndex(gtsb, idxb));
-          }
-        };
-        GTSOpsHelper.applyBinaryOp(result, gts1, gts2, op);
-        stack.push(result);
-      } else {
-        throw new WarpScriptException(getName() + " can only operate on long values or long GTS.");
-      }
-    } else if (op1 instanceof GeoTimeSerie || op2 instanceof GeoTimeSerie) {
-      GeoTimeSerie gts;
-      Long mask;
-      if (op1 instanceof GeoTimeSerie && op2 instanceof Long) {
-        gts = (GeoTimeSerie)op1;
-        mask = (Long)op2;
-      } else if (op2 instanceof GeoTimeSerie && op1 instanceof Long) {
-        gts = (GeoTimeSerie)op2;
-        mask = (Long)op1;
-      } else {
-        throw new WarpScriptException(getName() + " can only operate two GTS or one GTS and a long value.");
-      }
-      GeoTimeSerie result = gts.cloneEmpty();
-      result.setType(GeoTimeSerie.TYPE.LONG);
-      for (int i = 0; i < GTSHelper.nvalues(gts); i++) {
-        GTSHelper.setValue(result, GTSHelper.tickAtIndex(gts, i), GTSHelper.locationAtIndex(gts, i), GTSHelper.elevationAtIndex(gts, i),
-                mask & ((Number)GTSHelper.valueAtIndex(gts,i)).longValue() , false);
-      }
-      stack.push(result);
-    } else {
-      throw new WarpScriptException(getName() + " can only operate on long values or boolean GTS.");
-    }
-    
-    return stack;
+  public long operator(long op1, long op2) {
+    return op1 & op2;
   }
+
 }
--- a/warp10/src/main/java/io/warp10/script/binary/BitwiseAND.java
+++ b/warp10/src/main/java/io/warp10/script/binary/BitwiseAND.java
@@ -43,20 +43,39 @@ public class BitwiseAND extends NamedWarpScriptFunction implements WarpScriptSta
     } else if (op1 instanceof GeoTimeSerie && op2 instanceof GeoTimeSerie) {
       GeoTimeSerie gts1 = (GeoTimeSerie) op1;
       GeoTimeSerie gts2 = (GeoTimeSerie) op2;
-      if (GeoTimeSerie.TYPE.BOOLEAN == gts1.getType() && GeoTimeSerie.TYPE.BOOLEAN == gts2.getType()) {
+      if (GeoTimeSerie.TYPE.LONG == gts1.getType() && GeoTimeSerie.TYPE.LONG == gts2.getType()) {
         GeoTimeSerie result = new GeoTimeSerie(Math.max(GTSHelper.nvalues(gts1), GTSHelper.nvalues(gts2)));
-        result.setType(GeoTimeSerie.TYPE.BOOLEAN);
+        result.setType(GeoTimeSerie.TYPE.LONG);
         GTSOpsHelper.GTSBinaryOp op = new GTSOpsHelper.GTSBinaryOp() {
           @Override
           public Object op(GeoTimeSerie gtsa, GeoTimeSerie gtsb, int idxa, int idxb) {
-          return ((Boolean) GTSHelper.valueAtIndex(gtsa, idxa)) && ((Boolean) GTSHelper.valueAtIndex(gtsb, idxb));
+            return ((Long) GTSHelper.valueAtIndex(gtsa, idxa)) & ((Long) GTSHelper.valueAtIndex(gtsb, idxb));
           }
         };
         GTSOpsHelper.applyBinaryOp(result, gts1, gts2, op);
         stack.push(result);
       } else {
-        throw new WarpScriptException(getName() + " can only operate on long values or boolean GTS.");
+        throw new WarpScriptException(getName() + " can only operate on long values or long GTS.");
       }
+    } else if (op1 instanceof GeoTimeSerie || op2 instanceof GeoTimeSerie) {
+      GeoTimeSerie gts;
+      Long mask;
+      if (op1 instanceof GeoTimeSerie && op2 instanceof Long) {
+        gts = (GeoTimeSerie)op1;
+        mask = (Long)op2;
+      } else if (op2 instanceof GeoTimeSerie && op1 instanceof Long) {
+        gts = (GeoTimeSerie)op2;
+        mask = (Long)op1;
+      } else {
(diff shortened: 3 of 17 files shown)


hbs:
Apart from NOT, you modified the bitwise AND/OR/XOR operations which is misleading as your modifications act like logical operations.

Please modify this PR so it makes the logical AND/OR/XOR ops operate on Geo Time Series in the way +,-,/,* allow to perform arithmetics on GTS.

This means that they should also work when one of the operand is a boolean.

pi-r-p:
update : 
`booleanGTS ! `   : usefull, could replace mapper.not
`booleanGTS booleanGTS2 && `   : usefull
`booleanGTS booleanGTS2 || `   : usefull
`longGTS longGTS2  & | ^ ` : bitwise on long values. could be usefull.
`longGTS long  & | ^ ` : bitwise on long values. could be usefull for static masks.

`booleanGTS false &&` : useless. same as `false mapper.replace`. not implemented in this PR

warp10/src/main/java/io/warp10/script/binary/BitwiseAND.java
```diff
@@ -37,8 +40,44 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
     
     if (op2 instanceof Long && op1 instanceof Long) {
       stack.push(((Long) op1).longValue() & ((Long) op2).longValue());
+    } else if (op1 instanceof GeoTimeSerie && op2 instanceof GeoTimeSerie) {
+      GeoTimeSerie gts1 = (GeoTimeSerie) op1;
+      GeoTimeSerie gts2 = (GeoTimeSerie) op2;
+      if (GeoTimeSerie.TYPE.LONG == gts1.getType() && GeoTimeSerie.TYPE.LONG == gts2.getType()) {
+        GeoTimeSerie result = new GeoTimeSerie(Math.max(GTSHelper.nvalues(gts1), GTSHelper.nvalues(gts2)));
+        result.setType(GeoTimeSerie.TYPE.LONG);
+        GTSOpsHelper.GTSBinaryOp op = new GTSOpsHelper.GTSBinaryOp() {
```
hbs: It is better if you allocate this in the class once and for all instead of creating a new instance each time apply is called.

warp10/src/main/java/io/warp10/script/binary/BitwiseAND.java
```diff
@@ -37,8 +40,44 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+          }
+        };
+        GTSOpsHelper.applyBinaryOp(result, gts1, gts2, op);
+        stack.push(result);
+      } else {
+        throw new WarpScriptException(getName() + " can only operate on long values or long GTS.");
+      }
+    } else if (op1 instanceof GeoTimeSerie || op2 instanceof GeoTimeSerie) {
+      GeoTimeSerie gts;
+      Long mask;
```
hbs: Use the primitive type long

warp10/src/main/java/io/warp10/script/binary/BitwiseAND.java
```diff
@@ -37,8 +40,44 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      if (op1 instanceof GeoTimeSerie && op2 instanceof Long) {
+        gts = (GeoTimeSerie)op1;
+        mask = (Long)op2;
```
hbs: use ((Number) op2).longValue()

warp10/src/main/java/io/warp10/script/binary/BitwiseAND.java
```diff
@@ -37,8 +40,44 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      } else if (op2 instanceof GeoTimeSerie && op1 instanceof Long) {
+        gts = (GeoTimeSerie)op2;
+        mask = (Long)op1;
```
hbs: ditto above

warp10/src/main/java/io/warp10/script/binary/BitwiseAND.java
```diff
@@ -37,8 +40,44 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      } else {
+        throw new WarpScriptException(getName() + " can only operate two GTS or one GTS and a long value.");
+      }
+      GeoTimeSerie result = gts.cloneEmpty();
+      result.setType(GeoTimeSerie.TYPE.LONG);
+      for (int i = 0; i < GTSHelper.nvalues(gts); i++) {
+        GTSHelper.setValue(result, GTSHelper.tickAtIndex(gts, i), GTSHelper.locationAtIndex(gts, i), GTSHelper.elevationAtIndex(gts, i),
+                mask & ((Number)GTSHelper.valueAtIndex(gts,i)).longValue() , false);
```
hbs: It's more elegant to use & mask rather than mask &

warp10/src/main/java/io/warp10/script/binary/BitwiseOR.java
```diff
@@ -37,6 +40,42 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
     
     if (op2 instanceof Long && op1 instanceof Long) {
       stack.push(((Long) op1).longValue() | ((Long) op2).longValue());
+    } else if (op1 instanceof GeoTimeSerie && op2 instanceof GeoTimeSerie) {
+      GeoTimeSerie gts1 = (GeoTimeSerie) op1;
+      GeoTimeSerie gts2 = (GeoTimeSerie) op2;
+      if (GeoTimeSerie.TYPE.LONG == gts1.getType() && GeoTimeSerie.TYPE.LONG == gts2.getType()) {
+        GeoTimeSerie result = new GeoTimeSerie(Math.max(GTSHelper.nvalues(gts1), GTSHelper.nvalues(gts2)));
+        result.setType(GeoTimeSerie.TYPE.LONG);
+        GTSOpsHelper.GTSBinaryOp op = new GTSOpsHelper.GTSBinaryOp() {
```
hbs: Allocate once

warp10/src/main/java/io/warp10/script/binary/BitwiseOR.java
```diff
@@ -37,6 +40,42 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+          }
+        };
+        GTSOpsHelper.applyBinaryOp(result, gts1, gts2, op);
+        stack.push(result);
+      } else {
+        throw new WarpScriptException(getName() + " can only operate on long values or long GTS.");
+      }
+    } else if (op1 instanceof GeoTimeSerie || op2 instanceof GeoTimeSerie) {
+      GeoTimeSerie gts;
+      Long mask;
```
hbs: long

warp10/src/main/java/io/warp10/script/binary/BitwiseOR.java
```diff
@@ -37,6 +40,42 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      if (op1 instanceof GeoTimeSerie && op2 instanceof Long) {
+        gts = (GeoTimeSerie)op1;
+        mask = (Long)op2;
```
hbs: ((Number) op2).longValue();

warp10/src/main/java/io/warp10/script/binary/BitwiseXOR.java
```diff
@@ -37,6 +40,42 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
     
     if (op2 instanceof Long && op1 instanceof Long) {
       stack.push(((Long) op1).longValue() ^ ((Long) op2).longValue());
+    } else if (op1 instanceof GeoTimeSerie && op2 instanceof GeoTimeSerie) {
+      GeoTimeSerie gts1 = (GeoTimeSerie) op1;
+      GeoTimeSerie gts2 = (GeoTimeSerie) op2;
+      if (GeoTimeSerie.TYPE.LONG == gts1.getType() && GeoTimeSerie.TYPE.LONG == gts2.getType()) {
+        GeoTimeSerie result = new GeoTimeSerie(Math.max(GTSHelper.nvalues(gts1), GTSHelper.nvalues(gts2)));
+        result.setType(GeoTimeSerie.TYPE.LONG);
+        GTSOpsHelper.GTSBinaryOp op = new GTSOpsHelper.GTSBinaryOp() {
```
hbs: ditto

warp10/src/main/java/io/warp10/script/binary/CondShortCircuit.java
```diff
@@ -72,13 +75,32 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+        stack.push(operator((boolean) op1, (boolean) op2));
         return stack;
+      } else if (op1 instanceof GeoTimeSerie && op2 instanceof GeoTimeSerie) {
+        // logical operator between boolean GTS values
+        GeoTimeSerie gts1 = (GeoTimeSerie) op1;
+        GeoTimeSerie gts2 = (GeoTimeSerie) op2;
+        if (GeoTimeSerie.TYPE.BOOLEAN == gts1.getType() && GeoTimeSerie.TYPE.BOOLEAN == gts2.getType()) {
+          GeoTimeSerie result = new GeoTimeSerie(Math.max(GTSHelper.nvalues(gts1), GTSHelper.nvalues(gts2)));
+          result.setType(GeoTimeSerie.TYPE.BOOLEAN);
+          GTSOpsHelper.GTSBinaryOp op = new GTSOpsHelper.GTSBinaryOp() {
```
hbs: ditto

pi-r-p:
Refactored with changes. Is it OK with the GTSOpsHelper.GTSBinaryOp allocation ?

pi-r-p:
Need more unit tests to be sure everything is right... I'll comment again when it's done.

warp10/src/main/java/io/warp10/script/binary/BitwiseOperation.java
```diff
@@ -0,0 +1,89 @@
...
+  private GTSOpsHelper.GTSBinaryOp op;
+
+  public abstract long operator(long op1, long op2);
+
+  public BitwiseOperation(String name) {
+    super(name);
+    this.op = new GTSOpsHelper.GTSBinaryOp() {
+      @Override
+      public Object op(GeoTimeSerie gtsa, GeoTimeSerie gtsb, int idxa, int idxb) {
+        return operator(((Long) GTSHelper.valueAtIndex(gtsa, idxa)) , ((Long) GTSHelper.valueAtIndex(gtsb, idxb)));
```
hbs: Use ((Number) xxx).longValue()

pi-r-p:
Fixed long cast, unit tests and doc ready.

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -10270,4 +10270,8 @@ public static double kurtosis(GeoTimeSerie gts, boolean bessel) throws WarpScrip
   public static double skewness(GeoTimeSerie gts, boolean bessel) throws WarpScriptException {
     return standardizedMoment(3, gts, bessel);
   }
+
+  public static void booleanNot(GeoTimeSerie gts) {
+    gts.booleanValues.flip(0, gts.booleanValues.length());
```
hbs: Should check that the GTS is indeed a boolean one, otherwise gts.booleanValues might be null and this could lead to a NPE

warp10/src/main/java/io/warp10/script/binary/BitwiseOperation.java
```diff
@@ -0,0 +1,84 @@
...
+import io.warp10.continuum.gts.GeoTimeSerie;
+import io.warp10.script.NamedWarpScriptFunction;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.script.WarpScriptStackFunction;
+import io.warp10.standalone.Warp;
+
+public abstract class BitwiseOperation extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+
+  private GTSOpsHelper.GTSBinaryOp op;
```
hbs: Make it final

warp10/src/main/java/io/warp10/script/binary/BitwiseOperation.java
```diff
@@ -0,0 +1,84 @@
...
+      stack.push(operator(((Long) op1).longValue(), ((Long) op2).longValue()));
+    } else if (op1 instanceof GeoTimeSerie && op2 instanceof GeoTimeSerie) {
+      GeoTimeSerie gts1 = (GeoTimeSerie) op1;
+      GeoTimeSerie gts2 = (GeoTimeSerie) op2;
+      if (GeoTimeSerie.TYPE.LONG == gts1.getType() && GeoTimeSerie.TYPE.LONG == gts2.getType()) {
+        GeoTimeSerie result = new GeoTimeSerie(Math.max(GTSHelper.nvalues(gts1), GTSHelper.nvalues(gts2)));
+        result.setType(GeoTimeSerie.TYPE.LONG);
+        GTSOpsHelper.applyBinaryOp(result, gts1, gts2, op);
+        stack.push(result);
+      } else {
```
hbs: Should we tolerate the case of applying the op to empty GTS? If so you need to check that type is not UNDEFINED before throwing the exception

warp10/src/main/java/io/warp10/script/binary/BitwiseOperation.java
```diff
@@ -0,0 +1,84 @@
...
+        throw new WarpScriptException(exceptionMessage);
+      }
+    } else if (op1 instanceof GeoTimeSerie && op2 instanceof Long) {
+      GeoTimeSerie  gts = (GeoTimeSerie)op1;
+      long mask = ((Number)op2).longValue();
+      if (GeoTimeSerie.TYPE.LONG == gts.getType()){
```
hbs: ditto above to support empty GTS

warp10/src/main/java/io/warp10/script/binary/CondShortCircuit.java
```diff
@@ -34,16 +37,24 @@
 
   protected final boolean triggerValue;
 
+  private GTSOpsHelper.GTSBinaryOp op;
```
hbs: Make it final

warp10/src/main/java/io/warp10/script/binary/CondShortCircuit.java
```diff
@@ -72,13 +83,26 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+        GeoTimeSerie gts1 = (GeoTimeSerie) op1;
+        GeoTimeSerie gts2 = (GeoTimeSerie) op2;
+        if (GeoTimeSerie.TYPE.BOOLEAN == gts1.getType() && GeoTimeSerie.TYPE.BOOLEAN == gts2.getType()) {
+          GeoTimeSerie result = new GeoTimeSerie(Math.max(GTSHelper.nvalues(gts1), GTSHelper.nvalues(gts2)));
+          result.setType(GeoTimeSerie.TYPE.BOOLEAN);
+          GTSOpsHelper.applyBinaryOp(result, gts1, gts2, this.op);
+          stack.push(result);
+          return stack;
+        } else {
+          throw new WarpScriptException(getName() + " can only operate on long values or long GTS.");
```
hbs: Support empty GTS?

warp10/src/main/java/io/warp10/script/unary/NOT.java
```diff
@@ -36,8 +38,18 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      GeoTimeSerie gts = (GeoTimeSerie) op;
+      if (GeoTimeSerie.TYPE.BOOLEAN == gts.getType()) {
+        GeoTimeSerie result = gts.clone();
+        GTSHelper.booleanNot(result);
+        stack.push(result);
+      } else {
+        throw new WarpScriptException(getName() + " can only operate on a boolean value or a boolean GTS.");
+      }
+    }
+    else {
```
hbs: Extraneous LF

warp10/src/main/java/io/warp10/script/unary/NOT.java
```diff
@@ -36,8 +38,18 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
       stack.push(!((Boolean) op).booleanValue());
-    } else {
-      throw new WarpScriptException(getName() + " can only operate on a boolean value.");
+    } else if (op instanceof GeoTimeSerie) {
+      GeoTimeSerie gts = (GeoTimeSerie) op;
+      if (GeoTimeSerie.TYPE.BOOLEAN == gts.getType()) {
+        GeoTimeSerie result = gts.clone();
+        GTSHelper.booleanNot(result);
+        stack.push(result);
+      } else {
```
hbs: Support empty GTS?

pi-r-p:
- Fixed empty gts problem, bitwise and conditionnal operations should now be OK.
- Implements GT LT GE LE comparison for GTS:
  - $g1 number > : equivalent to mapper.gt
  - $g1 $g2 > : for each common tick, if g1 value > g2 value, keep g1 value. 
- Should EQ be implemented the same way, that is the question.

hbs:
What happens when performing $g1 $g2 > for the ticks which are not in common? Is there no value being produced?

warp10/src/main/java/io/warp10/script/binary/BitwiseOperation.java
```diff
@@ -0,0 +1,89 @@
...
+        GeoTimeSerie result = gts.cloneEmpty();
+        result.setType(GeoTimeSerie.TYPE.LONG);
+        for (int i = 0; i < GTSHelper.nvalues(gts); i++) {
+          GTSHelper.setValue(result, GTSHelper.tickAtIndex(gts, i), GTSHelper.locationAtIndex(gts, i), GTSHelper.elevationAtIndex(gts, i),
+              operator(((Number) GTSHelper.valueAtIndex(gts, i)).longValue(), mask), false);
+        }
+        stack.push(result);
+      } else if (GeoTimeSerie.TYPE.UNDEFINED == gts.getType()) {
+        // gts is empty return gts
+        stack.push(gts);
```
hbs: A new empty GTS (cloneEmpty) should be returned instead of returning the original one

warp10/src/main/java/io/warp10/script/binary/ComparisonOperation.java
```diff
@@ -0,0 +1,152 @@
...
+          }
+        };
+        GTSOpsHelper.applyBinaryOp(result, gts1, gts2, op);
+        stack.push(result);
+      } else {
+        throw new WarpScriptException(exceptionMessage);
+      }
+    } else if (op1 instanceof GeoTimeSerie && GeoTimeSerie.TYPE.UNDEFINED == ((GeoTimeSerie) op1).getType() && (op2 instanceof String || op1 instanceof Number)) {
+      // empty string compared to a string or a number
+      stack.push(op1);
```
hbs: This returns the original GTS, shouldn't we return a new one?

warp10/src/main/java/io/warp10/script/binary/ComparisonOperation.java
```diff
@@ -0,0 +1,152 @@
...
+      GeoTimeSerie gts1 = (GeoTimeSerie) op1;
+      GeoTimeSerie gts2 = (GeoTimeSerie) op2;
+      if (GeoTimeSerie.TYPE.UNDEFINED == gts1.getType() || GeoTimeSerie.TYPE.UNDEFINED == gts2.getType()) {
+        // gts1 or gts2 empty, return an empty gts
+        stack.push(new GeoTimeSerie());
+      } else if (GeoTimeSerie.TYPE.STRING == gts1.getType() && GeoTimeSerie.TYPE.STRING == gts2.getType()) {
+        // both strings, compare lexicographically
+        GeoTimeSerie result = new GeoTimeSerie(Math.max(GTSHelper.nvalues(gts1), GTSHelper.nvalues(gts2)));
+        result.setType(GeoTimeSerie.TYPE.STRING);
+        GTSOpsHelper.GTSBinaryOp op = new GTSOpsHelper.GTSBinaryOp() {
```
hbs: This should be created once only

warp10/src/main/java/io/warp10/script/unary/NOT.java
```diff
@@ -36,8 +38,20 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
       stack.push(!((Boolean) op).booleanValue());
+    } else if (op instanceof GeoTimeSerie) {
+      GeoTimeSerie gts = (GeoTimeSerie) op;
+      if (GeoTimeSerie.TYPE.BOOLEAN == gts.getType()) {
+        GeoTimeSerie result = gts.clone();
+        GTSHelper.booleanNot(result);
+        stack.push(result);
+      } else if (GeoTimeSerie.TYPE.UNDEFINED == gts.getType()) {
+        // empty gts
+        stack.push(gts);
```
hbs: Should return a new GTS instead of the original one?

pi-r-p:
Unit tests done, all is clean. 
Operations on NaN have exactly the same behaviour as previous version.

warp10/src/main/java/io/warp10/continuum/gts/GTSOpsHelper.java
```diff
@@ -62,7 +64,9 @@ public static void applyBinaryOp(GeoTimeSerie result, GeoTimeSerie gts1, GeoTime
     if (idxb < na) {
       tsb = GTSHelper.tickAtIndex(gts2, idxb);
     }
-
+  
+    Object value;
```
hbs: Is that useful?

warp10/src/main/java/io/warp10/script/binary/ComparisonOperation.java
```diff
@@ -0,0 +1,215 @@
...
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.script.WarpScriptStackFunction;
+
+public abstract class ComparisonOperation extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+  
+  public abstract boolean operator(int op1, int op2);
+  
+  //default behavior for > < operators. For >= <= == or !=, you need to set these.
+  public boolean ifOneNaNOperand = false;
```
hbs: Does it need to be public?

Shouldn't it be final?

warp10/src/main/java/io/warp10/continuum/gts/GTSOpsHelper.java
pi-r-p: My mistake.

warp10/src/main/java/io/warp10/script/binary/ComparisonOperation.java
pi-r-p: It needs to be set by subclasses, such NE or EQ or GE or LE. It could be protected instead of public.

pi-r-p: Fixed with a different constructor.

warp10/src/main/java/io/warp10/script/binary/BitwiseOperation.java
```diff
@@ -0,0 +1,89 @@
...
+        GTSOpsHelper.applyBinaryOp(result, gts1, gts2, op);
+        stack.push(result);
+      } else if ((GeoTimeSerie.TYPE.UNDEFINED == gts1.getType() && GeoTimeSerie.TYPE.LONG == gts2.getType())
+          || (GeoTimeSerie.TYPE.UNDEFINED == gts2.getType() && GeoTimeSerie.TYPE.LONG == gts1.getType())) {
+        // gts1 or gts2 empty, return an empty gts
+        stack.push(new GeoTimeSerie());
+      } else {
+        throw new WarpScriptException(exceptionMessage);
+      }
+    } else if (op1 instanceof GeoTimeSerie && op2 instanceof Long) {
```
hbs: What about the case where op1 is a Long and op2 a GTS?

warp10/src/main/java/io/warp10/script/binary/ComparisonOperation.java
```diff
@@ -0,0 +1,233 @@
...
+        stack.push(ifOneNaNOperand ? gts.clone() : result);
+      } else {
+        for (int i = 0; i < GTSHelper.nvalues(gts); i++) {
+          GTSHelper.setValue(result, GTSHelper.tickAtIndex(gts, i), GTSHelper.locationAtIndex(gts, i), GTSHelper.elevationAtIndex(gts, i),
+              operator(EQ.compare((Number) GTSHelper.valueAtIndex(gts, i), (Number) op2), 0) ? GTSHelper.valueAtIndex(gts, i) : null, false);
+        }
+        stack.push(result);
+      }
+    } else {
+      throw new WarpScriptException(getName() + " can only operate when GTS values and the top stack operand have the same type. Booleans are not supported.");
```
hbs: What about the case where op2 is a GTS and op1 a scalar?

warp10/src/main/java/io/warp10/script/binary/BitwiseOperation.java
pi-r-p: Signature is restricted to `GTS op GTS` or `GTS op (number or boolean)` everywhere. It allow the retro compatibility with strange but legal operations that could be done before 2.1 version, such as : 
```
true NEWGTS == ! ASSERT //also legal before 2.1
```

pi-r-p: It also simplify the understanding of non really commutative operations such as bitwise operations like `GTS 3 <<`

hbs:
Can you mark as resolved all the open comments you addressed

PR #465, opened 2019-06-17
Add a method to fully sort a GTS using tick, value, location, elevation.

ftence:
Makes the equality between GTSs much easier.
No WarpScript function exposing this yet.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -685,8 +685,7 @@ public class GTSHelper {
 
   /**
    * Sort GTS according to location, using HHCodes, between two indexes.
-   * The ticks with no locations are clustered somewhere in-between those with locations since the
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -781,7 +806,174 @@ public static GeoTimeSerie locationSort(GeoTimeSerie gts) {
...
+    } else if (TYPE.BOOLEAN == gts.type) {
+      if (!gts.booleanValues.get(index1) && gts.booleanValues.get(index2)) {
+        return -1;
+      } else if (gts.booleanValues.get(index1) && !gts.booleanValues.get(index2)) {
+        return 1;
+      }
+    }
+
+    // if ticks and values are equals, test locations
+    if (null != gts.locations) {
```
hbs: Your comparison considers that no location should be compared as a valid location, it should be treated separately, i.e. an index with no location should be less than one with a location

PR #466, opened 2019-06-17
Allow RUN to use registers

ftence:


warp10/src/main/java/io/warp10/script/functions/RUN.java
```diff
@@ -17,26 +17,33 @@
...
   @Override
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
     Object symbol = stack.pop();
-    
-    if (!(symbol instanceof String)) {
-      throw new WarpScriptException(getName() + " expects macro name to be a string.");
+
+    if (symbol instanceof String) {
+      stack.run(symbol.toString());
+    } else if (symbol instanceof Long || symbol instanceof Integer) {
```
hbs: It should never be an Integer from WarpScript

PR #463, opened 2019-06-14
Shapes of nested lists [WIP -to be splitted in 3 separate PRs]

randomboolean:
- SHAPE
- HULLSHAPE
- CHECKSHAPE
- PERMUTE
- Update GET and PUT for nested lists.
- update FormattedWarpscriptFunction to check subtypes of Lists and Maps

PR #468, opened 2019-06-18
Get set nested lists

randomboolean:
Update GET and PUT for nested lists viewed as multidimensional arrays.

Part 1/3 of PR #463 that is split.
Comments from PR #463 are answered.


warp10/src/main/java/io/warp10/script/functions/GET.java
```diff
@@ -85,4 +100,37 @@ public static int computeAndCheckIndex(NamedWarpScriptFunction func, int index,
...
+   * -> recNestedGet(func, (nestedList.get(0)).get(1), [ 2 ])
+   * -> return ((nestedList.get(0)).get(1)).get(2)
+   *
+   * @param func        The function that calls this method
+   * @param nestedList  The list to which to get the elements
+   * @param indexList   The list of indices.
+   * @return
+   * @throws WarpScriptException
+   */
+  static Object recNestedGet(NamedWarpScriptFunction func, List nestedList, List<Number> indexList) throws WarpScriptException {
```
randomboolean: From PR#463, hbs:"This method should not need the func attribute, the WarpScriptException should be caught by the caller which will wrap it and add its name"

randomboolean: I think it needs the func attribute in case it is used by another function to have a better error message.

hbs: Please remove the func attribute.

hbs: Make the function non recursive, also removing the copyIndices allocation since you need only track an index to the original list.

warp10/src/main/java/io/warp10/script/functions/GET.java
```diff
@@ -47,42 +47,74 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+    } else if (coll instanceof byte[]) {
+      throw new WarpScriptException(getName() + " expects the key to be an integer when operating on a byte array.");
+
+    } else if (!(key instanceof List)) {
+      throw new WarpScriptException(getName() + " expects the key to be an integer or a list of integers when operating on a List.");
+
+    } else {
+      for (Object o: (List) key) {
+        if (!(o instanceof Number)) {
+          throw new WarpScriptException(getName() + " expects the key to be an integer or a list of integers when operating on a List.");
```
hbs: A list of Double will get through here, should we check that elements are Longs which is the only integer type generated by WarpScript?

randomboolean: I don't think so if we want to be coherent with GET e.g. "[ 0 ] 0.3 GET" doesn't raise an exception and gives 0

randomboolean: Following our discussion, should also change behavior of GET on Doubles

warp10/src/main/java/io/warp10/script/functions/PUT.java
```diff
@@ -25,30 +25,30 @@
...
-    
+    Object value = stack.pop();
+
     Object maporlist = stack.peek();
 
     if (maporlist instanceof Map) {
-      ((Map) maporlist).put(key, value);      
+      ((Map) maporlist).put(key, value);
     } else if (maporlist instanceof List) {
       if (!(key instanceof Number)) {
```
hbs: Here too the key should be checked against Long instead of Number as Doubles will pass through

PR #482, opened 2019-06-27
ToLongBytes can take a list of long

pi-r-p:
A little improvement to build byte array faster from a list of long.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
+++ b/warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
@@ -16,31 +16,32 @@
 
 package io.warp10.script.functions;
 
+import com.google.common.primitives.Longs;
 import io.warp10.script.NamedWarpScriptFunction;
 import io.warp10.script.WarpScriptException;
 import io.warp10.script.WarpScriptStack;
 import io.warp10.script.WarpScriptStackFunction;
 
+import java.nio.ByteBuffer;
 import java.util.Arrays;
-
-import com.google.common.primitives.Longs;
+import java.util.List;
 
 /**
  * Converts a LONG to a byte array
  */
 public class TOLONGBYTES extends NamedWarpScriptFunction implements WarpScriptStackFunction {
-
+  
   public TOLONGBYTES(String name) {
     super(name);
   }
-
+  
   @Override
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
-
+    
     Object o = stack.pop();
-
+    
     if (o instanceof Long) {
-      long nbBytes = ((Long) o).longValue();
+      int nbBytes = ((Long) o).intValue();
       o = stack.pop();
       if (o instanceof Long && nbBytes > 0 && nbBytes <= 8) {
         //truncate the result to nbBytes when needed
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
```diff
@@ -49,13 +50,47 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
         } else {
           stack.push(Longs.toByteArray(((Long) o).longValue()));
         }
+      } else if (o instanceof List && nbBytes > 0 && nbBytes <= 8) {
+        try {
+          ByteBuffer bytes = ByteBuffer.allocate(nbBytes * ((List) o).size());
```
hbs: I'd rather see a = new byte[...]  then ByteByffer.wrap(a)

warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
```diff
@@ -49,13 +50,47 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+          if (8 == nbBytes) {
+            for (Object el : (List) o) {
```
hbs: extra space before :
ditto later in the file

warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
```diff
@@ -49,13 +50,47 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+              bytes.putLong((Long) el);
```
hbs: You are not checking that the elements are indeed Longs

warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
hbs: You also need to set the endianness to BIG_ENDIAN

warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
```diff
@@ -49,13 +50,47 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+          } else if (1 == nbBytes) {
+            for (Object el : (List) o) {
+              bytes.put(((Long) el).byteValue());
+            }
+          } else {
+            // weird length are not optimized
+            long v;
+            for (Object el : (List) o) {
+              v = ((Long) el).longValue() << (8 - nbBytes) * 8;
+              for (int i = nbBytes - 1; i >= 0; i--) {
```
hbs: since you are not using i, why not use an increasing index? It would make things clearer

warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
```diff
@@ -49,13 +50,47 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+                bytes.put((byte) ((v & 0xff00000000000000L) >> 56));
```
hbs: Probably better to shift first then apply a mask

warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
```diff
@@ -49,13 +50,47 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+                v <<= 8;
+              }
+            }
+          }
+          stack.push(bytes.array());
+        } catch (Exception e) {
+          throw new WarpScriptException(getName() + " operates on a LONG or a list of LONG and expects a number of output bytes between 1 and 8 on top of the stack.");
```
hbs: Output bytes per LONG

warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
pi-r-p: Ok, done.

warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
pi-r-p: No, that would be a time waste if there is one million long. That's why I put it in a try/catch.

hbs: ok

warp10/src/main/java/io/warp10/script/functions/TOLONGBYTES.java
```diff
@@ -49,13 +51,49 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+            for (Object el: (List) o) {
+              bytes.put(((Long) el).byteValue());
+            }
+          } else {
+            // weird length are not optimized
+            long v;
+            for (Object el: (List) o) {
+              v = ((Long) el).longValue() << (8 - nbBytes) * 8;
+              for (int i = 0; i < nbBytes; i++) {
+                bytes.put((byte) (v >> 56));
```
hbs: Do an unsigned shift instead

pi-r-p: Why ? I don't care about Long MSB to be set to 0 or 1. unsigned shift vs standard shift is just a different assembly instruction, I doubt there is a clock cycle gain.

hbs: you're right indeed

PR #483, opened 2019-06-27
Pdecode from bytes

pi-r-p:
The pixel per pixel test was a workaround. A more efficient one is to always set to ARGB. It is consistent with Pgraphics, where we always set ARGB:
```java
pg.image = new BufferedImage(pg.pixelWidth, pg.pixelHeight, BufferedImage.TYPE_INT_ARGB);
```

### DIFF
--- a/warp10/src/main/java/io/warp10/script/processing/image/Pdecode.java
+++ b/warp10/src/main/java/io/warp10/script/processing/image/Pdecode.java
@@ -63,7 +63,7 @@ public class Pdecode extends NamedWarpScriptFunction implements WarpScriptStackF
     
     PImage image = new PImage(awtImage);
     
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/processing/image/Pdecode.java
```diff
@@ -79,21 +63,10 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
     
     PImage image = new PImage(awtImage);
     
-    //
-    // Check transparency
-    //
+    //The transparency test in Pimage(awtImage) is not working correctly: set format to ARGB is mandatory to keep transparency.
```
hbs: I am not sold on this comment, the question is whether or not the behavior is different when forcing the image format to ARGB

PR #484, opened 2019-06-28
Gen doc

randomboolean:
improvements for automatic doc generation

### DIFF
--- a/warp10/src/main/java/io/warp10/script/formatted/DocumentationGenerator.java
+++ b/warp10/src/main/java/io/warp10/script/formatted/DocumentationGenerator.java
@@ -16,6 +16,7 @@
 
 package io.warp10.script.formatted;
 
+import com.google.common.collect.Lists;
 import io.warp10.script.WarpScriptException;
 import io.warp10.script.functions.SNAPSHOT;
 
@@ -61,7 +62,7 @@ public class DocumentationGenerator {
 
     List<List<List<Object>>> sig = new ArrayList<>();
     List<Object> output = new ArrayList<>();
-    for (ArgumentSpecification arg: outputs) {
+    for (ArgumentSpecification arg: Lists.reverse(outputs)) {
       output.add(arg.getName() + ":" + arg.WarpScriptType());
     }
 
@@ -76,7 +77,7 @@ public class DocumentationGenerator {
       input1.add(new LinkedHashMap<>());
     }
 
-    for (ArgumentSpecification arg: args) {
+    for (ArgumentSpecification arg: Lists.reverse(args)) {
       input1.add(arg.getName() + ":" + arg.WarpScriptType());
     }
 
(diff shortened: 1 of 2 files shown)


warp10/src/main/java/io/warp10/script/formatted/DocumentationGenerator.java
```diff
@@ -61,7 +62,7 @@
 
     List<List<List<Object>>> sig = new ArrayList<>();
     List<Object> output = new ArrayList<>();
-    for (ArgumentSpecification arg: outputs) {
+    for (ArgumentSpecification arg: Lists.reverse(outputs)) {
```
hbs: Use Collections.reverse() instead of introducing a dependency

warp10/src/main/java/io/warp10/script/formatted/DocumentationGenerator.java
```diff
@@ -76,7 +77,7 @@
       input1.add(new LinkedHashMap<>());
     }
 
-    for (ArgumentSpecification arg: args) {
+    for (ArgumentSpecification arg: Lists.reverse(args)) {
```
hbs: ditto

warp10/src/main/java/io/warp10/script/formatted/DocumentationGenerator.java
```diff
@@ -92,11 +93,11 @@
     List<Object> input2 = new ArrayList<>();
     LinkedHashMap<String, String> optMap = new LinkedHashMap<>();
 
-    for (ArgumentSpecification arg: args) {
+    for (ArgumentSpecification arg: Lists.reverse(optArgs)) {
       optMap.put(arg.getName(), arg.getName() + ":" + arg.WarpScriptType());
     }
 
-    for (ArgumentSpecification arg: optArgs) {
+    for (ArgumentSpecification arg: Lists.reverse(args)) {
```
hbs: Does the list need to be re-reversed?

warp10/src/main/java/io/warp10/script/formatted/DocumentationGenerator.java
```diff
@@ -112,13 +113,13 @@
     //
 
     LinkedHashMap<String, String> params = new LinkedHashMap<>();
-    for (ArgumentSpecification arg: args) {
+    for (ArgumentSpecification arg: Lists.reverse(optArgs)) {
       params.put(arg.getName(), arg.getDoc());
     }
-    for (ArgumentSpecification arg: optArgs) {
+    for (ArgumentSpecification arg: Lists.reverse(args)) {
```
hbs: same question

warp10/src/main/java/io/warp10/script/formatted/DocumentationGenerator.java
randomboolean: Lists.reverse(), contrary to Collections.reverse(), returns a view, so that there is no side effect on the list when using this function

warp10/src/main/java/io/warp10/script/formatted/DocumentationGenerator.java
randomboolean: doesn't re-reversed since it return a view

warp10/src/main/java/io/warp10/script/formatted/DocumentationGenerator.java
randomboolean: ditto

PR #488, opened 2019-07-05
Refactored WRAP to remove WRAPRAW

hbs:


warp10/src/main/java/io/warp10/script/functions/WRAP.java
```diff
@@ -46,22 +49,54 @@ public WRAP(String name, boolean opt) {
     super(name);
     this.opt = opt;
     this.compress = true;
-
+    this.raw = false;
```
ftence: Call `this(name, opt, true)`. Same idea with other constructors.

hbs:
PTAL

ftence:
No more remarks.

PR #511, opened 2019-07-23
Function to convert from/to hhcodes to handle NaN and NO_LOCATION.

ftence:
Also fix condition and use Java 8 Long.parseUnsignedLong for speed-up.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/HHCODETO.java
+++ b/warp10/src/main/java/io/warp10/script/functions/HHCODETO.java
@@ -31,9 +31,16 @@ import com.google.common.primitives.Longs;
  * Convert a GeoHash to lat/lon
  */
 public class HHCODETO extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+
+  private final boolean useGtsConvention;
   
   public HHCODETO(String name) {
+    this(name, false);
+  }
+
+  public HHCODETO(String name, boolean useGtsConvention) {
     super(name);
+    this.useGtsConvention = useGtsConvention;
   }
   
   @Override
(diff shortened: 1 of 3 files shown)


warp10/src/main/java/io/warp10/script/functions/HHCODETO.java
```diff
@@ -52,17 +53,21 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-      latlon = GeoXPLib.fromGeoXPPoint(hh);
+      hh = Longs.fromByteArray((byte[]) hhcode);
     } else {
       throw new WarpScriptException(getName() + " expects a long, a string or a byte array.");
     }
-    
-    stack.push(latlon[0]);
-    stack.push(latlon[1]);
+
+    if (GeoTimeSerie.NO_LOCATION == hh) {
```
hbs: This changes the behavior, HHCodes can be used in other contexts than GTS, so the special value NO_LOCATION should not lead to NaN NaN when using HHCODETO. Maybe create a variation of this function which produces NaN NaN for NO_LOCATION, but is called differently.

warp10/src/main/java/io/warp10/script/functions/TOHHCODE.java
```diff
@@ -43,12 +44,21 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-    
-    long geoxppoint = GeoXPLib.toGeoXPPoint(((Number) lat).doubleValue(), ((Number) lon).doubleValue());
-    
+
+    double latDouble = ((Number) lat).doubleValue();
+    double lonDouble = ((Number) lon).doubleValue();
+
+    long geoxppoint = GeoTimeSerie.NO_LOCATION;
+
+    if (Double.isNaN(latDouble) != Double.isNaN(lonDouble)) {
```
hbs: Should there be an alternate version which pushed NO_LOCATION on the stack when NaN and NaN are passed for lat/lon?

ftence:
Agreed, will add GTSHHCODE-> and ->GTSHHCODE.

PR #518, opened 2019-07-29
Fix buildSelector to prefix classname with = when starting with = or ~

ftence:


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -7891,7 +7891,13 @@ public static String buildSelector(GeoTimeSerie gts) {
   
   public static String buildSelector(Metadata metadata) {
     StringBuilder sb = new StringBuilder();
+
+    char nameFirstChar = metadata.getName().charAt(0);
```
hbs: This could fail with a GTS with an empty class name

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -7891,7 +7891,13 @@ public static String buildSelector(GeoTimeSerie gts) {
...
+    if ('=' == nameFirstChar || '~' == nameFirstChar) {
+      sb.append("%3D"); // Prepend '=' for the special character not to be interpreted
```
hbs: Why add it as %3D instead of '='?

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
ftence: My bad.

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
ftence: It was to be consistent with the escaping of name and labels but I realize it is unnecessary because '=', '{' and '}' are not escaped.

PR #519, opened 2019-07-30
Integrate formatted functions into "list or not list" warpscript pattern

randomboolean:
First argument of a formatted function can now be defined as "list-expandable".

In this argument is a list, a list of results as though it was applied foreach element of this list is returned.


hbs:
This is becoming too complicated for new users to use, we might consider pulling the plug altogether on this approach in the near future.

PR #521, opened 2019-07-30
Add a PIVOT function

hbs:


warp10/src/main/java/io/warp10/script/functions/PIVOT.java
```diff
@@ -0,0 +1,302 @@
...
+            break;
+          }
+        }
+        
+        if (skip) {
+          continue;
+        }    
+        
+        // Increase the index of the first labeling GTS
+        labelingIndices[0]++;
```
ftence: Too early, this index is used line 176 & 178.
Increment should be done line 261.

warp10/src/main/java/io/warp10/script/functions/PIVOT.java
```diff
@@ -0,0 +1,302 @@
...
+import io.warp10.continuum.gts.GTSHelper;
+import io.warp10.continuum.gts.GeoTimeSerie;
+import io.warp10.script.NamedWarpScriptFunction;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.script.WarpScriptStackFunction;
+
+public class PIVOT extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+  
+  private final boolean synchronous;
```
ftence: There is little code in common between synchronous and non-synchronous version, maybe splitting the two versions and sharing the common code could make the class clearer?

warp10/src/main/java/io/warp10/script/functions/PIVOT.java
```diff
@@ -0,0 +1,302 @@
...
+          throw new WarpScriptException(getName() + " labeling class '" + entry.getKey() + "' is already a label of a Geo Time Series™ to label.");
+        }
+      }
+    }
+        
+    //
+    // Now sort all GTS
+    //
+    
+    for (Object g: gts) {
```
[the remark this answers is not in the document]
hbs: The reason why it is done here is that we only do it if no exception was thrown in the previous loop.

warp10/src/main/java/io/warp10/script/functions/PIVOT.java
```diff
@@ -0,0 +1,302 @@
...
+      GTSHelper.sort((GeoTimeSerie) g);
+    }
+    
+    for (Object g: labeling) {
```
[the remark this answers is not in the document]
hbs: Ditto

warp10/src/main/java/io/warp10/script/functions/PIVOT.java
hbs: Leaving that for a future PR of yours ;-)

hbs:
PTAL

ftence:
Looks good.

PR #522, opened 2019-07-31
Initial commit of GEO.COVER, GEO.COVER.RL

hbs:


warp10/src/main/java/io/warp10/script/functions/GEOCOVER.java
```diff
@@ -0,0 +1,120 @@
...
+            while (decoder.next()) {
+              long location = decoder.getLocation();
+              if (GeoTimeSerie.NO_LOCATION != location) {
+                long cell = ((location & mask) >>> 4) | prefix;
+                cells.add(cell);
+              }
+            }            
+          }
+        }
+
```
ftence: ElementOrListStackFunction can work on any type so you still need to throw a WarpScriptException if the type of `element` is not what you expect.
So add a `else { throw ... }`.

warp10/src/main/java/io/warp10/script/functions/GEOCOVER.java
```diff
@@ -0,0 +1,120 @@
...
+              long location = GTSHelper.locationAtIndex(gts, i);
+              if (GeoTimeSerie.NO_LOCATION != location) {
+                long cell = ((location & mask) >>> 4) | prefix;
+                cells.add(cell);
+              }
+            }            
+          }
+        } else if (element instanceof GTSEncoder) {          
+          if (rhumblines) {
+            throw new WarpScriptException(getName() + " hmm weird, we should never end here!");
```
ftence: So weird it can never happen, maybe change that to a comment?

hbs:
PTAL

ftence:
Looks good.

PR #523, opened 2019-07-31
GTS to Encoder conversions

ftence:


warp10/src/main/java/io/warp10/script/functions/TOENCODER.java
```diff
@@ -50,6 +57,40 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      // If raw wrap, convert to GTS
+      if (element instanceof byte[]) {
+        TDeserializer deser = new TDeserializer(new TCompactProtocol.Factory());
+
+        try {
+          GTSWrapper wrapper = new GTSWrapper();
+
+          deser.deserialize(wrapper, (byte[]) element);
+
+          element = GTSWrapperHelper.fromGTSWrapperToGTS(wrapper);
```
hbs: It would be better to get a GTSDecoder rather than a GeoTimeSeries, this would prove faster and less hungry in memory

warp10/src/main/java/io/warp10/script/functions/GTSTO.java
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Create a GTSEncoder for each instance of GTS, wrapped GTS or raw wrapped GTS.
+ */
+public class GTSTO extends ElementOrListStackFunction {
+
+  private final ElementStackFunction gtstoFunction = new ElementStackFunction() {
+    @Override
+    public Object applyOnElement(Object element) throws WarpScriptException {
+      // If wrap, convert to raw wrap
+      if (element instanceof String) {
```
hbs: It bothers me that a function called GTS-> takes as input wrappers which are not GTS per say. A better name should be picked.

warp10/src/main/java/io/warp10/script/functions/ASENCODER.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.thrift.TDeserializer;
+import org.apache.thrift.TException;
+import org.apache.thrift.protocol.TCompactProtocol;
+
+import java.io.IOException;
+
+/**
+ * Create a GTSEncoder for each instance of GTS, wrapped GTS or raw wrapped GTS.
+ */
+public class ASENCODER extends ElementOrListStackFunction {
```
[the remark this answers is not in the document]
ftence: It can also operate on a single GTS which could make `$gts ASENCODERS` a bit misleading... but less misleading than `$lgts ASENCODER` as it is logical that a single GTS produces a single encoder but a list of GTS could produce a single encoder (as `->ENCODER`) or a list of encoders (as this function).

Agreed, I will make it plural.

warp10/src/main/java/io/warp10/script/functions/ASENCODER.java
```diff
@@ -0,0 +1,89 @@
...
+      // If raw wrap, convert to GTS
+      if (element instanceof byte[]) {
+        TDeserializer deser = new TDeserializer(new TCompactProtocol.Factory());
+
+        try {
+          GTSWrapper wrapper = new GTSWrapper();
+
+          deser.deserialize(wrapper, (byte[]) element);
+
+          element = GTSWrapperHelper.fromGTSWrapperToGTS(wrapper);
```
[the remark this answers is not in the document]
ftence: It will also allow a wrapped encoder to be given to `ASENCODERS` without error if it contains multiple types. Not sure if it is useful but it is logical.

warp10/src/main/java/io/warp10/script/functions/TOENCODER.java
```diff
@@ -50,6 +58,49 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+        } catch (IOException ioe) {
+          throw new WarpScriptException(getName() + " cannot get encoder from wrapper.", ioe);
+        }
+
+        // Wrap all added to encoder, next.
+        continue;
+      }
+
+      // If GTS, add it all to the encoder
+      if(element instanceof GeoTimeSerie){
```
hbs: Missing space.

warp10/src/main/java/io/warp10/script/functions/ASENCODERS.java
```diff
@@ -0,0 +1,91 @@
...
+          return GTSWrapperHelper.fromGTSWrapperToGTSEncoder(wrapper);
+        } catch (TException te) {
+          throw new WarpScriptException(getName() + " failed to unwrap encoder.", te);
+        } catch (IOException ioe) {
+          throw new WarpScriptException(getName() + " failed to convert wrapper to encoder.", ioe);
+        }
+      }
+
+      // If GTS, convert to an encoder
+      if (element instanceof GeoTimeSerie) {
```
hbs: Maybe encoders should be returned verbatim rather than throwing an exception.

randomboolean:
Actual documentation on [->ENCODER](https://warp10.io/doc/AIt4IZCEG3KH) seems logical but is wrong. Maybe this PR can also make it match the documentation? (and calling UNWRAPENCODER if encountering String or byte array ).

If so then I think ASENCODERS would not be needed.

ftence:
The documentation of ->ENCODER was already modified for this PR but was indeed wrong as it takes only lists and not single elements. This is now fixed and will be updated soon.

The difference between ->ENCODER and ASENCODERS is that the former encodes all GTSs into a single encoder while the latter encodes each GTS separately.

PR #524, opened 2019-07-31
Pickle

randomboolean:
->PICKLE converts GeoTimeSerie instances to a pickled python dict containing classname, labels, attributes, timestamp, value, and optionally latitude, longitude and elevation (data is copied into lists).

It also pickle GTSEncoder instances by applying ->GTS beforehand


warp10/src/main/java/io/warp10/continuum/gts/GTSPickler.java
```diff
@@ -0,0 +1,102 @@
...
+  private static String TIMESTAMPS_KEY = "timestamp";
+  private static String VALUES_KEY = "value";
+  private static String LATITUDE_KEY = "latitude";
+  private static String LONGITUDE_KEY = "longitude";
+  private static String ELEVATION_KEY = "elevation";
+
+  public void pickle(Object o, OutputStream out, Pickler currentPickler) throws PickleException, IOException {
+    GeoTimeSerie gts = (GeoTimeSerie) o;
+
+    Map<String, Object> gtsAsMap = new HashMap<>();
```
hbs: Do not use the diamond notation, put explicit types for clarity.

warp10/src/main/java/io/warp10/continuum/gts/GTSPickler.java
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Pickler for Geo Time Series
+ */
+public class GTSPickler implements IObjectPickler {
+
+  private static String CLASSNAME_KEY = "classname";
+  private static String LABELS_KEY = "labels";
+  private static String ATTRIBUTES_KEY = "attributes";
+  private static String TIMESTAMPS_KEY = "timestamp";
```
hbs: timestamp / value / latitude / longitude / elevation should be plurals as they contain arrays of multiple values

warp10/src/main/java/io/warp10/continuum/gts/GTSPickler.java
```diff
@@ -0,0 +1,102 @@
...
+  private static String VALUES_KEY = "value";
+  private static String LATITUDE_KEY = "latitude";
+  private static String LONGITUDE_KEY = "longitude";
+  private static String ELEVATION_KEY = "elevation";
+
+  public void pickle(Object o, OutputStream out, Pickler currentPickler) throws PickleException, IOException {
+    GeoTimeSerie gts = (GeoTimeSerie) o;
```
hbs: This will throw a ClassCastException if o is not a GeoTimeSerie, either change the pickle signature so o is a GTS or throw a clean exception.

warp10/src/main/java/io/warp10/continuum/gts/GTSPickler.java
```diff
@@ -0,0 +1,102 @@
...
+
+      for (int i = 0; i < gts.values; i++) {
+        values.add(GTSHelper.valueAtIndex(gts, i));
+      }
+      gtsAsMap.put(VALUES_KEY, values);
+    }
+
+    if (gts.hasLocations()) {
+      long[] locations = gts.locations;
+      List<Double> lats = new ArrayList<Double>();
```
hbs: If pickle does not care if those are floats or doubles, make that floats as it will be less resource hungry.

warp10/src/main/java/io/warp10/script/functions/TOPICKLE.java
```diff
@@ -16,29 +16,45 @@
...
 
+import net.razorvine.pickle.Pickler;
 import net.razorvine.pyro.serializer.PickleSerializer;
 
 /**
  * Encode an object as Pickle
  */
 public class TOPICKLE extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+
+  private static final TOGTS TOGTS = new TOGTS("TOGTS");
```
hbs: TOGTS is not the name under which this function is usually used, please use the name ->GTS by making it a constant in WarpScriptLib.

warp10/src/main/java/io/warp10/script/functions/TOPICKLE.java
```diff
@@ -16,29 +16,45 @@
...
     super(name);
   }
   
   @Override
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
-    Object o = stack.pop();
+    Object o = stack.peek();
+
+    if (o instanceof GTSEncoder) {
+      stack = (WarpScriptStack) TOGTS.apply(stack);
```
hbs: return value of call to apply can be safely discarded, don't assign it to stack

warp10/src/main/java/io/warp10/script/functions/TOPICKLE.java
```diff
@@ -16,29 +16,45 @@
...
+    }
+
+    o = stack.pop();
     
     try {
+      // register pickler for custom classes
+      Pickler.registerCustomPickler(GeoTimeSerie.class, new GTSPickler());
```
hbs: Can this be factored out? Is GTSPickler MT safe?

warp10/src/test/java/io/warp10/continuum/gts/PicklingTest.java
```diff
@@ -0,0 +1,116 @@
...
+    stack.exec("[ 1 10 <% s %> FOR ] [ 1 10 <% us %> FOR ] [ 1 10 <% us %> FOR ] [] [ 1 10 <% DROP RAND %> FOR ] MAKEGTS" + System.lineSeparator());
+    stack.exec("'gts' STORE" + System.lineSeparator());
+    stack.exec("$gts 'gts' RENAME { 'k1' 'v1' 'k2' 'v2' } RELABEL { 'k3' 'v3' 'k4' 'v4' } SETATTRIBUTES" + System.lineSeparator());
+    stack.exec("->PICKLE PICKLE-> 'map' STORE" + System.lineSeparator());
+
+    //
+    // Asserts
+    //
+
+    stack.exec("$map SIZE 7 == ASSERT" + System.lineSeparator());
```
hbs: The line separator is useless as the whole line is parsed regardless of the fact that it ends with a LF or CRLF

warp10/src/test/java/io/warp10/continuum/gts/PicklingTest.java
```diff
@@ -0,0 +1,116 @@
...
+    stack.exec("[ 1 10 <% s %> FOR ] [ 1 10 <% us %> FOR ] [ 1 10 <% us %> FOR ] [ 1 10 <% ms %> FOR ] [ 1 10 <% DROP RAND %> FOR ] MAKEGTS" + System.lineSeparator());
+    stack.exec("'gts' STORE" + System.lineSeparator());
+    stack.exec("$gts 'gts' RENAME { 'k1' 'v1' 'k2' 'v2' } RELABEL { 'k3' 'v3' 'k4' 'v4' } SETATTRIBUTES" + System.lineSeparator());
+    stack.exec("->PICKLE PICKLE-> 'map' STORE" + System.lineSeparator());
+
+    //
+    // Asserts
+    //
+
+    stack.exec("$map SIZE 8 == ASSERT" + System.lineSeparator());
```
hbs: ditto

warp10/src/main/java/io/warp10/continuum/gts/GTSPickler.java
randomboolean: Python does care about double precision (in fact what it calls float have double precision and it doesn't call anything double)

warp10/src/main/java/io/warp10/script/functions/TOPICKLE.java
randomboolean: I don't think it can be factored out if you are using the same serializer for other objects that may contain geo time series or be contained by geo time series.

warp10/src/main/java/io/warp10/continuum/gts/GTSPickler.java
hbs: The question is, does pickle actually convert Float to doubles? If so you can allocate Floats instead of Doubles since lat/lon are really stored on 32bits each

warp10/src/main/java/io/warp10/script/functions/TOPICKLE.java
hbs: The Serializer is thread safe IIRC

warp10/src/main/java/io/warp10/continuum/gts/GTSPickler.java
randomboolean: yes that's the case, it converts java floats to python floats (which are the same precision than java doubles)

warp10/src/main/java/io/warp10/script/functions/TOPICKLE.java
randomboolean: No it doesn't seem thread safe: [Pickler](https://github.com/irmen/Pyrolite/blob/master/java/src/main/java/net/razorvine/pickle/Pickler.java) used by [PickleSerializer](https://github.com/irmen/Pyrolite/blob/master/java/src/main/java/net/razorvine/pyro/serializer/PickleSerializer.java) is not

hbs: The call ro registerCustomPickler at each call to apply does not seem right, the registration is probably to be done once and for all since your GTSPickler is MT safe

randomboolean: Sure. I didn't get that was what you meant by "factored out".

warp10/src/main/java/io/warp10/continuum/gts/GTSPickler.java
```diff
@@ -0,0 +1,106 @@
...
+  private static String ATTRIBUTES_KEY = "attributes";
+  private static String TIMESTAMPS_KEY = "timestamps";
+  private static String VALUES_KEY = "values";
+  private static String LATITUDE_KEY = "latitudes";
+  private static String LONGITUDE_KEY = "longitudes";
+  private static String ELEVATION_KEY = "elevations";
+
+  public void pickle(Object o, OutputStream out, Pickler currentPickler) throws PickleException, IOException {
+    if (!(o instanceof GeoTimeSerie)) {
+      throw new IllegalArgumentException("Invalid type for first argument. It must be a GeoTimeSerie.");
```
hbs: Geo Time Series™

warp10/src/main/java/io/warp10/script/functions/TOPICKLE.java
```diff
@@ -16,32 +16,54 @@
...
     super(name);
   }
   
   @Override
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
-    Object o = stack.pop();
+    Object o = stack.peek();
+
+    if (o instanceof GTSEncoder) {
+      ((TOGTS) WarpScriptLib.getFunction(WarpScriptLib.TOGTS)).apply(stack);
```
hbs: You need to allocate your own instance in the static block, because this function may have been disabled or redefined by an extension, so retrieving it in WarpScriptLib is not safe

PR #525, opened 2019-08-01
Multi-threading for FormattedWarpscriptFunction

randomboolean:
add the option for formatted functions to use multi-threading on their first argument in case it is list-expandable


hbs:
MT should never be allowed by default, this causes too much risk of resource exhaustion.

MT execution should be handled by a dedicated function inside of an optional extension with configuration keys to limit the resources consumed globally and per call.

Closing this PR as this is out of those guidelines.

PR #526, opened 2019-08-02
Fixed commonticks so it correctly copies duplicate ticks

hbs:
This fix allows the following to correctly preserve datapoints at tick 0

NEWGTS 'a' RENAME
0 NaN NaN NaN 0 ADDVALUE
0 NaN NaN NaN 10 ADDVALUE
1 NaN NaN NaN 1 ADDVALUE
2 NaN NaN NaN 2 ADDVALUE

NEWGTS 'b' RENAME
0 NaN NaN NaN 2 ADDVALUE
0 NaN NaN NaN 3 ADDVALUE

2 ->LIST
COMMONTICKS


ftence:
Looks good.

PR #529, opened 2019-08-09
Binary

randomboolean:
Apply a macro as a binary operator on two input GTS.

For example:
`$gtsa $gtsb <% == ASSERT NULL %> MACROBINARYOP DROP`

### DIFF
--- a/warp10/src/main/java/io/warp10/script/WarpScriptLib.java
+++ b/warp10/src/main/java/io/warp10/script/WarpScriptLib.java
@@ -89,6 +89,7 @@ import io.warp10.script.binary.GT;
 import io.warp10.script.binary.INPLACEADD;
 import io.warp10.script.binary.LE;
 import io.warp10.script.binary.LT;
+import io.warp10.script.binary.MACROBINARYOP;
 import io.warp10.script.binary.MOD;
 import io.warp10.script.binary.MUL;
 import io.warp10.script.binary.NE;
@@ -690,6 +691,7 @@ public class WarpScriptLib {
     addNamedWarpScriptFunction(new BitwiseOR("|"));
     addNamedWarpScriptFunction(new BitwiseXOR("^"));
     addNamedWarpScriptFunction(new ALMOSTEQ("~="));
+    addNamedWarpScriptFunction(new MACROBINARYOP("MACROBINARYOP"));
 
     // Bitset ops
     addNamedWarpScriptFunction(new BITGET("BITGET"));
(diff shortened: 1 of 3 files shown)


warp10/src/main/java/io/warp10/script/binary/MACROBINARYOP.java
```diff
@@ -0,0 +1,125 @@
...
+    //
+
+    final WarpScriptStack stack_ = stack;
+
+    GTSOpsHelper.GTSBinaryOp op = new GTSOpsHelper.GTSBinaryOp() {
+
+      @Override
+      public Object op(GeoTimeSerie gtsa, GeoTimeSerie gtsb, int idxa, int idxb) throws WarpScriptException {
+
+        stack_.push(GTSHelper.valueAtIndex(gtsa, idxa));
```
hbs: A function as generic as this one aims to be should take into account ticks,location,elevation and value, not only value.

warp10/src/main/java/io/warp10/script/binary/MACROBINARYOP.java
```diff
@@ -0,0 +1,125 @@
...
+import io.warp10.continuum.gts.GTSHelper;
+import io.warp10.continuum.gts.GTSOpsHelper;
+import io.warp10.continuum.gts.GeoTimeSerie;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.script.formatted.FormattedWarpScriptFunction;
+
+import java.util.Map;
+
+public class MACROBINARYOP extends FormattedWarpScriptFunction {
```
hbs: Given this function applies the macro instead of producing a wrapper around it (in the way MACROREDUCER, MACROMAPPER do), the name should be changed to something else.

warp10/src/main/java/io/warp10/script/binary/MACROBINARYOP.java
```diff
@@ -0,0 +1,125 @@
+package io.warp10.script.binary;
```
hbs: Missing copyright notice

warp10/src/main/java/io/warp10/script/WarpScriptLib.java
```diff
@@ -89,6 +89,7 @@
 import io.warp10.script.binary.INPLACEADD;
 import io.warp10.script.binary.LE;
 import io.warp10.script.binary.LT;
+import io.warp10.script.binary.MACROBINARYOP;
```
hbs: What purpose does this function serve that neither REDUCE nor APPLY provide?

This is the first point to address before we decide we want to add this to the core of WarpScript.

hbs:
Closing this for now.

PR #533, opened 2019-08-19
Systemd unit fix

pi-r-p:
It is a forking unit, not a oneshot with remainAfterExit.
Tested with several level of kill, systemd relaunch warp 10 correctly.
journalctl is now cleaner.

### DIFF
--- a/warp10/src/main/sh/warp10.service
+++ b/warp10/src/main/sh/warp10.service
@@ -6,6 +6,7 @@ Type=forking
 User=warp10
 ExecStart=/opt/warp10-@VERSION@/bin/warp10-standalone.sh start
 ExecStop=/opt/warp10-@VERSION@/bin/warp10-standalone.sh stop
+SuccessExitStatus=143
 RestartSec=10s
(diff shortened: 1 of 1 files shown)


warp10/src/main/sh/warp10.service
```diff
@@ -2,11 +2,21 @@
...
-Type=oneshot
+Type=forking
 User=warp10
 ExecStart=/opt/warp10-@VERSION@/bin/warp10-standalone.sh start
 ExecStop=/opt/warp10-@VERSION@/bin/warp10-standalone.sh stop
-RemainAfterExit=yes
+SuccessExitStatus=143
+RestartSec=10s
+TimeoutStartSec=60s
+#If you do not want systemd to monitor and restart Warp&nbsp;10™ automatically, comment out this line:
```
hbs: Could you use non HTML please

pi-r-p: sorry. copy paste of the documentation...

PR #535, opened 2019-08-27
Add an aggregator and mapper to extend comparison mappers to tick, lat, lon, etc.

ftence:
This implementation is around 1.4% slower than the original one.


warp10/src/main/java/io/warp10/script/WarpScriptLib.java
```diff
@@ -1059,12 +1060,42 @@
     addNamedWarpScriptFunction(new NULL(NULL));
     addNamedWarpScriptFunction(new ISNULL("ISNULL"));
     addNamedWarpScriptFunction(new MapperReplace.Builder("mapper.replace"));
-    addNamedWarpScriptFunction(new MAPPERGT("mapper.gt"));
-    addNamedWarpScriptFunction(new MAPPERGE("mapper.ge"));
-    addNamedWarpScriptFunction(new MAPPEREQ("mapper.eq"));
-    addNamedWarpScriptFunction(new MAPPERNE("mapper.ne"));
-    addNamedWarpScriptFunction(new MAPPERLE("mapper.le"));
-    addNamedWarpScriptFunction(new MAPPERLT("mapper.lt"));
+    addNamedWarpScriptFunction(new MapperCompareTo("mapper.gt", CompareTo.Compared.VALUE, CompareTo.Comparison.GT));
```
hbs: Could you use constants as per PR #538

hbs:
Did you identify the cause of the lesser performance?

warp10/src/main/java/io/warp10/script/WarpScriptLib.java
ftence: Done.

ftence:
No, i did not used a profiler to check that but I expected such decrease in performance because of the use of `signum`, `compare`, `==` and `^` instead of a single comparison operator.

I think that the flexibility this PR adds outweighs the small performance loss. If performance is seen as critical, the current code can be optimized but I fear it would be far less readable.

PR #537, opened 2019-08-30
add error location tip into error message

pi-r-p:
Idea: to know exactly what instruction in the line causes the failure.
- Independant of LINEON, there is no cost to do it.
- comments are preserved, long statements are cut to 30 characters (also helps to preserve secrets).

Example:
```java
//will fail
[ 1 ] 0    GET 1 GET 1 GET STOP    'very very very very very very very very long string'  //very very very very very very very very long comment

```

Error message:
`2 in section '[TOP]':  [ 1 ] 0    GET 1 =>GET<= 1 GET STOP    'very%20very%20very%20very%...  //very very very very very very very very long comment  : GET operates on a map, list or byte array.`

Possible enhancement : copy `StringUtils.abbreviate` function somewhere to avoid another import.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
+++ b/warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
@@ -47,6 +47,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.util.Progressable;
 
 public class MemoryWarpScriptStack implements WarpScriptStack, Progressable {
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -594,308 +595,321 @@ public void exec(String line) throws WarpScriptException {
...
-            }          
-          } finally {
-            Sensision.update(SensisionConstants.SENSISION_CLASS_WARPSCRIPT_FUNCTION_COUNT, labels, 1);
-            Sensision.update(SensisionConstants.SENSISION_CLASS_WARPSCRIPT_FUNCTION_TIME_US, labels, (System.nanoTime() - nano) / 1000L);
+            } finally {
+              Sensision.update(SensisionConstants.SENSISION_CLASS_WARPSCRIPT_FUNCTION_COUNT, labels, 1);
+              Sensision.update(SensisionConstants.SENSISION_CLASS_WARPSCRIPT_FUNCTION_TIME_US, labels, (System.nanoTime() - nano) / 1000L);
+            }
+          }
+        } catch (WarpScriptException e) {
```
hbs: You cannot catch WarpScriptException without previously catching WarpScriptATCException, otherwise your throw clause will break any ongoing asynchronous transfer of control, you need to add a catch clause for at least WarpScriptATCException with a body that rethrows it as is.

PR #538, opened 2019-08-30
Static names

randomboolean:
Extensions that use a function cannot use a static variable of a name since they cannot modify WarpScriptLib.

This PR refactors WarpScriptLib with static variables for (almost all) names.


warp10/src/main/java/io/warp10/script/WarpScriptLib.java
```diff
@@ -447,192 +446,835 @@
...
     addNamedWarpScriptFunction(new CHRONOEND(CHRONOEND));
-    addNamedWarpScriptFunction(new CHRONOSTATS("CHRONOSTATS"));
-    addNamedWarpScriptFunction(new TOLIST("->LIST"));
-    addNamedWarpScriptFunction(new LISTTO("LIST->"));
-    addNamedWarpScriptFunction(new UNLIST("UNLIST"));
+    addNamedWarpScriptFunction(new CHRONOSTATS(CHRONOSTATS));
+    addNamedWarpScriptFunction(new TOLIST(TOLIST));
+    addNamedWarpScriptFunction(new LISTTO(LISTTO));
+    addNamedWarpScriptFunction(new UNLIST(UNLIST));
     addNamedWarpScriptFunction(new TOSET(TO_SET));
```
hbs: Why use TO_SET instead of TOSET?

warp10/src/main/java/io/warp10/script/WarpScriptLib.java
```diff
@@ -447,192 +446,835 @@
...
-    addNamedWarpScriptFunction(new SETTO("SET->"));
+    addNamedWarpScriptFunction(new SETTO(SETTO));
     addNamedWarpScriptFunction(new TOVECTOR(TO_VECTOR));
```
hbs: Why use TO_VECTOR instead of TOVECTOR?

warp10/src/test/java/io/warp10/script/WarpScriptLibHasSameNamesTest.java
```diff
@@ -0,0 +1,1644 @@
+//
```
hbs: What is the use of this test class beyond checking once that your PR did not introduce naming errors?

warp10/src/main/java/io/warp10/script/WarpScriptLib.java
```diff
@@ -447,192 +446,835 @@
...
+  public static final String INCREMENTEXACT = "INCREMENTEXACT";
+  public static final String DECREMENTEXACT = "DECREMENTEXACT";
+  public static final String NEGATEEXACT = "NEGATEEXACT";
+  public static final String TOINTEXACT = "TOINTEXACT";
+  public static final String SCALB = "SCALB";
+  public static final String RANDOM = "RANDOM";
+  public static final String NEXTDOWN = "NEXTDOWN";
+  public static final String GETEXPONENT = "GETEXPONENT";
+  public static final String IDENT = "IDENT";
+  public static final String Pencode = "Pencode";
```
hbs: Constants should be all uppercase

warp10/src/test/java/io/warp10/script/WarpScriptLibHasSameNamesTest.java
randomboolean: None, I will remove it when PR is ready to be accepted

warp10/src/main/java/io/warp10/script/WarpScriptLib.java
randomboolean: That was already set as this. Should this PR also fix that ?

warp10/src/main/java/io/warp10/script/WarpScriptLib.java
randomboolean: ditto

PR #542, opened 2019-09-09
Remove sorted flag when sorting by values or locations

ftence:


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -341,6 +341,9 @@ public static final int binarySearchTick(GeoTimeSerie gts, int fromIndex, int to
   
   public static final GeoTimeSerie valueSort(GeoTimeSerie gts, boolean reversed) {
     quicksortByValue(gts, 0, gts.values - 1, reversed);
+
```
hbs: Shouldn't the flags be set before sorting?

ftence: It does not matter, `quicksort`, `quicksortByValue`, `quicksortByLocation` and `fullquicksort` do not check for this flag.

PR #543, opened 2019-09-09
Cleaner and more understandable error messages from Throwables

ftence:


warp10/src/main/java/io/warp10/ThrowableUtils.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public static String getErrorMessage(Throwable t) {
+    // First, simply get the message of the root Throwable
+    String message = t.getMessage();
+
+    // If there is a cause to this Throwable, find if adding it is a good option.
+    while (null != t.getCause()) {
+      // Test if the message of the root Throwable is simply the toString of the cause or its message.
+      // The former happens when the cause Throwable has been passed as a parameter to the root Throwable.
+      if (message.equals(t.getCause().toString()) || message.equals(t.getCause().getMessage())) {
```
hbs: Can message be null?

ftence: Could be, fixed.

PR #548, opened 2019-09-11
Doc gen

randomboolean:
Move code that generate unit tests and docs to main so that extensions can pull it from warpscript dependency


warp10/src/test/java/io/warp10/script/formatted/UnitTestAndDocGenerationExample.java
```diff
@@ -0,0 +1,71 @@
...
+  //
+  // Overridden test run parameters
+  //
+
+  protected boolean WRITE() {
+    return false;
+  }
+
+  protected String OUTPUT_FOLDER() {
+    return "/my/output/folder/";
```
hbs: This should be settable in a way, via properties?

PR #549, opened 2019-09-11
Add SORTWITH to sort a list using a comparator macro.

ftence:
Addresses https://groups.google.com/forum/#!topic/warp10-users/7dGQq5LXeD8

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/SORTWITH.java
+++ b/warp10/src/main/java/io/warp10/script/functions/SORTWITH.java
@@ -1,0 +1,89 @@
+//
+//   Copyright 2019  SenX S.A.S.
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


warp10/src/main/java/io/warp10/script/functions/SORTWITH.java
```diff
@@ -0,0 +1,89 @@
...
+        public int compare(Object o1, Object o2) {
+          try {
+            stack.push(o1);
+            stack.push(o2);
+            stack.exec(macro);
+
+            Object topComp = stack.pop();
+
+            if (!(topComp instanceof Long)) {
+              throw new WarpScriptException(getName() + " was given a macro which did not returned an Integer.");
```
hbs: a LONG

ftence: True it's a LONG but it should be within the bounds of a INT. I thought it was simpler to just say INT, do I make that more precise?

hbs: Type is technically a LONG so LONG should appear in the message, you could complete the message saying it should be within the bounds of a 32 bit integer

PR #559, opened 2019-10-04
seeded version of RANDPDF

randomboolean:


warp10/src/main/java/io/warp10/script/functions/RANDPDF.java
```diff
@@ -40,11 +41,15 @@
...
   
   public static class Builder extends NamedWarpScriptFunction implements WarpScriptStackFunction {
-    
-    public Builder(String name) {
+
+    private final boolean seeded;
+
+    public Builder(String name, boolean seeded) {
       super(name);
+      this.seeded =seeded;
```
hbs: missing whitespace after '='

warp10/src/test/java/io/warp10/script/functions/SRANDPDFtest.java
```diff
@@ -0,0 +1,96 @@
+//
```
hbs: The unit tests should be in the documentation .mc2 file, written in WarpScript

PR #564, opened 2019-10-14
Truncate error message for HTTP reasons to properly display cause

ftence:


warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
```diff
@@ -1127,7 +1127,9 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
       }      
     } catch (Throwable t) { // Catch everything else this handler could return 200 on a OOM exception
       if (!response.isCommitted()) {
-        String msg = "Error when updating data: " + ThrowableUtils.getErrorMessage(t);
+        String prefix = "Error when updating data: ";
+        // HTTP reason in Jetty is 1024 chars max
+        String msg = prefix + ThrowableUtils.getErrorMessage(t, 1024 - prefix.length());
```
hbs: Could this 1024 be defined as a constant somewhere, with an accompanying comment so its meaning is clear when reading the code.

hbs: Or better yet, use Jetty's own constant as the default value of that new constant.

warp10/src/main/java/io/warp10/ThrowableUtils.java
```diff
@@ -80,7 +80,7 @@ public static String getErrorMessage(Throwable t, int maxSize) {
       message = simpleClassName;
     }
 
-    return message.substring(0, Math.min(message.length(), maxSize));
+    return message.substring(0, Math.min(message.length(), maxSize - 3)) + "...";
```
hbs: Doesn't this mean the "..." is always added?

PR #571, opened 2019-10-15
Motionsplit

hbs:


warp10/src/main/java/io/warp10/script/functions/MOTIONSPLIT.java
```diff
@@ -0,0 +1,222 @@
...
+      if (GeoTimeSerie.NO_LOCATION != refLocation && GeoTimeSerie.NO_LOCATION != location && GeoXPLib.orthodromicDistance(refLocation, location) > proximityZoneRadius) {
+        if (timestamp - refTimestamp > proximityZoneTime && proximityZoneTraveledDistance / ((timestamp - refTimestamp) / Constants.TIME_UNITS_PER_S) < proximityZoneSpeed) {
+          mustSplit = true;
+        }
+        refLocation = location;
+        refTimestamp = timestamp;
+        proximityZoneTraveledDistance = 0;
+      }
+      
+      if (GeoTimeSerie.NO_LOCATION != refLocation && GeoTimeSerie.NO_LOCATION != location && GeoXPLib.orthodromicDistance(refLocation, location) <= proximityZoneRadius) {
```
ftence: Should be a `else` because when `orthodromicDistance(refLocation, location) > proximityZoneRadius` is `true` then `refLocation = location` and then `orthodromicDistance(refLocation, location) <= proximityZoneRadius` (well, highly likely, because it is 0) so `orthodromicDistance(previousLocation, location)` is added to `proximityZoneTraveledDistance`.

This means, the next iteration, in the case `orthodromicDistance(refLocation, location) <= proximityZoneRadius`, `proximityZoneTraveledDistance` is also updated but it already contains the distance from the previous iteration although this distance was not traveled inside the proximity zone.

pi-r-p:
It takes 6 parameters on the stack. too much, easier for readability to do a map.
where default value for split could be 'split' string, where proximity parameters could be -1 for people who did not want to use this.

The timeThreshold is functionnaly already present in TIMESPLIT. Why is it here (without minimum points notion) ?

More than a MOTIONSPLIT function with 6 inputs, why not having a SPLIT framework, taking a map of parameters as input. In the map, we could do TIMESPLIT and/or distanceTheshold split and/or proximityZone detection, and maybe more in the future (split by elevation, by heading change, a MACROSPLIT that may be less powerfull than a multi gts output macromapper, but simpler to implement).

hbs:
PTAL

warp10/src/main/java/io/warp10/script/functions/MOTIONSPLIT.java
```diff
@@ -0,0 +1,205 @@
...
+      }
+      
+      //
+      // If the current point is farther away from the reference point than 'proximityZoneRadius'
+      // change the reference point and reset the traveled distance.
+      // If the distance traveled in the proximity zone was traveled at less than proximityZoneSpeed and the time spent in the zone is above proximityZoneTime, force a split
+      //
+      
+      if (GeoTimeSerie.NO_LOCATION != refLocation && GeoTimeSerie.NO_LOCATION != location && GeoXPLib.orthodromicDistance(refLocation, location) > proximityZoneRadius) {
+        if (previousTimestamp - refTimestamp > proximityZoneTime && proximityZoneTraveledDistance / ((previousTimestamp - refTimestamp) / Constants.TIME_UNITS_PER_S) < proximityZoneSpeed) {
```
ftence: `(previousTimestamp - refTimestamp) / Constants.TIME_UNITS_PER_S` is still an integer division which could, if timestamps are very close, lead to a division by 0.

hbs:
PT(yet)AL

hbs:
@pi-r-p feel free to contribute your suggested changes to the parameter passing

PR #572, opened 2019-10-15
feat(directory): add a gts per app metrics

aurrelhebert:
In a context of user in Warp10, it can be great to keep trace of the series account per Warp application. 

In this PR, we simply add a Metrics `warp.directory.gts.perapp` directly in the directory.

Written by @StevenLeRoux and me.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
+++ b/warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
@@ -119,6 +119,11 @@ public class SensisionConstants {
    * Number of distinct owners known by a Directory
    */
   public static final String SENSISION_CLASS_CONTINUUM_DIRECTORY_OWNERS = "warp.directory.owners";
+  
+  /**
+   * Number of distinct classes per owner known by a Directory
+   */
+  public static final String SENSISION_CLASS_CONTINUUM_DIRECTORY_GTS_PERAPP = "warp.directory.gts.perapp";
 
   /**
    * Number of times the Thrift directory client cache was changed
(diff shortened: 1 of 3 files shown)


warp10/src/main/java/io/warp10/continuum/store/Directory.java
```diff
@@ -1851,6 +1867,9 @@ public void run() {
               // 128bits
               if (null == directory.metadatas.get(metadata.getName()).put(labelsId, metadata)) {
                 Sensision.update(SensisionConstants.SENSISION_CLASS_CONTINUUM_DIRECTORY_GTS, Sensision.EMPTY_LABELS, 1);
+                if (!tracked) {
```
hbs: Is this case actually plausible? How can tracked be true if the metadata were not known?

warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
```diff
@@ -119,6 +119,11 @@
    * Number of distinct owners known by a Directory
    */
   public static final String SENSISION_CLASS_CONTINUUM_DIRECTORY_OWNERS = "warp.directory.owners";
+  
+  /**
+   * Number of distinct classes per owner known by a Directory
+   */
+  public static final String SENSISION_CLASS_CONTINUUM_DIRECTORY_GTS_PERAPP = "warp.directory.gts.perapp";
```
hbs: The same metric needs to be updated in the standalone version of Directory.

warp10/src/main/java/io/warp10/continuum/store/Directory.java
aurrelhebert: True, I can remove the tracked variable in this case

PR #577, opened 2019-10-22
Clearer, standardized and more secure error messages.

ftence:


warp10/src/main/java/io/warp10/ThrowableUtils.java
```diff
@@ -16,76 +16,119 @@
...
+   * @param messages The collection of message to use for the generation
+   * @return The error message.
+   */
+  public static StringBuilder buildErrorMessage(Collection<String> messages) {
+    StringBuilder errorMessageBuilder = new StringBuilder();
+
+
+    boolean nextMessage = false;
+    for (String message: messages) {
+      if (nextMessage) {
```
hbs: you could simply have checked that the StringBuilder size was not zero

PR #579, opened 2019-10-28
Directory unregister

hbs:


warp10/src/main/java/io/warp10/continuum/store/Directory.java
```diff
@@ -1071,6 +1073,27 @@ public void run() {
...
+    
+    if (this.register) {
+      final Directory self = this;
+      Runtime.getRuntime().addShutdownHook(new Thread() {
+        @Override
+        public void run() {
+          try {
+            LOG.info("Unregistering from ZooKeeper.");
+            self.sd.unregisterService(self.instance);
+            self.sd.close();
```
ftence: `close` should suffice as it unregisters all services: https://github.com/Netflix/curator/blob/curator-1.3.3/curator-x-discovery/src/main/java/com/netflix/curator/x/discovery/details/ServiceDiscoveryImpl.java#L128

warp10/src/main/java/io/warp10/continuum/store/Directory.java
```diff
@@ -1071,6 +1073,27 @@ public void run() {
   @Override
   public void run() {
     
+    //
+    // Add a shutdown hook to unregister Directory
+    //
+    
+    if (this.register) {
+      final Directory self = this;
+      Runtime.getRuntime().addShutdownHook(new Thread() {
```
ftence: Could be nice to remove this hook in the `finally` clause in line 1183.

hbs:
PTAL

PR #586, opened 2019-11-05
Added configuration of max labels/attributes sizes

hbs:


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -2710,6 +2711,11 @@ public static GTSEncoder parse(GTSEncoder encoder, String str, Map<String,String
       ignoredCount.addAndGet(1);
     }
     
+    // Check labels/attributes sizes, subtract 6 to account for '// {} '
+    if (str.length() - 6 > MetadataUtils.SIZE_THRESHOLD && !MetadataUtils.validateMetadata(encoder.getMetadata())) {
```
ftence: `str.length() - 6 - valuestr.length()` should be a more precise estimation of the labels and attributes size, especially for long String values.

hbs: PTAL

hbs:
PTAL

PR #587, opened 2019-11-05
Attributes update

hbs:


warp10/src/main/java/io/warp10/continuum/store/Directory.java
```diff
@@ -1761,10 +1773,38 @@ public void run() {
...
+                for (Entry<String,String> attr: metadata.getAttributes().entrySet()) {
+                  hasChanged = true;
+                  if ("".equals(attr.getValue())) {
+                    meta.getAttributes().remove(attr.getKey());
+                  } else {
+                    meta.putToAttributes(attr.getKey(), attr.getValue());
+                  }
+                }
+                
+                if (hasChanged) {
```
ftence: Could use `metadata.getAttributesSize() > 0` instead of `hasChanged` and remove that variable

hbs: The attributes should be updated unconditionally any way...

hbs:
PTAL

PR #588, opened 2019-11-07
Attributes are lost when a sorted GTS is time-clipped

Giwi:


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -1825,6 +1825,7 @@ public static final GeoTimeSerie subSerie(GeoTimeSerie gts, long starttimestamp,
       
       if (copyLabels) {
         subgts.setLabels(gts.getLabels());
+        subgts.getMetadata().setAttributes(new HashMap<String,String>(gts.getMetadata().getAttributes()));
```
hbs: Attributes may be null, this will lead to an NPE

hbs: My mistake, attributes are set

PR #598, opened 2019-11-28
Initial commit of skip/count/sample support

hbs:
This PR adds the following features to the fetch process:

- ability to specify a number of datapoints to fetch (count) within a given time range
- ability to skip a certain number of datapoints prior to returning results
- ability to sample the fetched datapoints


warp10/src/main/java/io/warp10/standalone/InMemoryChunkSet.java
```diff
@@ -266,7 +272,7 @@ public GTSEncoder fetchEncoder(long now, long timespan, int preBoundary, int pos
     int nowchunk = chunk(now) + this.chunkcount;
     
     // Compute the first timestamp (included)
-    long firstTimestamp = now - timespan + 1;
+    long firstTimestamp = then;
```
ftence: Not sure that `then` should be included, still that convention problem.

warp10/src/main/java/io/warp10/standalone/StandaloneStoreClient.java
```diff
@@ -97,15 +98,26 @@ public GTSDecoderIterator fetch(final ReadToken token, final List<Metadata> meta
...
-
+    
+    if (count < -1L) {
+      count = -1L;
+    }
+    
+    //
+    // If we are fetching up to Long.MIN_VALUE, then don't fetch a pre boundary
+    //
+    if (Long.MIN_VALUE == then && preBoundary > 0) {
```
ftence: Useless test on `preBoundary`.

warp10/src/main/java/io/warp10/continuum/egress/HBaseStoreClient.java
```diff
@@ -229,12 +244,13 @@ public GTSDecoderIterator fetch(final ReadToken token, final List<Metadata> meta
     if (metadatas.size() < ParallelGTSDecoderIteratorWrapper.getMinGTSPerScanner() || !ParallelGTSDecoderIteratorWrapper.useParallelScanners()) {
       if (optimized) {
         //return new SlicedRowFilterGTSDecoderIterator(now, timespan, metadatas, this.conn, this.tableName, this.colfam, this.keystore, metadatas.size() <= blockcacheThreshold);
+        long timespan = count > 0 ? -count : (now - then + 1);
```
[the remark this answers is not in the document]
hbs: If 'then' is included, timespan should be now - then + 1

warp10/src/main/java/io/warp10/continuum/store/MultiScanGTSDecoderIterator.java
```diff
@@ -208,42 +217,19 @@ public boolean hasNext() {
...
-    
-      bb.putLong(Long.MAX_VALUE - modulus);
-    } else {
-      // Long.MAX_VALUE - Long.MIN_VALUE, the beginning of time...
-      bb.putLong(0xffffffffffffffffL);
-      endAtMinlong = true;
-    }
+    boolean endAtMinlong = Long.MIN_VALUE == then;
+
+    bb.putLong(Long.MAX_VALUE - then);
```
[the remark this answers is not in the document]
hbs: It will eventually be removed, but this should not happen in this PR.

warp10/src/main/java/io/warp10/continuum/store/MultiScanGTSDecoderIterator.java
```diff
@@ -423,8 +409,19 @@ public GTSDecoder next() {
           while(nvalues > 0 && decoder.next()) {
             long timestamp = decoder.getTimestamp();
                         
-            if (preBoundaryScan || postBoundaryScan || (timestamp <= now && (timespan < 0 || (timestamp > (now - timespan))))) {
+            if (preBoundaryScan || postBoundaryScan || (timestamp <= now && timestamp >= then)) {
```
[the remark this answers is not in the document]
hbs: 'then' is included, not excluded

warp10/src/main/java/io/warp10/continuum/store/StoreClient.java
```diff
@@ -29,6 +29,21 @@
   public void store(GTSEncoder encoder) throws IOException;
   public void archive(int chunk, GTSEncoder encoder) throws IOException;
   public long delete(WriteToken token, Metadata metadata, long start, long end) throws IOException;
-  public GTSDecoderIterator fetch(ReadToken token, final List<Metadata> metadatas, final long now, final long timespan, boolean fromArchive, boolean writeTimestamp, final int preBoundary, final int postBoundary) throws IOException;
+  /**
+   * 
+   * @param token Read token to use for reading data
+   * @param metadatas List of Metadata for the GTS to fetch
+   * @param now End timestamp (included)
+   * @param then Start timestamp (included)
```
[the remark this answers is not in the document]
hbs: start should be included, otherwise there is no way to fetch from Long.MIN_VALUE

warp10/src/main/java/io/warp10/standalone/InMemoryChunkSet.java
hbs: It should, see above.

hbs:
PTAL

warp10/src/main/java/io/warp10/continuum/store/MultiScanGTSDecoderIterator.java
ftence: `then` will be excluded, which is not what we want.

hbs:
PTAL

warp10/src/main/java/io/warp10/continuum/store/MultiScanGTSDecoderIterator.java
```diff
@@ -208,42 +217,19 @@ public boolean hasNext() {
...
-        modulus = (modulus - (modulus % Constants.DEFAULT_MODULUS)) - Constants.DEFAULT_MODULUS;
-      }
-    
-      bb.putLong(Long.MAX_VALUE - modulus);
-    } else {
-      // Long.MAX_VALUE - Long.MIN_VALUE, the beginning of time...
-      bb.putLong(0xffffffffffffffffL);
-      endAtMinlong = true;
-    }
+    boolean endAtMinlong = Long.MIN_VALUE == then;
```
ftence: Not sure `endAtMinlong` is still needed because in the case of `preBoundaryScan`, both keys should be 0xFFFFFFFFFFFFFFFFL plus a 0x0 byte, thus fetching nothing.

warp10/src/main/java/io/warp10/continuum/store/MultiScanGTSDecoderIterator.java
```diff
@@ -271,12 +257,8 @@ public boolean hasNext() {
...
     } else {
       scan.setStartRow(startkey);
-      if (endAtMinlong) {
-        // When the end timestamp should have been after MIN_LONG add a trailing 0x00
-        scan.setStopRow(Arrays.copyOf(endkey, endkey.length + 1));
-      } else {
-        scan.setStopRow(endkey);
-      }
+      // Then endkey is 'endkey' plus a trailing 0x00 so we include 'endkey'
+      scan.setStopRow(Arrays.copyOf(endkey, endkey.length + 1));
```
ftence: This line is OK, but the preBoundaryScan will also fetch `endkey`. I think `endkey` should always have a trailing 0x00.

warp10/src/main/java/io/warp10/standalone/StandaloneStoreClient.java
```diff
@@ -392,44 +401,19 @@ public boolean hasNext() {
...
-              } else {
-                bb.putLong(0xffffffffffffffffL);
-              }
-            }            
+            stoprow = new byte[startrow.length];
+            bb = ByteBuffer.wrap(stoprow).order(ByteOrder.BIG_ENDIAN);
+            bb.put(Constants.HBASE_RAW_DATA_KEY_PREFIX);
+            bb.putLong(metadatas.get(idx).getClassId());
+            bb.putLong(metadatas.get(idx).getLabelsId());                            
+            bb.putLong(Long.MAX_VALUE - then);
```
ftence: Either change the key or make sure `then` in included in the "core" fetch and not the pre-boundary one.

warp10/src/main/java/io/warp10/continuum/store/MultiScanGTSDecoderIterator.java
hbs: The use of endAtMinLong avoids a larger allocation if it is true, so let's keep it.

warp10/src/main/java/io/warp10/standalone/StandaloneStoreClient.java
hbs: The previous comparison with stoprow actually takes care of this.

hbs:
Guess what? PTAL!

warp10/src/main/java/io/warp10/continuum/store/MultiScanGTSDecoderIterator.java
ftence: Agreed, but if you care about allocations, you should always add a trailing 0x00 for endkey instead of copying it each time ;).
Also, in the case of preBoundary and endAtMinlong you could `scan.setStopRow(endkey);` to avoid a new allocation. In the end, it doesn't really matter as this case should never happen because of previous tests.

hbs:
PTAL

PR #601, opened 2019-12-02
Initial implementation of IngressPlugin support

hbs:


KannarFr:
What about have a storage properties accessible from "plugin master class" that enables to storage plugin state on kafka? Like schema registry stored on a kafka topic in my case? Sure I can do it from the plugin impl. Just wanted your opinion on a generic way for ingresses plugins to store a state.

hbs:
IngressPlugins do not require that Kafka be present as they can be deployed on the Standalone version, so this is something that only specific plugins should handle.

KannarFr:
Got it. Thanks.

warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
```diff
@@ -1758,7 +1803,7 @@ public void close() throws Exception {
             throw error.get();
           }
         } finally {
-          try { shufflediterator.close(); } catch (Exception e) {}
+          try { shufflediterator.close(); } catch (Exception e) {}
```
ftence: Unnecessary spaces.

hbs:
> Would it make sense to chain several plugins instead of restricting to only one?

If daisy chaining plugins is needed, the idea is to have a daisy chaining plugin, but from the Ingress point of view, only a single plugin is known. The daisy chaining plugin could decide to implement various strategies for deciding if processing should continue or not for example. The general principle is to keep Ingress simple and delegate any advanced stuff to plugins.

ftence:
Good point.

PR #613, opened 2019-12-11
Rewrite FETCH and fetch endpoint parameters handling

ftence:
To account for new functionalities and standardization.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/egress/EgressFetchHandler.java
+++ b/warp10/src/main/java/io/warp10/continuum/egress/EgressFetchHandler.java
@@ -18,7 +18,6 @@ package io.warp10.continuum.egress;
 
 import io.warp10.ThrowableUtils;
 import io.warp10.continuum.Configuration;
-import io.warp10.continuum.TimeSource;
 import io.warp10.continuum.Tokens;
 import io.warp10.continuum.gts.GTSDecoder;
 import io.warp10.continuum.gts.GTSEncoder;
@@ -41,6 +40,7 @@ import io.warp10.crypto.OrderPreservingBase64;
 import io.warp10.crypto.SipHashInline;
 import io.warp10.quasar.token.thrift.data.ReadToken;
 import io.warp10.script.WarpScriptException;
+import io.warp10.script.functions.FETCH;
 import io.warp10.sensision.Sensision;
 
 import java.io.BufferedReader;
@@ -61,7 +61,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.NoSuchElementException;
@@ -80,8 +79,6 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import org.apache.commons.codec.binary.Hex;
-import org.apache.commons.lang3.JavaVersion;
-import org.apache.commons.lang3.SystemUtils;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.thrift.TDeserializer;
 import org.apache.thrift.TException;
@@ -91,14 +88,8 @@ import org.boon.json.JsonSerializer;
 import org.boon.json.JsonSerializerFactory;
 import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.handler.AbstractHandler;
-import org.joda.time.Duration;
-import org.joda.time.Instant;
-import org.joda.time.MutablePeriod;
-import org.joda.time.Period;
-import org.joda.time.ReadWritablePeriod;
 import org.joda.time.format.DateTimeFormatter;
 import org.joda.time.format.ISODateTimeFormat;
-import org.joda.time.format.ISOPeriodFormat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -179,40 +170,45 @@ public class EgressFetchHandler extends AbstractHandler {
       //
       
       resp.setHeader("Access-Control-Allow-Origin", "*");
-
-      String start = null;
-      String stop = null;
       
       long now = Long.MIN_VALUE;
-      long timespan = 0L;
       long then = Long.MIN_VALUE;
       long count = -1;
       long skip = 0;
       double sample = 1.0D;
+      int preBoundary = 0;
+      int postBoundary = 0;
 
+      String startParam = null;
+      String stopParam = null;
       String nowParam = null;
+      String endParam = null;
       String timespanParam = null;
       String dedupParam = null;
       String showErrorsParam = null;
       String countParam = null;
       String skipParam = null;
       String sampleParam = null;
-      
+      String preBoundaryParam = null;
+      String postBoundaryParam = null;
+
       if (splitFetch) {
         nowParam = req.getHeader(Constants.getHeader(Configuration.HTTP_HEADER_NOW_HEADERX));
         timespanParam = req.getHeader(Constants.getHeader(Configuration.HTTP_HEADER_TIMESPAN_HEADERX));
         showErrorsParam = req.getHeader(Constants.getHeader(Configuration.HTTP_HEADER_SHOW_ERRORS_HEADERX));
       } else {
-        start = req.getParameter(Constants.HTTP_PARAM_START);
-        stop = req.getParameter(Constants.HTTP_PARAM_STOP);
-        
+        startParam = req.getParameter(Constants.HTTP_PARAM_START);
+        stopParam = req.getParameter(Constants.HTTP_PARAM_STOP);
         nowParam = req.getParameter(Constants.HTTP_PARAM_NOW);
+        endParam = req.getParameter(Constants.HTTP_PARAM_END);
         timespanParam = req.getParameter(Constants.HTTP_PARAM_TIMESPAN);
         dedupParam = req.getParameter(Constants.HTTP_PARAM_DEDUP);      
         showErrorsParam = req.getParameter(Constants.HTTP_PARAM_SHOW_ERRORS);
         countParam = req.getParameter(Constants.HTTP_PARAM_COUNT);
         skipParam = req.getParameter(Constants.HTTP_PARAM_SKIP);
         sampleParam = req.getParameter(Constants.HTTP_PARAM_SAMPLE);
+        preBoundaryParam = req.getParameter(Constants.HTTP_PARAM_PREBOUNDARY);
+        postBoundaryParam = req.getParameter(Constants.HTTP_PARAM_POSTBOUNDARY);
       }
           
       String maxDecoderLenParam = req.getParameter(Constants.HTTP_PARAM_MAXSIZE);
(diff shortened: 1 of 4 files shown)


warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+    //
+
+    Long startTs = null;
+
+    if (start instanceof Long) {
+      // Simple case: start is a long
+      startTs = (long) start;
+    } else if (start instanceof String) {
+      // If it's a string, it may be the string representation of a Long, a ISO8601 date or 'now'.
+      if ("now".equals(start)) {
```
hbs: Make this a constant.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+        startTs = now;
+      } else {
+        try {
+          startTs = Long.valueOf((String) start);
```
hbs: Could use Long.parseLong

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+    //
+
+    Long endTs = null;
+
+    if (end instanceof Long) {
+      // Simple case: start is a long
+      endTs = (long) end;
+    } else if (end instanceof String) {
+      // If it's a string, it may be the string representation of a Long, a ISO8601 date or 'now'.
+      if ("now".equals(start)) {
```
hbs: ditto

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+        endTs = now;
+      } else {
+        try {
+          endTs = Long.valueOf((String) end);
```
hbs: ditto

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+
+      //
+      // Cast or convert timespan to a numeric one.
+      //
+      if (timespan instanceof Long) {
+        numericTimespan = (Long) timespan;
+      } else if (timespan instanceof String) {
+        // If it's a string, it may be the string representation of a Long or a ISO8601 duration.
+        // Try the former then the latter if it fails.
+        try {
```
hbs: You could check if it starts with a 'P', if not it cannot be an ISO8601 duration

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+        } catch (NumberFormatException nfe) {
+          // Not string representation of a Long, try ISO8601 duration.
+          try {
+            ReadWritablePeriod period = new MutablePeriod();
+
+            ISOPeriodFormat.standard().getParser().parseInto(period, (String) timespan, 0, Locale.US);
+
+            Period p = period.toPeriod();
+
+            if (p.getMonths() != 0 || p.getYears() != 0) {
```
hbs: Actually since we do have either a start or an end, we could technically support months and years as you are not interested in determining how long the period is but simply compute an offset to an existing timestamp.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+          } catch (WarpScriptException | IllegalArgumentException e) {
+            throw new WarpScriptException("Invalid format for parameter '" + timespanParamName + "'.", e);
+          }
+        }
+      } else {
+        // If timespan is not null and not a Long nor a String, throw an error.
+        throw new WarpScriptException("Invalid format for parameter '" + timespanParamName + "'.");
+      }
+
+      if (startTs == null) {
```
hbs: Coherency, use null == startTs

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      }
+    }
+
+    // Check that at least endTs is defined.
+    if (null == endTs) {
+      throw new WarpScriptException("Missing '" + endParamName + "' or '" + startParamName + "' and '" + timespanParamName + "' parameter.");
+    }
+
+    // Make sure startTs is defined.
+    if(null == startTs) {
```
hbs: Missing space

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      if (null == count) {
+        throw new WarpScriptException("Invalid time range specification: '" + countParamName + "' is mandatory if neither '" + startParamName + "' and '" + timespanParamName + "' are not specified.");
+      } else {
+        // Fetch with end and count: start is set to the beginnings of time.
+        startTs = Long.MIN_VALUE;
+      }
+    }
+
+    return new Long[]{startTs, endTs, numericTimespan};
```
hbs: missing space

warp10/src/main/java/io/warp10/continuum/egress/EgressFetchHandler.java
```diff
@@ -237,77 +233,48 @@ public void handle(String target, Request baseRequest, HttpServletRequest req, H
...
-            throw new IOException("No support for ambiguous durations containing years or months, please convert those to days.");
-          }
+      //
+      // Handle aliases
+      //
 
-          Duration duration = p.toDurationFrom(new Instant());
+      // negative timespan is count
+      if (null != timespanParam && null != countParam) {
+        throw new WarpScriptException("Cannot specify both '" + Constants.HTTP_PARAM_COUNT + "' and negative '" + Constants.HTTP_PARAM_TIMESPAN + "'.");
```
hbs: Should it be a WarpScriptException or an IOException?

warp10/src/main/java/io/warp10/continuum/egress/EgressFetchHandler.java
```diff
@@ -759,34 +732,34 @@ public void close() throws Exception {
...
+            try(GTSDecoderIterator iterrsc = storeClient.fetch(rtoken, metas, now, then, count, skip, sample, writeTimestamp, preBoundary, postBoundary)) {
               GTSDecoderIterator iter = iterrsc;
                           
               if (unpack) {
                 iter = new UnpackingGTSDecoderIterator(iter, suffix);
-                timespan = Long.MIN_VALUE + 1;
+                count = Long.MAX_VALUE;
               }
               
               if("text".equals(format)) {
```
hbs: Please add a long missing whitespace

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      // Check that endTs and startTs are not both already defined.
+      if (startTs != null && endTs != null) {
+        throw new WarpScriptException("Invalid time range specification: '" + startParamName + "', '" + endParamName + "' and '" + timespanParamName + "' cannot all be defined. Only 2 out of those 3 parameters should be defined.");
+      }
+
+      //
+      // Cast or convert timespan to a numeric one.
+      //
+      if (timespan instanceof Long) {
+        numericTimespan = (Long) timespan;
```
hbs: Timespan is assumed to be >= 0 in this method, that is because the calling path you set up ensures so, but maybe this should be checked here and after line 1106.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      } else {
+        // If timespan is not null and not a Long nor a String, throw an error.
+        throw new WarpScriptException("Invalid format for parameter '" + timespanParamName + "'.");
+      }
+
+      if (startTs == null) {
+        // In that case startTs is not defined, so we compute it.
+
+        // Check edge case
+        if (0 == numericTimespan && Long.MAX_VALUE == endTs) {
```
hbs: 0L will avoid an implicit cast

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+          // No need to check for overflow for +1 because this edge case has already been checked.
+          startTs = Math.subtractExact(endTs, numericTimespan) + 1;
+        } catch (ArithmeticException ae) {
+          startTs = Long.MIN_VALUE;
+        }
+      } else { // endTs == null
+        // In that case endTs is not defined, so we compute it.
+
+        // Check edge case
+        if (0 == numericTimespan && Long.MIN_VALUE == startTs) {
```
hbs: 0L

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1099,4 +984,197 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      if (startTs == null) {
+        // In that case startTs is not defined, so we compute it.
+
+        // Check edge case
+        if (0 == numericTimespan && Long.MAX_VALUE == endTs) {
+          throw new WarpScriptException("Cannot set '" + timespanParamName + "' to 0 and '" + endParamName + "' to MAX_VALUE.");
+        }
+
+        try {
+          // No need to check for overflow for +1 because this edge case has already been checked.
```
hbs: Specify where, it is not obvious from reading the code around the comment

warp10/src/main/java/io/warp10/script/functions/FETCH.java
ftence: In this case this is totally equivalent. But in case `startTs` is changed to a `long` in the future, `Long.parseLong` should be preferred.

warp10/src/main/java/io/warp10/continuum/egress/EgressFetchHandler.java
ftence: Actually better to return SC_BAD_REQUEST code with the error message.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
ftence: It should be possible but it's not trivial. Noted as TODO.

PR #616, opened 2019-12-16
Add optional index or counter to loops.

ftence:
Performance impact not already assessed.


warp10/src/main/java/io/warp10/script/functions/FOREACH.java
```diff
@@ -65,8 +74,11 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
     }
     
     if (obj instanceof List) {
-      for (Object o: ((List<Object>) obj)) {
-        stack.push(o);
+      for (int i = 0; i < ((List) obj).size(); i++) {
```
hbs: Why not keep the for (Object o: ...) loop and use an additional variable to use as the counter.

ftence:
I added a counter for all type in FOREACH, which is much more logical.

warp10/src/main/java/io/warp10/script/functions/LMAP.java
```diff
@@ -27,26 +27,41 @@
...
- * macro and list are popped out of the stack
- * macro if called for each element of 'list' with the index of the current element and the element itself on the stack
+ * The macro and the list are popped out of the stack.
+ * The macro is called for each element of 'list' with the index of the current element, by default, and the element itself on the stack.
+ *
  */
 public class LMAP extends NamedWarpScriptFunction implements WarpScriptStackFunction {
-  
+
+  private boolean flat;
```
hbs: Could make it final

PR #620, opened 2019-12-18
Fixes #619 - Corrected URL decoding to use WarpURLDecoder

### LINKED ISSUE #619: Incorrect decoding of class and labels selectors
The decoding of class and label selectors incorrectly interprets the '+' as a white space. The rest of Warp 10 only considers %hh encoding, so any occurrence of '+' in a string which contains a '%' should be replaced by %2B prior to decoding the string

hbs:


warp10/src/main/java/io/warp10/WarpURLDecoder.java
```diff
@@ -0,0 +1,41 @@
...
+    if (!input.contains("%")) {
+      return input;
+    }
+    
+    // The input contains a '%', if it contains a '+' we need to replace the '+' with '%2B'
+    // so decoding is correct.
+    
+    if (input.contains("+")) {
+      input = input.replaceAll("+", "%2B");
+    }
```
pi-r-p: Use input.replaceAll("\\+", "%2B") or input.replace() which is optimized for literal targets, otherwise it can't work.

Test with this WarpScript:
```
  '.*[!%25&\(\)\*+/<=>\[\]^\{\|\}~].*'  MATCHER 'ForbiddenCharacters' STORE
```
You will have exception `Dangling meta character '+' near index 0 + ^`

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -2589,11 +2590,7 @@ public static GTSEncoder parse(GTSEncoder encoder, String str, Map<String,String
       
       //if (name.contains("%")) {
       if (-1 != UnsafeString.indexOf(name, '%')) {
-        try {      
-          name = URLDecoder.decode(name, StandardCharsets.UTF_8.name());
-        } catch (UnsupportedEncodingException uee) {
-          // Can't happen, we're using UTF-8
-        }      
+        name = WarpURLDecoder.decode(name, StandardCharsets.UTF_8);
```
ftence: Don't check for '%' in the string, `WarpURLDecoder.decode` already does.

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -2749,11 +2746,7 @@ public static Object parseValue(String valuestr) throws ParseException {
           || ('"' == firstChar && valuestr.endsWith("\""))) {
         value = valuestr.substring(1, valuestr.length() - 1);
         if (((String)value).contains("%")) {
-          try {
-            value = URLDecoder.decode((String) value, StandardCharsets.UTF_8.name());
-          } catch (UnsupportedEncodingException uee) {
-            // Can't happen, we're using UTF-8
-          }
+          value = WarpURLDecoder.decode((String) value, StandardCharsets.UTF_8);
```
ftence: Don't check for '%' in the string, `WarpURLDecoder.decode` already does.

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -689,9 +690,7 @@ public void exec(String line) throws WarpScriptException {
               String str = stmt.substring(1, stmt.length() - 1);
 
               if (-1 != UnsafeString.indexOf(str, '%')) {
-                // replace occurrences of '+' with '%2B'
-                str = str.replaceAll("\\+", "%2B");
-                str = URLDecoder.decode(str, StandardCharsets.UTF_8.name());
+                str = WarpURLDecoder.decode(str, StandardCharsets.UTF_8);
```
ftence: Don't check for '%' in the string, `WarpURLDecoder.decode` already does.

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -65,6 +65,7 @@
 
 import io.warp10.CapacityExtractorOutputStream;
 import io.warp10.DoubleUtils;
+import io.warp10.WarpURLDecoder;
```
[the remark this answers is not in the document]
hbs: It is still used in a legacy parse_regexp method which will vanish soon....

hbs:
PTAL

PR #631, opened 2020-01-14
fix median and percentile aggregators bugs when used as reducer

pi-r-p:
- median aggregator was not usable as reducer, or returned the wrong result.
- position median was done on raw hhcode

example of non working code : 
https://snapshot.senx.io/00059c2130c8d5f8-0-4-5f8666e22df55aa3
https://snapshot.senx.io/00059c21362565e0-0-7-025b60d11920474c

### DIFF
--- a/warp10/src/main/java/io/warp10/script/aggregator/Median.java
+++ b/warp10/src/main/java/io/warp10/script/aggregator/Median.java
@@ -100,9 +100,7 @@ public class Median extends NamedWarpScriptFunction implements WarpScriptAggrega
     // singleton case
     //
     if (1 == nonNullLength) {
-      return new Object[] {
-          tick, locations[indices[0]], elevations[indices[0]], values[indices[0]]
-      };
+      return new Object[] {tick, locations[indices[0]], elevations[indices[0]], values[indices[0]]};
     } else {
       if (0 == nonNullLength % 2) {
         //
--- a/warp10/src/main/java/io/warp10/script/aggregator/Median.java
+++ b/warp10/src/main/java/io/warp10/script/aggregator/Median.java
@@ -26,20 +26,21 @@ import io.warp10.script.WarpScriptMapperFunction;
 import io.warp10.script.WarpScriptReducerFunction;
 import io.warp10.script.binary.EQ;
 
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.Comparator;
 
 /**
  * Return the median of the values on the interval.
- * The returned location will be the median of all locations.
- * The returned elevation will be the median of all elevations.
+ * If median data point has an associated location and elevation, return it
+ * If forbidNulls and null among inputs, the function will raise an exception.
  */
 public class Median extends NamedWarpScriptFunction implements WarpScriptAggregatorFunction, WarpScriptMapperFunction, WarpScriptBucketizerFunction, WarpScriptReducerFunction {
 
-  public Median(String name) {
+  private final boolean forbidNulls;
+
+  public Median(String name, boolean forbidNulls) {
     super(name);
+    this.forbidNulls = forbidNulls;
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


hbs:
Your modification computes the median for latitude and longitude independently, this is awkward.

pi-r-p:
Indeed... Better than the median of raw HHcodes. Another way would be to compute the median of values, and keep the position of the median point (or centroid in case of even number of points).
Another way is that this mapper/reducer do not return any position.

By the way, reducer.percentile is also buggy.

hbs:
Retaining the location of the actual median point would be more logical, the only issue would be when the median is the mean of two elements and only one of those elements has a location.

warp10/src/main/java/io/warp10/script/aggregator/Median.java
```diff
@@ -26,147 +26,107 @@
...
+      public int compare(Integer idx1, Integer idx2) {
+        if (null == values[idx1] && null == values[idx2]) {
+          return 0;
+        } else if (null == values[idx1] || null == values[idx2]) {
+          return null == values[idx1] ? 1 : -1;
+        } else if (values[idx1] instanceof Number && values[idx2] instanceof Number) {
+          return EQ.compare((Number) values[idx1], (Number) values[idx2]);
         } else {
-          throw new WarpScriptException(this.getName() + " cannot compute median of non numeric values.");
+          throw new RuntimeException("MEDIAN can only operate on numeric Geo Time Series.");
```
hbs: Retrieve the name from outside the comparator.

warp10/src/main/java/io/warp10/script/aggregator/Median.java
```diff
@@ -26,147 +26,107 @@
...
     //
-    ArrayList<Number> validValues = new ArrayList<Number>(values.length);
-    for (int i = 0; i < values.length; i++) {
-      if (null != values[i]) {
-        if (values[i] instanceof Number) {
-          if (!(values[i] instanceof Double && Double.isNaN((Double) values[i]))) {
-            validValues.add((Number) values[i]);
-          }
+    Arrays.sort(indices, new Comparator<Integer>() {
+      @Override
```
hbs: Explain strategy for sorting in the presence of nulls

warp10/src/main/java/io/warp10/script/aggregator/Median.java
```diff
@@ -26,147 +26,107 @@
...
+      if (0 == nonNullLength % 2) {
+        //
+        // even number of non null values, return mean of both values.
+        // If there is a location for both points, return centroid of locations
+        // If there is an elevation for both points, return mean of elevations
+        //
+        int low = indices[nonNullLength / 2 - 1];
+        int high = indices[nonNullLength / 2];
+        median = (((Number) values[low]).doubleValue() + ((Number) values[high]).doubleValue()) / 2.0D;
+        if (GeoTimeSerie.NO_ELEVATION != elevations[low] && GeoTimeSerie.NO_ELEVATION != elevations[high]) {
```
hbs: If there is only one location (same goes for elevation), wouldn't it be better to return that location (resp. elevation)?

warp10/src/main/java/io/warp10/script/aggregator/Median.java
```diff
@@ -26,147 +26,107 @@
...
-    // Could be Long or Double, or a mix, if used as a reducer.
-    //
-    Collections.sort(validValues, new Comparator<Number>() {
-      @Override
-      public int compare(Number o1, Number o2) {
-        return EQ.compare(o1, o2);
-      }
     });
 
+    long location = 0;
```
hbs: Shouldn't that be GeoTimeSeries.NO_LOCATION?

warp10/src/main/java/io/warp10/script/aggregator/Median.java
```diff
@@ -26,147 +26,107 @@
...
-      int len = validLat.size();
-      if (0 == len % 2) {
-        medianLatitude = (validLat.get(len / 2) + validLat.get(len / 2 - 1)) / 2.0D;
-        medianLongitude = (validLong.get(len / 2) + validLong.get(len / 2 - 1)) / 2.0D;
-      } else {
-        medianLatitude = validLat.get(len / 2);
-        medianLongitude = validLong.get(len / 2);
-      }
-      location = GeoXPLib.toGeoXPPoint(medianLatitude, medianLongitude);
+    if (nullCounter != 0 && this.forbidNulls) {
```
hbs: 0 != nullCounter

warp10/src/main/java/io/warp10/script/aggregator/Median.java
pi-r-p: To be discussed... What is your preference ?

pi-r-p: fixed.

pi-r-p:
- Fixed percentile aggregator with the same method. 
- adds "forbid-nulls" variants to keep the old behaviour.

warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
```diff
@@ -31,150 +31,175 @@
...
       if (0 == n && this.percentile < pn) {
-        return new Object[] { ticks[indices[0]], locations[indices[0]], elevations[indices[0]], values[indices[0]] };
-      } else if (m == indices.length - 1 && this.percentile > pm) {
-        return new Object[] { ticks[indices[m]], locations[indices[m]], elevations[indices[m]], values[indices[m]] };
+        return new Object[]{ticks[indices[0]], locations[indices[0]], elevations[indices[0]], values[indices[0]]};
+      } else if (m == nonNullLength - 1 && this.percentile > pm) {
+        return new Object[]{ticks[indices[m]], locations[indices[m]], elevations[indices[m]], values[indices[m]]};
       } else if (pn == this.percentile) {
-        return new Object[] { ticks[indices[n]], locations[indices[n]], elevations[indices[n]], values[indices[n]] };
+        return new Object[]{ticks[indices[n]], locations[indices[n]], elevations[indices[n]], values[indices[n]]};
```
ftence: Missing space after `Object[]`.

warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
```diff
@@ -31,150 +31,175 @@
...
+      int m = (int) Math.floor(0.5 + this.percentile * nonNullLength / 100.0) - 1;
+
+      double pn = (100.0 / nonNullLength) * (n + 1 - 0.5D);
+      double pm = (100.0 / nonNullLength) * (m + 1 - 0.5D);
 
       if (0 == n && this.percentile < pn) {
-        return new Object[] { ticks[indices[0]], locations[indices[0]], elevations[indices[0]], values[indices[0]] };
-      } else if (m == indices.length - 1 && this.percentile > pm) {
-        return new Object[] { ticks[indices[m]], locations[indices[m]], elevations[indices[m]], values[indices[m]] };
+        return new Object[]{ticks[indices[0]], locations[indices[0]], elevations[indices[0]], values[indices[0]]};
```
ftence: Missing space after `Object[]`.

warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
```diff
@@ -31,150 +31,175 @@
...
+      } else if (m == nonNullLength - 1 && this.percentile > pm) {
+        return new Object[]{ticks[indices[m]], locations[indices[m]], elevations[indices[m]], values[indices[m]]};
       } else if (pn == this.percentile) {
-        return new Object[] { ticks[indices[n]], locations[indices[n]], elevations[indices[n]], values[indices[n]] };
+        return new Object[]{ticks[indices[n]], locations[indices[n]], elevations[indices[n]], values[indices[n]]};
       } else if (pm == this.percentile) {
-        return new Object[] { ticks[indices[m]], locations[indices[m]], elevations[indices[m]], values[indices[m]] };
+        return new Object[]{ticks[indices[m]], locations[indices[m]], elevations[indices[m]], values[indices[m]]};
```
ftence: Missing space after `Object[]`.

warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
```diff
@@ -31,150 +31,175 @@
...
       if (!(value instanceof Number)) {
         throw new WarpScriptException("Invalid parameter for " + getName());
       }
-      
+
       double percentile = ((Number) value).doubleValue();
-      
+
       if (percentile < 0.0D || percentile > 100.0D) {
         throw new WarpScriptException("Invalid percentile, MUST be between 0 and 100.");
```
ftence: Add function name.

pi-r-p:
all fixed

warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
```diff
@@ -67,7 +67,7 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       double percentile = ((Number) value).doubleValue();
 
       if (percentile < 0.0D || percentile > 100.0D) {
-        throw new WarpScriptException("Invalid percentile, MUST be between 0 and 100.");
+        throw new WarpScriptException("Invalid percentile for" + getName() + ", MUST be between 0 and 100.");
```
hbs: missing whitespace before call to getName

warp10/src/main/java/io/warp10/script/aggregator/Median.java
```diff
@@ -100,7 +100,7 @@ public int compare(Integer idx1, Integer idx2) {
     // singleton case
     //
     if (1 == nonNullLength) {
-      return new Object[]{
+      return new Object[] {
           tick, locations[indices[0]], elevations[indices[0]], values[indices[0]]
```
hbs: Indentation should be 2 spaces, not 4.

pi-r-p:
all fixed?

PR #642, opened 2020-01-24
Add functions from HHCodeHelper

steven-gueguen:
This allows getting the center or the bounding box given one hhcode.
Allow also to compute neighboring cells.


hbs:
You can probably simplify this PR by creating an enum HHFUNC or similar and having a single function which you parameterize with an enum value to determine which HHCode to return (N,S,W,E,NE,NW,SE,SW), ditto with center and bbox.

ftence:
It could be nice to return a Long if the given HHCode is a Long and to return a String if the given HHCode is a String for the N, S, W, E, NE, NW, SE, SW functions.

steven-gueguen:
PTAL

warp10/src/main/java/io/warp10/script/functions/HHCODEFUNC.java
```diff
@@ -0,0 +1,138 @@
...
+ * Template function to interface with HHCodeHelper
+ */
+public class HHCODEFUNC extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+
+  public enum HHCodeAction {
+    NORTH, SOUTH, EAST, WEST, NORTH_EAST, NORTH_WEST, SOUTH_EAST, SOUTH_WEST, BBOX, CENTER
+  }
+
+  private HHCodeAction action;
+  private boolean isLongFormat;
```
hbs: This should not be a field, it should be a local variable updated for each invocation of the function

warp10/src/main/java/io/warp10/script/functions/HHCODEFUNC.java
```diff
@@ -0,0 +1,138 @@
...
+    this.isLongFormat = true;
+  }
+
+  @Override
+  public Object apply(WarpScriptStack stack) throws WarpScriptException {
+
+    Object o = stack.pop();
+
+    if (!(o instanceof Long)) {
+      throw new WarpScriptException(getName() + " expects resolution (even number between 2 and 30) and a boolean as the top 3 elements of the stack.");
```
hbs: Don't reference the stack, and in the present case this would be only 2 elements.

warp10/src/main/java/io/warp10/script/functions/HHCODEFUNC.java
```diff
@@ -0,0 +1,138 @@
...
+
+    if (0 != res && (res < 2 || res > 30 || (0 != (res & 1)))) {
+      throw new WarpScriptException(getName() + " expects a maximum resolution which is an even number between 2 and 30 or 0.");
+    }
+
+
+    Object hhcode = stack.pop();
+
+    long hh;
+    this.isLongFormat = true;
```
hbs: use a local variable

warp10/src/main/java/io/warp10/script/functions/HHCODEFUNC.java
```diff
@@ -0,0 +1,138 @@
...
+ */
+public class HHCODEFUNC extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+
+  public enum HHCodeAction {
+    NORTH, SOUTH, EAST, WEST, NORTH_EAST, NORTH_WEST, SOUTH_EAST, SOUTH_WEST, BBOX, CENTER
+  }
+
+  private HHCodeAction action;
+  private boolean isLongFormat;
+  private int res;
```
hbs: Same goes for res, make it a local variable

warp10/src/main/java/io/warp10/script/functions/HHCODEFUNC.java
```diff
@@ -0,0 +1,138 @@
...
+        stack.push(latlon[1]);
+        break;
+      default:
+        throw new WarpScriptException("Unknown HHCODE action");
+    }
+
+    return stack;
+  }
+
+  private Object manageFormat(long hh) {
```
hbs: drop reference of isLongFormat, simply return a long and in apply convert to string at the given resolution if the input was a string

warp10/src/main/java/io/warp10/script/functions/HHCODEFUNC.java
```diff
@@ -0,0 +1,135 @@
...
+        stack.push(latlon[1]);
+        break;
+      default:
+        throw new WarpScriptException("Unknown HHCODE action");
+    }
+
+    return stack;
+  }
+
+  private Object manageFormat(long hh, int res, boolean isLongFormat) {
```
hbs: Actually since you accept either LONG, STRING or BYTES as input it would be more logical to output a byte array if the input was a byte array. So instead of passing isLongFormat, why not simply pass the input and check its type in manageFormat?

steven-gueguen: done

warp10/src/main/java/io/warp10/script/functions/HHCODEFUNC.java
```diff
@@ -0,0 +1,135 @@
...
+    Object o = stack.pop();
+
+    if (!(o instanceof Long)) {
+      throw new WarpScriptException(getName() + " expects resolution (even number between 2 and 30).");
+    }
+
+    int res = ((Number) o).intValue();
+
+    if (0 != res && (res < 2 || res > 30 || (0 != (res & 1)))) {
+      throw new WarpScriptException(getName() + " expects a maximum resolution which is an even number between 2 and 30 or 0.");
```
hbs: Actually for a HHCode resolution can be up to 32, it is only for cells in a GeoXPShape that resolution is limited to 30.

steven-gueguen: done

PR #644, opened 2020-01-24
Filter any all

randomboolean:
Filters that keep series with any or all value that verify a comparison to a constant.
Code factoring between filters will come in a later revision.


hbs:
Can you factor them in a single FilterAnyAll with a parameter specifying the comparison to perform,

randomboolean:
They have been factored with 8b2fd83.

warp10/src/main/java/io/warp10/script/filter/FilterAny.java
```diff
@@ -0,0 +1,178 @@
...
+          }
+        }
+      }
+    }
+
+    if (complementSet) {
+      List<GeoTimeSerie> retained_ = new ArrayList<GeoTimeSerie>();
+
+      for (List<GeoTimeSerie> gtsinstances: series) {
+        for (GeoTimeSerie serie : gtsinstances) {
```
hbs: drop whitespace before ':'

randomboolean: outdated, see 5721f67

warp10/src/main/java/io/warp10/script/filter/FilterAny.java
```diff
@@ -0,0 +1,164 @@
...
+              break;
+            case BOOLEAN:
+              if (!(((Boolean) threshold).equals(val) ^ this.comparator == Comparator.EQ)) {
+                found = true;
+              }
+              break;
+          }
+        }
+
+        if (found ^ this.complementSet) {
```
hbs: Add a comment explaining the logic as it is not obvious from just looking at the code since the operator is the opposite of what the function actually checks

warp10/src/main/java/io/warp10/script/filter/FilterAny.java
```diff
@@ -0,0 +1,165 @@
...
+            case STRING:
+              found = verify(((String) threshold).compareTo(val.toString()));
+              break;
+            case BOOLEAN:
+              found = ((Boolean) threshold).equals(val) ^ this.comparator == Comparator.NE;
+              break;
+          }
+        }
+
+        // In case of wanting the complementSet, the logic is reversed.
```
hbs: Be more expressive in the description, the loop above exits when one datapoint which satisfies the comparison is found, in the case of an 'all' match, the operator checked is the opposite of the wanted one, therefore when one datapoint matches the opposite comparison, it means that not all the datapoints satisfy it and therefore, as the comparison operator is the opposite of the one we really want, it means that not all the datapoints satisfy the opposite comparison either and therefore we do not retain the series. The boolean logic is not enough of an explanation for the person discovering the code.

PR #648, opened 2020-01-25
Added support for fetching TrueType fonts via http/https

hbs:


warp10/src/main/java/io/warp10/script/processing/typography/PcreateFont.java
```diff
@@ -45,21 +80,62 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+    if (params.get(1).toString().startsWith("http://") || params.get(1).toString().startsWith("https://")) {
+      String url = resolver.resolve(params.get(1).toString());
 
+      InputStream in = null;
+      
+      try {
+        //
+        // FIXME(hbs): support WOFF/WOFF2 on top of TTF/OTF
+        //
+        in = new URL(url).openStream();
```
ftence: Fails if `url` contains a space. This can be avoided by double escaping in WarpScript ie `%2520`.
Escape in code or to be documented.

warp10/src/main/java/io/warp10/script/processing/typography/PcreateFont.java
```diff
@@ -31,9 +39,36 @@
...
+    
+    public String resolve(String url) throws WarpScriptException {
+      if (null == this.warpscript) {
+        throw new WarpScriptException("Font resolver not set, rejecting all URLs.");
+      }
+      try {
+        MemoryWarpScriptStack stack = new MemoryWarpScriptStack(null, null);
+        stack.push(url);
+        stack.exec(this.warpscript);
+        return (String) stack.pop();
```
ftence: Why use a resolver instead of a validator?
If it needs to stay a resolver, can't this macro return a special value for forbidden access, like `NULL` or an empty stack and produce a clear error message?

hbs: Simply do a FAILMSG in your resolver if you want to reject the font. The WarpScriptException will be forwarded upstream.

hbs:
PTAL

warp10/src/main/java/io/warp10/script/processing/typography/PcreateFont.java
```diff
@@ -94,6 +96,12 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
         //
         // FIXME(hbs): support WOFF/WOFF2 on top of TTF/OTF
         //
+        
+        // Encode the URL if it contains a whitespace
+        if (url.contains(" ")) {
+          url = URLEncoder.encode(url, StandardCharsets.UTF_8.name());
+        }
```
ftence: This will fail because it will also escape part of the URL which should not be escaped resulting it "https%3A%2F%2F"...
You could unconditionally do a `url.replace(" ", "%20")`
or a nice solution from [SO](https://stackoverflow.com/q/10786042#answer-25735202):
```
      String urlStr = resolver.resolve(params.get(1).toString());

      InputStream in = null;
      
      try {
        URL url = new URL(urlStr);
        URI uri = new URI(url.getProtocol(), url.getUserInfo(), url.getHost(), url.getPort(), url.getPath(), url.getQuery(), url.getRef());
        
        in = new URL(uri.toASCIIString()).openStream();
```

warp10/src/main/java/io/warp10/script/processing/typography/PcreateFont.java
```diff
@@ -45,21 +82,68 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-          ((Number) params.get(2)).floatValue(),
-          Boolean.TRUE.equals(params.get(3)),
-          params.get(4).toString().toCharArray());
-    }
+    //
+    // Handle the case of remote font files differently, passing the URL
+    // to the defined validator
+    //
+    
+    if (params.get(1).toString().startsWith("http://") || params.get(1).toString().startsWith("https://")) {
```
ftence: Could be nice to also allow `file://`.
(Tested and working with the SO solution below).

PR #659, opened 2020-02-10
Fix JSON dump test with expose owner and producer

steven-gueguen:


warp10/src/test/java/io/warp10/continuum/egress/EgressFetchHandlerTest.java
```diff
@@ -1,45 +1,74 @@
...
+  public void testJsonDumpWithExpose() throws Exception {
+    String json = jsonDumpWithExpose(true);
+    String expect = IOUtils.toString(CL.getResourceAsStream("EgressFetchHandlerExposeTest.json"), StandardCharsets.UTF_8);
+
+    assertEquals(expect.trim(), json);
+  }
+
+  private String jsonDumpWithExpose(boolean expose) throws Exception {
+    Map<String, String> labels = new HashMap<String, String>();
+    labels.put("foo", "bar");
```
hbs: Would be clearer if the values used for producer and owner were actual UUIDs, it would not lead people into thinking other values could be there

steven-gueguen: done

PR #660, opened 2020-02-10
Change JSON lib from Boon to Jackson

ftence:
# Intro

The initial motive for removing the dependency to Boon was because it relies on the Unsafe class and there is no plan to remove that. This makes it impossible to upgrade to Java 9+, if the need arise in the future.

During the development it also appeared that Boon is not the fastest JSON serialization/deserialization library anymore and it suffers from minor issues (more on that later).

To replace Boon, Jackson was chosen because it is a fast, well maintained, feature-rich and widely used JSON serialization/deserialization library.

This PR first isolates the dependency to the library to a single package, `io.warp10.json`, by providing a single class `io.warp10.json.JsonUtils` with methods to serialize/deserialize to/from JSON.

Next, the dependency has been changed to the last version (2.10.2) of `jackson-databind` to be able to easily update the library as the API slightly changed in 2.10.

These changes impacted the behavior and results as listed below:

# Serialization

## Type by type differences
- Bitset: serialized to `null`, no change.
- Boolean: no change.
- byte[]: serialized to base64 instead of `null`.
- Counter (AtomicLong): serialized like Long instead of its String representation.
- Double: no change.
- GTSEncoder: added missing lastActivity. Except that, no change.
- GeoShape: no change.
- GTS: no change.
- List: no change.
- Long: no change.
- Macro: no change.
- Map: null keys are serialized as `"null"` instead of `null`. This makes the JSON valid.
- Matrix (RealMatrix): serialized to `null`, no change.
- NamedWarpScriptFunction: no change.
- null: no change.
- PFont, PGraphics, PImage, PShape: serialized to `null`, no change.
- Set: serialized as a list, was previously `null`.
- String: Unicode escaping use capital letter (ie `\u00C3` instead of `\u00c3`). Except that, no change.
- Vector (ArrayRealVector): serialized to `null`, no change.
- VList (Vector): no change.

## ->JSON
`->JSON` now uses the exact same function as the Egress output to generate JSON and the call to `validate` has been removed.
This means:
- All the types handled above are also handled by `->JSON` instead of only List, Map, Number and Boolean.
- If there is a cycle in the structure, this is handled by Jackson with a `StackOverflowError`.

On top of that, as it was pretty easy to reach a OOM with ->JSON there is now a limit on the size of a generated JSON **only** for `->JSON`. The size limit can be changed in the conf with `warpscript.maxjson` and `warpscript.maxjson.hard` and in an authenticated WarpScript stack using `MAXJSON`.

## Custom serialization
Like SNAPSHOT, it is possible to register custom serializers for JSON with `io.warp10.json.JsonUtils#addEncoder`.

# Deserialization
Deserialization with Boon suffer from a number of issues which has been fixed in this PR:
- Boolean: object type is now Boolean instead of ValueContainer.
- Double: object type is now Double instead of NumberValue. Infinity and NaN are handled by Jackson contrary to Boon.
- List: no change.
- Long: object type is now Double instead of NumberValue.
- Map: no change.
- null: object is now really `null` instead of a `ValueContainer` instance.
- String: no change.
- String representing a ISO8601 date in a List: was converted to a Date with Boon, now correctly deserialized as a String.


ftence:
For the serialization, these choices are open for discussion, comments are welcome:
- byte[] serialized as base64
- Set serialized as list
- ArrayRealVector and RealMatrix not serialized

warp10/src/main/java/io/warp10/script/functions/TOJSON.java
```diff
@@ -43,120 +37,22 @@ public TOJSON(String name) {
...
     //
     // Only allow the serialization of simple lists and maps, otherwise JSON might
     // expose internals
     //
 
-    if (!validate(o)) {
-      throw new WarpScriptException(getName() + " can only serialize structures containing numbers, strings, booleans, lists and maps which do not reference the same list/map multiple times.");
+    try {
+      Long maxJsonSize = (Long)stack.getAttribute(WarpScriptStack.ATTRIBUTE_JSON_MAXSIZE);
+      String json = JsonUtils.objectToJson(o, false, maxJsonSize);
```
hbs: This call will fail with an NPE if the stack attribute retrieved above is not set. It is set in the case of MemoryWarpScriptStack instances but potentially not in other implementations of WarpScriptStack.

Same in MAXJSON

warp10/src/main/java/io/warp10/json/JsonUtils.java
```diff
@@ -0,0 +1,211 @@
...
+        StringBuilder sb = new StringBuilder();
+        boolean encoded = false;
+        for (JsonEncoder encoder: encoders) {
+          encoded = encoder.addElement(sb, value);
+          if (encoded) {
+            break;
+          }
+        }
+        if (encoded) {
+          gen.writeString(sb.toString());
```
hbs: Won't that surround sb's content with double quotes and escape it? This might not be what is expected.

warp10/src/main/java/io/warp10/json/MetadataSerializer.java
```diff
@@ -0,0 +1,42 @@
...
+package io.warp10.json;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import io.warp10.continuum.store.thrift.data.Metadata;
+
+import java.io.IOException;
+
+public class MetadataSerializer extends JsonSerializer<Metadata> {
```
hbs: Why isn't this called by the serializer for GTS and GTS Encoders?

warp10/src/main/java/io/warp10/script/functions/TOJSON.java
ftence: I think this is the case for all other limits.
It this is not set, what should be the value? The default one or the max one?

warp10/src/main/java/io/warp10/json/GTSEncoderSerializer.java
```diff
@@ -0,0 +1,86 @@
...
+    gen.writeNumberField("la", metadata.getLastActivity());
+    gen.writeFieldName("v");
+    gen.writeStartArray();
+
+    GTSDecoder decoder = encoder.getUnsafeDecoder(false);
+    while (decoder.next()) {
+      long ts = decoder.getTimestamp();
+      long location = decoder.getLocation();
+      long elevation = decoder.getElevation();
+      // We do not call getBinaryValue because JSON cannot represent byte arrays
```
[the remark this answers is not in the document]
ftence: You're right.
If you're fine with the b64 serialization of byte[] we could call `getBinaryValue` here and b64 encode if the returned value it `byte[]`.
However, in the case of a GTS, we're stuck with an ISO-8859-1 string, correct me if I'm wrong.

warp10/src/main/java/io/warp10/json/GTSEncoderSerializer.java
```diff
@@ -0,0 +1,86 @@
...
+  public void serialize(GTSEncoder encoder, JsonGenerator gen, SerializerProvider provider) throws IOException {
+    Metadata metadata = encoder.getMetadata();
+    String name = metadata.getName();
+    if (null == name) {
+      name = "";
+    }
+
+    gen.writeStartObject();
+    gen.writeStringField("c", name);
+    gen.writeObjectField("l", metadata.getLabels());
```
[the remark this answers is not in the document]
ftence: Problem is that we don't  have access to the token here, so we can't know if internal labels should be exposed or not.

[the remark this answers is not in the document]
hbs: Actually it should be fine since the labels will have been removed just after the FETCH if they needed to be.

warp10/src/main/java/io/warp10/json/GTSEncoderSerializer.java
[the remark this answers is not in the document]
hbs: I am not too fond of emitting b64 content for byte arrays, why not simply output an ISO-8859-1 string?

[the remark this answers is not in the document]
ftence: This would unify the GTS, Encoder and byte[] representation for binary data which is nice.
There is a slight downside as it will produce a lot of special characters which will be escaped in JSON, making that not very compact. I think it's fine because JSON shouldn't be used for large transfer of binary data.

warp10/src/main/java/io/warp10/script/functions/TOJSON.java
ftence: Better fix that in another PR as all other limit attributes suffer from the same problem.

warp10/src/main/java/io/warp10/json/GTSEncoderSerializer.java
```diff
@@ -0,0 +1,84 @@
...
+    super(GTSEncoder.class);
+  }
+
+  @Override
+  public void serialize(GTSEncoder encoder, JsonGenerator gen, SerializerProvider provider) throws IOException {
+    Metadata metadata = encoder.getMetadata();
+
+    gen.writeStartObject();
+    MetadataSerializer.serializeMetadataFields(metadata, gen);
+    gen.writeFieldName("v");
```
hbs: Use the constant

warp10/src/main/java/io/warp10/continuum/egress/EgressFetchHandler.java
```diff
@@ -1341,8 +1338,7 @@ static void jsonDump(PrintWriter pw, Iterator<GTSDecoder> iter, long now, long c
           
           sb.append("{\"c\":");
```
hbs: Use the constants defined in the MetadataSerializer?

warp10/src/main/java/io/warp10/continuum/egress/EgressFetchHandler.java
```diff
@@ -1365,38 +1367,40 @@ static void jsonDump(PrintWriter pw, Iterator<GTSDecoder> iter, long now, long c
...
+            sb.append(JsonUtils.objectToJson(entry.getKey()));
             sb.append(":");
-            //sb.append(gson.toJson(entry.getValue()));
-            sb.append(serializer.serialize(entry.getValue()));
+            sb.append(JsonUtils.objectToJson(entry.getValue()));
             first = false;
           }
           
           sb.append("}");
           sb.append(",\"i\":\"");
```
hbs: FIELD_ID

warp10/src/main/java/io/warp10/continuum/egress/EgressFetchHandler.java
```diff
@@ -1365,38 +1367,42 @@ static void jsonDump(PrintWriter pw, Iterator<GTSDecoder> iter, long now, long c
...
-            //sb.append(gson.toJson(entry.getValue()));
-            sb.append(serializer.serialize(entry.getValue()));
+            sb.append(JsonUtils.objectToJson(entry.getValue()));
             first = false;
           }
           
           sb.append("}");
-          sb.append(",\"i\":\"");
+          sb.append(",\"");
+          sb.append(MetadataSerializer.FIELD_LABELSID);
```
hbs: It is a an id, not directly the labelsId, so better call this field FIELD_ID as it could very well not stay related to the labels id only

warp10/src/main/java/io/warp10/json/JsonUtils.java
```diff
@@ -0,0 +1,214 @@
...
+    looseMapper.getSerializerProvider().setNullKeySerializer(nullKeySerializer);
+    looseMapper.registerModule(module);
+  }
+
+  //
+  // Method to deserialize JSON to Objects.
+  //
+
+  public static Object jsonToObject(String json) throws JsonProcessingException {
+    return strictMapper.readValue(json, Object.class);
```
hbs: Are integer numbers correctly converted to LONGs by this?

warp10/src/main/java/io/warp10/continuum/egress/EgressFetchHandler.java
ftence: True. As this is not related to Metadata, I have to put this constant in EgressFetchHandler.

warp10/src/main/java/io/warp10/json/JsonUtils.java
ftence: Yes, and floating-point numbers to Double.

PR #661, opened 2020-02-10
Add types and clean TYPEOF

ftence:
Also merged reducer, mapper and bucketizer under aggregator because they cannot be differentiated.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/TYPEOF.java
+++ b/warp10/src/main/java/io/warp10/script/functions/TYPEOF.java
@@ -76,6 +76,11 @@ public class TYPEOF extends NamedWarpScriptFunction implements WarpScriptStackFu
   public static final String TYPE_PSHAPE = "PSHAPE";
   public static final String TYPE_COUNTER = "COUNTER";
 
+  /**
+   * Interface to be used by extensions and plugins to define a new type.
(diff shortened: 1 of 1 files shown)


randomboolean:
The function typeof(Class c) is needed by some other classes.

ftence:
Which ones? Couldn't find any.

randomboolean:
Mostly in extensions.

ftence:
Can you be more precise?

randomboolean:
Here for example: https://github.com/senx/warp10-ext-arrow/search?q=typeof&unscoped_q=typeof

ftence:
Thanks for the heads-up @randomboolean, this is fixed.

warp10/src/main/java/io/warp10/script/functions/TYPEOF.java
```diff
@@ -97,57 +108,72 @@ public static String typeof(Object o) {
...
     } else {
-      return (c.getSimpleName().toUpperCase());
+      return defaultType(c);
+    }
+  }
+
+  private static String defaultType(Class c) {
+    String canonicalName = c.getCanonicalName();
+    if (null == canonicalName) {
+      return "X-(Local/Anonymous Class)";
```
hbs: It would be better to use neither parens nor spaces.

X-Local-Anonymous-Class?

PR #665, opened 2020-02-14
Duration.bucketize

randomboolean:


warp10/src/main/java/io/warp10/script/functions/DURATION_BUCKETIZE.java
```diff
@@ -0,0 +1,368 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Bucketizes some GTS instances using a bucketduration rather than a bucketspan.
+ */
+public class DURATION_BUCKETIZE extends NamedWarpScriptFunction implements WarpScriptStackFunction {
```
hbs: No underscore in class names

warp10/src/main/java/io/warp10/script/functions/DURATION_BUCKETIZE.java
```diff
@@ -0,0 +1,368 @@
...
+
+    DateTimeZone dtz = DateTimeZone.UTC;
+    if (params.get(params.size() - 1) instanceof String) {
+      String tz = (String) params.remove(params.size() - 1);
+      dtz = DateTimeZone.forID(tz);
+    }
+
+    for (int i = 0; i < params.size() - 4; i++) {
+      if (!(params.get(i) instanceof GeoTimeSerie) && !(params.get(i) instanceof List)) {
+        throw new WarpScriptException(getName() + " expects a list of Geo Time Series as first parameter.");
```
hbs: or a Geo Time Serie

warp10/src/main/java/io/warp10/script/functions/DURATION_BUCKETIZE.java
```diff
@@ -0,0 +1,368 @@
...
+      throw new WarpScriptException("Bucket count (" + bucketcount + ") would exceed maximum value of " + maxbuckets);
+    }
+
+    //
+    // Check that input gts are not already duration-bucketized
+    //
+
+    for (GeoTimeSerie gts : series) {
+      if (gts.getMetadata().getAttributes().get(DURATION_ATTRIBUTE_KEY) != null || gts.getMetadata().getAttributes().get(OFFSET_ATTRIBUTE_KEY) != null || gts.getMetadata().getAttributes().get(TIMEZONE_ATTRIBUTE_KEY) != null) {
+        throw new WarpScriptException(getName() + " expects GTS for which the attributes " + DURATION_ATTRIBUTE_KEY + ", " + OFFSET_ATTRIBUTE_KEY + " and " + TIMEZONE_ATTRIBUTE_KEY + " are not be set. If an input GTS is supposed to be already duration-bucketized, duration-unbucketize them first before applying a new duration-bucketization.");
```
hbs: "are not set"
"duration-unbucketize IT first"

warp10/src/main/java/io/warp10/script/functions/DURATION_BUCKETIZE.java
```diff
@@ -0,0 +1,368 @@
...
+      }
+    }
+
+    //
+    // Check nullity of bucketizer
+    //
+
+    if (null == bucketizer) {
+      throw new WarpScriptException(getName() + " expects a non null bucketizer.");
```
hbs: On line 90 you said you could use NULL

warp10/src/main/java/io/warp10/script/functions/DURATION_BUCKETIZE.java
```diff
@@ -0,0 +1,368 @@
...
+      }
+      lastbucket_index--;
+      bucketoffset = -(ADDDURATION.addPeriod(boundary, bucketperiod, dtz, -1) + 1);
+    }
+
+    //
+    // Duration-Bucketize
+    //
+
+    List<GeoTimeSerie> bucketized = new ArrayList<GeoTimeSerie>();
```
hbs: Initialize it with the known size of series

warp10/src/main/java/io/warp10/script/functions/DURATION_BUCKETIZE.java
```diff
@@ -0,0 +1,368 @@
...
+    for (GeoTimeSerie gts : series) {
+
+      GeoTimeSerie b = durationBucketize(gts, bucketperiod, dtz, bucketcount, lastbucket, lastbucket_index, bucketizer, maxbuckets, bucketizer instanceof Macro ? stack : null);
+      b.getMetadata().getAttributes().put(DURATION_ATTRIBUTE_KEY, bucketduration);
```
hbs: Use putToAttributes as attributes could be null

warp10/src/main/java/io/warp10/script/functions/DURATION_UNBUCKETIZE.java
```diff
@@ -0,0 +1,81 @@
...
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptStack;
+import org.joda.time.DateTimeZone;
+
+import java.util.Map;
+
+/**
+ * Bucketizes some GTS instances using a bucketduration rather than a bucketspan.
+ */
+public class DURATION_UNBUCKETIZE extends GTSStackFunction {
```
hbs: Remove underscore

warp10/src/main/java/io/warp10/script/WarpScriptLib.java
```diff
@@ -806,6 +806,7 @@
   public static final String BUCKETSPAN = "BUCKETSPAN";
   public static final String BUCKETCOUNT = "BUCKETCOUNT";
   public static final String UNBUCKETIZE = "UNBUCKETIZE";
+  public static final String DURATION_UNBUCKETIZE = io.warp10.script.functions.DURATION_UNBUCKETIZE.getDefaultName();
```
hbs: Better define the name here and reference it in the function

warp10/src/main/java/io/warp10/script/WarpScriptLib.java
```diff
@@ -868,6 +869,7 @@
   public static final String NBOUNDS = "NBOUNDS";
   public static final String LBOUNDS = "LBOUNDS";
   public static final String BUCKETIZE = "BUCKETIZE";
+  public static final String DURATION_BUCKETIZE = io.warp10.script.functions.DURATION_BUCKETIZE.getDefaultName();
```
hbs: ditto

warp10/src/main/java/io/warp10/script/functions/ADDDURATION.java
```diff
@@ -78,16 +78,75 @@ public WarpScriptStack apply(WarpScriptStack stack) throws WarpScriptException {
       throw new WarpScriptException(getName() + " operates on a tselements list, timestamp, or timestamp and timezone.");
     }
 
+    //
+    // Handle time zone
+    //
+
+    if (null == tz) {
+      tz = "UTC";
```
hbs: Use a constant known to be a valid timezone for the parsing library

randomboolean:
> There's a bug somewhere because
> 
> ```
> [
>   NEWGTS
>   42 NaN NaN NaN '42' ADDVALUE
>   bucketizer.last
>   24 h
>   'PT1M'
>   3
> ]
> DURATION.BUCKETIZE
> ```
> 
> returns a GTS with 1 bucket.

That is the wanted behaviour. See https://warp10.io/doc/DURATION.BUCKETIZE

warp10/src/main/java/io/warp10/script/functions/ADDDURATION.java
```diff
@@ -78,16 +78,75 @@ public WarpScriptStack apply(WarpScriptStack stack) throws WarpScriptException {
...
+    DateTimeZone dtz = DateTimeZone.UTC;
+    if (null != tz) {
+      dtz = DateTimeZone.forID(tz);
+    }
+
     //
     // Handle duration
     //
 
+    ReadWritablePeriodWithSubSecondOffset period = durationToPeriod(duration);
```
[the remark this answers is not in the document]
randomboolean: I think there was an previous case where @hbs disagreed. See https://github.com/senx/warp10-platform/pull/468#discussion_r295652492

warp10/src/main/java/io/warp10/script/functions/ADDDURATION.java
```diff
@@ -100,26 +159,39 @@ public WarpScriptStack apply(WarpScriptStack stack) throws WarpScriptException {
...
-    boolean tselements = false;
-    if (stack.peek() instanceof List) {
-      FROMTSELEMENTS.apply(stack);
-      tselements = true;
-    }
-
-    long instant = ((Number) stack.pop()).longValue();
     DateTime dt = new DateTime(instant / Constants.TIME_UNITS_PER_MS, dtz);
 
     //
```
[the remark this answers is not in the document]
randomboolean: In english, non-negative is used to refer to a number that is either positive or zero.

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
```diff
@@ -0,0 +1,369 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Bucketizes some GTS instances using a bucketduration rather than a bucketspan.
```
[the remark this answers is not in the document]
randomboolean: Shouldn't the explanation be left to the documentation ?

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
```diff
@@ -0,0 +1,369 @@
...
+      }
+    }
+
+    //
+    // Check nullity of bucketizer
+    //
+
+    if (null == bucketizer) {
+      throw new WarpScriptException(getName() + " expects a non null bucketizer.");
+    }
```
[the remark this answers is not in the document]
randomboolean: durationBucketize checks if it is of the correct type, and the error message will only inform that it is not. Here the error message also specifies that it was null, and save some computations.

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
```diff
@@ -0,0 +1,369 @@
...
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Bucketizes some GTS instances using a bucketduration rather than a bucketspan.
+ */
+public class DURATIONBUCKETIZE extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+
+  private static final String DEFAULT_NAME = WarpScriptLib.DURATION_BUCKETIZE;
```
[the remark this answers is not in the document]
randomboolean: Initially that was the other way around but made a modification after a PR comment that kept it this way.

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
```diff
@@ -0,0 +1,369 @@
...
+
+    if (null == bucketizer) {
+      throw new WarpScriptException(getName() + " expects a non null bucketizer.");
+    }
+
+    //
+    // Convert duration to joda.time.Period
+    //
+
+    if (!DURATION_RE.reset(bucketduration).matches()) {
```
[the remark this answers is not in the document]
randomboolean: Yes, since here we do not support negative values but we do in durationToPeriod.

warp10/src/main/java/io/warp10/script/functions/DURATIONUNBUCKETIZE.java
```diff
@@ -0,0 +1,82 @@
...
+import io.warp10.script.GTSStackFunction;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptLib;
+import io.warp10.script.WarpScriptStack;
+import org.joda.time.DateTimeZone;
+
+import java.util.Map;
+
+/**
+ * Bucketizes some GTS instances using a bucketduration rather than a bucketspan.
```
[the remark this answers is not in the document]
randomboolean: ditto

warp10/src/main/java/io/warp10/script/functions/DURATIONUNBUCKETIZE.java
```diff
@@ -0,0 +1,82 @@
...
+      throw new WarpScriptException(getName() + " expects input GTS to be bucketized.");
+    }
+
+    if (!DURATIONBUCKETIZE.isDurationBucketized(gts)) {
+      throw new WarpScriptException(getName() + " expects input GTS to be duration-bucketized. This information is stored in attributes.");
+    }
+
+    ADDDURATION.ReadWritablePeriodWithSubSecondOffset bucketperiod = ADDDURATION.durationToPeriod(gts.getMetadata().getAttributes().get(DURATIONBUCKETIZE.DURATION_ATTRIBUTE_KEY));
+    long bucketoffset = Long.parseLong(gts.getMetadata().getAttributes().get(DURATIONBUCKETIZE.OFFSET_ATTRIBUTE_KEY));
+    DateTimeZone dtz = DateTimeZone.forID(gts.getMetadata().getAttributes().get(DURATIONBUCKETIZE.TIMEZONE_ATTRIBUTE_KEY));
```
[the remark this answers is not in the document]
randomboolean: This is done on the previous line by DURATIONBUCKETIZE.isDurationBucketized(gts).

warp10/src/main/java/io/warp10/script/functions/ADDDURATION.java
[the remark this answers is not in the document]
ftence: Then, use the other solution, that is wrap the exception in another one with the name.

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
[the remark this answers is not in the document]
ftence: Keep it short if possible without too much detail. In this state, the current comment does not help.

ftence:
I'm not sure this is the wanted behavior.
True there are 3 buckets, but all should be empty, look at
```
[
  NEWGTS
  42 NaN NaN NaN '42' ADDVALUE
  bucketizer.last
  24 h
  1 m
  3
]
BUCKETIZE
```
In the case of `DURATION.BUCKETIZE` the GTS has 1 value while `BUCKETIZE` returns 0 value.

randomboolean:
Yes you are right, I'm looking into it.

warp10/src/main/java/io/warp10/script/functions/ADDDURATION.java
```diff
@@ -78,16 +78,75 @@ public WarpScriptStack apply(WarpScriptStack stack) throws WarpScriptException {
...
+
+  /**
+   * Convert an ISO8601 duration to a Period.
+   * @param duration
+   * @return
+   * @throws WarpScriptException
+   */
+  public static ReadWritablePeriodWithSubSecondOffset durationToPeriod(String duration) throws WarpScriptException {
     // Separate seconds from  digits below second precision
     String[] tokens = UnsafeString.split(duration, '.');
```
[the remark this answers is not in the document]
hbs: UnsafeString is a helper class which today uses Unsafe but will no longer do so in the JDK11+ branch, so using it is safe (!) as its split method will be refactored to drop the use of Unsafe in the future.

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
```diff
@@ -0,0 +1,369 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Bucketizes some GTS instances using a bucketduration rather than a bucketspan.
+ */
+public class DURATIONBUCKETIZE extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+
+  private static final String DEFAULT_NAME = WarpScriptLib.DURATION_BUCKETIZE;
+  private static final Matcher DURATION_RE = Pattern.compile("^P(?!$)(\\d+Y)?(\\d+M)?(\\d+W)?(\\d+D)?(T(?=\\d)(\\d+H)?(\\d+M)?((\\d+|\\d.(\\d)+)S)?)?$").matcher("");
```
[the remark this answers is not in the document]
hbs: Actually checking that it starts with a P and can be parsed by a time library is probably safer.

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
```diff
@@ -0,0 +1,370 @@
...
+    if (null == bucketizer) {
+      throw new WarpScriptException(getName() + " expects a non null bucketizer.");
+    }
+
+    //
+    // Convert duration to joda.time.Period
+    //
+
+    if (!DURATION_RE.reset(bucketduration).matches()) {
+      throw new WarpScriptException(getName() + " expects the bucketduration parameter to be a valid ISO8601 duration with positive coefficients.");
```
hbs: Why positive coefficients? Wouldn't P1M-1d be legitimate to support?

randomboolean: Yes it could be. We could just check that they the overall duration is positive.

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
```diff
@@ -0,0 +1,369 @@
...
+    } else {
+      long boundary = ADDDURATION.addPeriod(lastbucket, bucketperiod, dtz);
+
+      lastbucket_index = -1;
+      while (boundary < 0) {
+        boundary = ADDDURATION.addPeriod(boundary, bucketperiod, dtz);
+      }
+      lastbucket_index--;
+      bucketoffset = -(ADDDURATION.addPeriod(boundary, bucketperiod, dtz, -1) + 1);
+    }
```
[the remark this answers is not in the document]
randomboolean: fixed by 728b40792447f0569e299dc4d2288e4fb1866557

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
randomboolean: Actually this one is tricky, the duration can be positive from some date and negative from others. For example, allowing P1M-31DT1H would make the computation go back and forth.

hbs: You can throw an exception if you encounter an inconsistency while generating the indices

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
```diff
@@ -0,0 +1,420 @@
...
+      bucketperiod = ADDDURATION.durationToPeriod(bucketduration);
+    } catch (WarpScriptException wse) {
+      throw new WarpScriptException(getName() + " encountered an exception: " + wse.getMessage());
+    }
+
+    //
+    // Check that the bucketduration is positive
+    //
+
+    long averageSpan = bucketperiod.getPeriod().toPeriod().toDurationFrom(new Instant()).getMillis() * Constants.TIME_UNITS_PER_MS + bucketperiod.getOffset();
```
ftence: Are you sure about creating a new Instant each time?
Could it be better to chose a fixed instant away from known special dates for instance to avoid leap-second, February and daylight saving changes?

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
```diff
@@ -0,0 +1,420 @@
...
+      flag = addNonNegativePeriod(flag, bucketperiod, dtz, -1);
+      lastbucketIndex--;
+    }
+
+    while (flag <= lastbucket) {
+      flag = addNonNegativePeriod(flag, bucketperiod, dtz, 1);
+      lastbucketIndex++;
+    }
+
+    bucketoffset = flag - (lastbucket + 1);
```
ftence: That's working very well to speed-up things for small bucket durations, congrats.
However it seems it is extremely slow with negative last bucket:
```
-1582560100948303 'now' STORE

[
  NEWGTS
  $now NaN NaN NaN T ADDVALUE
  bucketizer.last
  $now 1.2 s +
  'PT1.2S'
  3
]
DURATION.BUCKETIZE
```

randomboolean: Bugfix ac3ddee

warp10/src/main/java/io/warp10/script/functions/DURATIONBUCKETIZE.java
randomboolean: Leap seconds are not supported by Unix timestamp. We don't use any particular timezone here so it is UTC which does not has daylight saving time. There can be a point with February but 1970 is not a leap year so that would only applies for durations > 2 years and that's not very impactful on our index hint.

warp10/src/main/java/io/warp10/script/functions/DURATIONUNBUCKETIZE.java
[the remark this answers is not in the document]
randomboolean: If they have been modified by the user then it's up to him to know what this function is doing. (Similarly, for a bucketized GTS you can also only set its bucketize parameters for then applying functions that take into account that bucketize structure).

PR #666, opened 2020-02-14
Warp 10 Accelerator

hbs:


warp10/src/main/java/io/warp10/standalone/StandaloneAcceleratedStoreClient.java
```diff
@@ -0,0 +1,298 @@
...
+    
+    // Use the persistent store unless ACCEL.NOPERSIST was called 
+    if (((now > cacheend || then < cachestart) || preBoundary > 0 || postBoundary > 0 || nocache.get()) && !nopersist.get()) {
+      accelerated.set(Boolean.FALSE);
+      return this.persistent.fetch(token, metadatas, now, then, count, skip, sample, writeTimestamp, preBoundary, postBoundary);
+    }
+    
+    // Last resort, use the cache
+    accelerated.set(Boolean.TRUE);
+    return this.cache.fetch(token, metadatas, now, then, count, skip, sample, writeTimestamp, preBoundary, postBoundary);
```
ftence: Should the result be deduped to have consistent result between cache and persist?
Or modify the cache so it does not store duplicate ticks.

Also the result will not be sorted using the cache but will be using persist. I don't think this is as problematic as the duplicate ticks is.

warp10/src/main/java/io/warp10/standalone/StandaloneAcceleratedStoreClient.java
```diff
@@ -0,0 +1,298 @@
...
+        while(decoders.hasNext()) {
+          GTSDecoder decoder = decoders.next();
+          decoder.next();
+          GTSEncoder encoder = decoder.getEncoder(true);
+          this.cache.store(encoder);
+          datapoints.addAndGet(decoder.getCount());
+        }
+      }
+
+      exec.shutdown();
```
ftence: You have to wait for the tasks to complete after that. If you want to log the waiting, you can do:
```
while (!exec.awaitTermination(30, TimeUnit.SECONDS)) {
  LOG.info("Awaiting completion of accelerator preloading.");
}
```

warp10/src/main/java/io/warp10/standalone/StandaloneAcceleratedStoreClient.java
```diff
@@ -0,0 +1,298 @@
...
+      final AtomicReference<Throwable> error = new AtomicReference<Throwable>();
+      
+      while(iter.hasNext()) {
+        batch.add(iter.next());
+        
+        if (null != error.get()) {
+          throw new RuntimeException("Error populating the accelerator", error.get());
+        }
+        
+        if (BATCH_SIZE == batch.size()) {
```
ftence: You could  add `|| !iter.hasNext` to avoid adding the `if (!batch.isEmpty())` after.

warp10/src/main/java/io/warp10/standalone/InMemoryChunkSet.java
```diff
@@ -839,4 +854,41 @@ long optimize(CapacityExtractorOutputStream out, long now, AtomicLong allocation
     
     return reclaimed;
   }
+  
+  public long delete(long start, long end) {
+    long count = 0L;
+    
+    for (int i = 0; i < chunks.length; i++) {
+      if (!this.ephemeral && (chunkends[i] < start || chunkends[i] > end + chunklen)) {
```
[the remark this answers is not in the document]
hbs: chunkends[] keeps track of the last timestamp in a chunk. Each chunk is chunklen time units wide, so we are checking here if the ith chunk ends more than chunklen timeunits after end, making the whole chunk past 'end'

warp10/src/main/java/io/warp10/standalone/StandaloneAcceleratedStoreClient.java
hbs: The semantics are those of the in-memory store, duplication is not handled at the store level, for the rare cases where you want to avoid duplicates datapoints, simply call DEDUP after FETCH if it was against the in-memory store.

warp10/src/main/java/io/warp10/standalone/StandaloneAcceleratedStoreClient.java
```diff
@@ -0,0 +1,298 @@
...
+      return Boolean.FALSE;
+    }
+  };
+
+  private static final ThreadLocal<Boolean> nopersist = new ThreadLocal<Boolean>() {
+    @Override
+    protected Boolean initialValue() {
+      return Boolean.FALSE;
+    }
+  };
```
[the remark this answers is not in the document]
hbs: Indeed, that is why both FETCH and EgressFetchHandler call either cache/nocache and persist/nopersist prior to calling fetch. The only case which should be a concern is when the StoreClient is exposed via 'egress.clients.expose' set to true, as the use of the StoreClient should be preceded by the same logic of calling cache/nocache/persist/nopersist.

Adding a warning in the configuration.

waxzce:
gr8 idea

warp10/src/main/java/io/warp10/script/functions/ACCELREPORT.java
```diff
@@ -0,0 +1,41 @@
...
+package io.warp10.script.functions;
+
+import io.warp10.script.NamedWarpScriptFunction;
+import io.warp10.script.WarpScriptStackFunction;
+import io.warp10.standalone.StandaloneAcceleratedStoreClient;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptStack;
+
+public class ACCELREPORT extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+
```
pi-r-p: Is it a debug function or not ? If so, it should report more than the last fetch status. Maybe a map with :
- accelerator activation status
- accelerator preload status
- accelerator max time depth
- last fetch was accelerated
- number of gts in the accelerator (help for debug remote customers too) (as directory is not affected, FIND won't make any difference when called after ACCEL.NOPERSIST)

warp10/src/main/java/io/warp10/standalone/StandaloneAcceleratedStoreClient.java
```diff
@@ -0,0 +1,304 @@
...
+import io.warp10.continuum.store.thrift.data.Metadata;
+import io.warp10.quasar.token.thrift.data.ReadToken;
+import io.warp10.quasar.token.thrift.data.WriteToken;
+
+public class StandaloneAcceleratedStoreClient implements StoreClient {
+  
+  private static final Logger LOG = LoggerFactory.getLogger(StandaloneAcceleratedStoreClient.class);
+  
+  private final StoreClient persistent;
+  private final StandaloneChunkedMemoryStore cache;
```
pi-r-p: So it must be a chunked memory store. Warp 10 must fail or warn if not properly configured.
Maybe the most clean will be to have `accelerator.chunk.count` and `accelerator.chunk.length` configurations.

hbs:
PTAL

warp10/src/main/java/io/warp10/standalone/StandaloneAcceleratedStoreClient.java
```diff
@@ -0,0 +1,377 @@
...
+    
+    // Use the persistent store unless ACCEL.NOPERSIST was called 
+    if (((now > cacheend || then < cachestart) || preBoundary > 0 || postBoundary > 0 || nocache.get()) && !nopersist.get()) {
+      accelerated.set(Boolean.FALSE);
+      return this.persistent.fetch(token, metadatas, now, then, count, skip, sample, writeTimestamp, preBoundary, postBoundary);
+    }
+    
+    // Last resort, use the cache
+    accelerated.set(Boolean.TRUE);
+    return this.cache.fetch(token, metadatas, now, then, count, skip, sample, writeTimestamp, preBoundary, postBoundary);
```
ftence: Shouldn't an error be thrown if `nocache.get()`?

warp10/src/main/java/io/warp10/standalone/StandaloneAcceleratedStoreClient.java
```diff
@@ -72,11 +89,29 @@ protected Boolean initialValue() {
...
+    }
+            
+    //
+    // Force accelerator parameters to be replicated on inmemory ones and clear other in memory params
+    //
+    
+    if (null == WarpConfig.getProperty(Configuration.ACCELERATOR_CHUNK_COUNT)
+        || null == WarpConfig.getProperty(Configuration.ACCELERATOR_CHUNK_LENGTH)) {
+      throw new RuntimeException("Missing configuration key '" + Configuration.ACCELERATOR_CHUNK_COUNT + "' or '" + Configuration.ACCELERATOR_CHUNK_LENGTH + "'");
+    }
```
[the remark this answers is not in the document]
hbs: The accelerator will load its in-memory cache with data from disk upon startup, having default values similar to those of in.memory.xxx would lead to massive fetch ops, so better not have default values and force the user to choose wisely.

warp10/src/main/java/io/warp10/standalone/InMemoryChunkSet.java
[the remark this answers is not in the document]
hbs: indeed

PR #672, opened 2020-02-26
Initial commit of GEOSHIFT

hbs:


pi-r-p:
```
@senx/geo/southHemisphere
50 90 8 GEOSHIFT
```

stops with `Exception at 'GEOSHIFT' in section '[TOP]' (GEOSHIFT cannot use resolution finer than 8)`

It is annoying, because we can't know the precision of the input shape.
There is no obvious workaround. Adding a tiny cell to the initial shape slows down everything.

hbs:
Specifying a resolution is for cases when you know what you are doing, in the general case simply specify 0 to let GEOSHIFT determine the finest possible resolution to use.

pi-r-p:
Shifting geoshape is OK. Troubles come when you impose a meridian on a shape that covers the pole: there is a weird shift:
```
@senx/geo/northHemisphere
0 50 0 0 GEOSHIFT

true ->GEOJSON JSON-> 'j' STORE
{ 'data' [ $j ] }
```

image


```
180 0 6000e3 4 @senx/geo/earthCircle 
DUP true ->GEOJSON JSON-> 'original' STORE

0 40 0 0 GEOSHIFT
true ->GEOJSON JSON-> 'shifted' STORE

{ 'data' [ 
  $original
  $shifted
] }

```
image
becomes 
image

hbs:
This is linked to rounding at the coarse resolution of your example. The cells are shifted correctly.

PR #675, opened 2020-02-27
Added support for GTSEncoder instances in META

hbs:


warp10/src/main/java/io/warp10/script/functions/META.java
```diff
@@ -66,27 +68,31 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
     
     if (!(otoken instanceof String)) {
-      throw new WarpScriptException(getName() + " expects a token on top of the stack.");
+      throw new WarpScriptException(getName() + " expects a token.");
     }
     
     String token = (String) otoken;
 
-    List<GeoTimeSerie> series = new ArrayList<GeoTimeSerie>();
+    List<Metadata> series = new ArrayList<Metadata>();
```
ftence: Variable name should be changed to reflect the type.

warp10/src/main/java/io/warp10/script/functions/META.java
```diff
@@ -158,11 +164,11 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       
       StringBuilder sb = new StringBuilder();
       
-      for (GeoTimeSerie gts: series) {
+      for (Metadata gts: series) {
```
ftence: Variable name should be changed to reflect the type.

PR #676, opened 2020-02-27
fix #635

### LINKED ISSUE #635: arithmetics on empty GTS
This code fails:
```
NEWGTS 'no value' RENAME
0.1 *
```

It should return a clone of the empty GTS, but not fail...

pi-r-p:
Add more tolerance to empty GTS in all operators.

```
NEWGTS 'foo' RENAME 0.1 *
```
now valid.

```
NEWGTS 'dodo' RENAME 
0 NaN NaN NaN 8 ADDVALUE
NEWGTS 'fuu' RENAME
0 NaN NaN NaN 1 ADDVALUE
[ SWAP 5 mapper.gt 0 0 0 ] MAP 0 GET // non typed gts
*
```
now valid

```
NEWGTS NEWGTS *
```
now valid.

Same for + / - operators.
Also fix a bug in comparison operators, when comparing an empty gts to a number.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/binary/DIV.java
+++ b/warp10/src/main/java/io/warp10/script/binary/DIV.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018  SenX S.A.S.
+//   Copyright 2020  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
--- a/warp10/src/main/java/io/warp10/script/binary/DIV.java
+++ b/warp10/src/main/java/io/warp10/script/binary/DIV.java
@@ -58,7 +58,6 @@ public class DIV extends NamedWarpScriptFunction implements WarpScriptStackFunct
       // Returns immediately a new gts if both inputs are empty
       if (0 == GTSHelper.nvalues(gts1) || 0 == GTSHelper.nvalues(gts2)) {
         GeoTimeSerie result = new GeoTimeSerie();
-        result.setType(TYPE.DOUBLE);
         stack.push(result);
         return stack;
       }
@@ -162,7 +161,6 @@ public class DIV extends NamedWarpScriptFunction implements WarpScriptStackFunct
 
       // Returns immediately a new clone if gts is empty.
       if (0 == n) {
-        result.setType(TYPE.DOUBLE);
         stack.push(result);
         return stack;
       }
--- a/warp10/src/main/java/io/warp10/script/binary/DIV.java
+++ b/warp10/src/main/java/io/warp10/script/binary/DIV.java
@@ -55,6 +55,14 @@ public class DIV extends NamedWarpScriptFunction implements WarpScriptStackFunct
       GeoTimeSerie gts1 = (GeoTimeSerie) op1;
       GeoTimeSerie gts2 = (GeoTimeSerie) op2;
 
+      // Returns immediately a new gts if both inputs are empty
+      if (0 == GTSHelper.nvalues(gts1) || 0 == GTSHelper.nvalues(gts2)) {
+        GeoTimeSerie result = new GeoTimeSerie();
+        result.setType(TYPE.DOUBLE);
+        stack.push(result);
+        return stack;
+      }
+
       if (!(gts1.getType() == TYPE.DOUBLE || gts1.getType() == TYPE.LONG) || !(gts2.getType() == TYPE.DOUBLE || gts2.getType() == TYPE.LONG)) {
         throw new WarpScriptException(typeCheckErrorMsg);
       }
@@ -151,6 +159,13 @@ public class DIV extends NamedWarpScriptFunction implements WarpScriptStackFunct
       
       GeoTimeSerie result = op1gts ? ((GeoTimeSerie) op1).cloneEmpty(n) : ((GeoTimeSerie) op2).cloneEmpty();
       GeoTimeSerie gts = op1gts ? (GeoTimeSerie) op1 : (GeoTimeSerie) op2;
+
+      // Returns immediately a new clone if gts is empty.
+      if (0 == n) {
+        result.setType(TYPE.DOUBLE);
+        stack.push(result);
+        return stack;
+      }
 
       if (!(gts.getType() == TYPE.LONG || gts.getType() == TYPE.DOUBLE)) {
         throw new WarpScriptException(typeCheckErrorMsg);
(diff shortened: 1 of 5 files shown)


warp10/src/main/java/io/warp10/script/binary/DIV.java
```diff
@@ -55,6 +55,14 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       GeoTimeSerie gts1 = (GeoTimeSerie) op1;
       GeoTimeSerie gts2 = (GeoTimeSerie) op2;
 
+      // Returns immediately a new gts if both inputs are empty
+      if (0 == GTSHelper.nvalues(gts1) || 0 == GTSHelper.nvalues(gts2)) {
+        GeoTimeSerie result = new GeoTimeSerie();
+        result.setType(TYPE.DOUBLE);
```
ftence: Until now I think all empty GTSs are of type UNDEFINED. I understand the logic but are you sure it does not break anything?
What is your opinion @hbs?

warp10/src/main/java/io/warp10/script/binary/DIV.java
```diff
@@ -152,6 +160,13 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       GeoTimeSerie result = op1gts ? ((GeoTimeSerie) op1).cloneEmpty(n) : ((GeoTimeSerie) op2).cloneEmpty();
       GeoTimeSerie gts = op1gts ? (GeoTimeSerie) op1 : (GeoTimeSerie) op2;
 
+      // Returns immediately a new clone if gts is empty.
+      if (0 == n) {
+        result.setType(TYPE.DOUBLE);
```
ftence: Ditto.

warp10/src/main/java/io/warp10/script/binary/MUL.java
```diff
@@ -68,6 +68,14 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       GeoTimeSerie gts1 = (GeoTimeSerie) op1;
       GeoTimeSerie gts2 = (GeoTimeSerie) op2;
 
+      // Returns immediately a new gts if both inputs are empty
+      if (0 == GTSHelper.nvalues(gts1) || 0 == GTSHelper.nvalues(gts2)) {
+        GeoTimeSerie result = new GeoTimeSerie();
+        result.setType(TYPE.DOUBLE);
```
ftence: Ditto.

warp10/src/main/java/io/warp10/script/binary/MUL.java
```diff
@@ -161,6 +169,13 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       GeoTimeSerie result = op1gts ? ((GeoTimeSerie) op1).cloneEmpty(n) : ((GeoTimeSerie) op2).cloneEmpty();
       GeoTimeSerie gts = op1gts ? (GeoTimeSerie) op1 : (GeoTimeSerie) op2;
 
+      // Returns immediately a new clone if gts is empty.
+      if (0 == n) {
+        result.setType(TYPE.DOUBLE);
```
ftence: Ditto.

warp10/src/main/java/io/warp10/script/binary/SUB.java
```diff
@@ -58,6 +58,14 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       GeoTimeSerie gts1 = (GeoTimeSerie) op1;
       GeoTimeSerie gts2 = (GeoTimeSerie) op2;
 
+      // Returns immediately a new gts if both inputs are empty
+      if (0 == GTSHelper.nvalues(gts1) || 0 == GTSHelper.nvalues(gts2)) {
+        GeoTimeSerie result = new GeoTimeSerie();
+        result.setType(TYPE.DOUBLE);
```
ftence: Ditto.

warp10/src/main/java/io/warp10/script/binary/SUB.java
```diff
@@ -155,6 +163,13 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       GeoTimeSerie result = op1gts ? ((GeoTimeSerie) op1).cloneEmpty(n) : ((GeoTimeSerie) op2).cloneEmpty();
       GeoTimeSerie gts = op1gts ? (GeoTimeSerie) op1 : (GeoTimeSerie) op2;
 
+      // Returns immediately a new clone if gts is empty.
+      if (0 == n) {
+        result.setType(TYPE.DOUBLE);
```
ftence: Ditto.

hbs:
Indeed, it's better to leave the GTS with an undefined type so as to allow more flexibility when the first value will be added.

pi-r-p:
I did the changed, tested it, but it leads to more inconsistent situations.

- In current release, there is an inconsistency between ADD and the other functions. ADD keep type (LONG or DOUBLE). All the other operation return a DOUBLE typed gts, whatever the input. I did not fixed this, and would like your opinion too, maybe it is time to fix this.

- If I change to non typed output in MUL, SUB, DIV, I introduce a new weird case:
```
NEWGTS 'foo' RENAME 
0 NaN NaN NaN 8 ADDVALUE
NEWGTS 'bar' RENAME
2 NaN NaN NaN 8 ADDVALUE // tick not aligned
*
NOW NaN NaN NaN '42.5' ADDVALUE 
// output gts has one point, double


NEWGTS 'foo' RENAME 
0 NaN NaN NaN 8 ADDVALUE
NEWGTS 'bar' RENAME
*
NOW NaN NaN NaN '42.5' ADDVALUE 
// output gts has one point, string

```

Are you still sure you want to let the output not typed ?

ftence:
- I think all operators should behave the same for single values as for GTSs. i.e. `2 3 /` is a LONG and `NEWGTS 0 0 0 0 2 ADDVALUE 3 /` should be a LONG GTS. This logic has to be adapted to UNDEFINED GTS though. 
- I also think it's safer to type the empty GTS, if possible. However I was wondering if it could mess with other functions which may be considering that an empty GTS is always of type UNDEFINED. Without looking too much into the details it seems there is no problem in doing that, but I'd like a second opinion.

hbs:
functions which create empty GTS (NEWGTS, CLONEMPTY) leave the type UNDEFINED. A GTS is a container for accepting data points, it only gets typed when the first value is added and all values after that will be converted, if possible, to that type

ftence:
I changed my mind and agree with @hbs because if you apply `[ SWAP MAXLONG mapper.gt 0 0 0 ] MAP` on a LONG GTS, you will have an UNDEFINED GTS, not a LONG one.
So you should remove typing of empty GTSs, even in the case of unaligned ticks.

The question of the resulting type for non-empty GTS is still open: for instance, MUL does not have the same behavior as mapper.mul or reducer.product for the product of LONGs.

pi-r-p:
Empty gts input => empty gts output, not typed.

There are other inconsistencies, listed in #698, I will correct them once this PR is merged.

PR #678, opened 2020-03-02
added WarpScriptAggregatorFunction to remaining mappers

pi-r-p:
Just for consistent `TYPEOF` output.


pi-r-p:
@ftence : Indeed, completely forget strict ones.

pi-r-p:
I also updated the doc to be consistent with this new TYPEOF.

PR #683, opened 2020-03-04
Fixes #682 - Added support for GTS Encoders in PIVOT

### LINKED ISSUE #682: Allow PIVOTing of Encoders
Hi,
To use the PIVOT function after an MVINDEXSPLIT, it seems that we have to convert the Encoders into GTSes. As Encoders are able to store values and labels, it seems that the PIVOT function should be able to take them as an input. What do you think?

hbs:


warp10/src/main/java/io/warp10/continuum/gts/GTSDecoder.java
```diff
@@ -528,26 +528,52 @@ public Object getBinaryValue() {
...
     gts.setMetadata(this.getMetadata());
-    
-    while(next()) {
-      GTSHelper.setValue(gts, getTimestamp(), getLocation(), getElevation(), getValue(), false);
+  
+    if (strict) {
+      Class lastClass = null;
+      
+      while(next()) {        
+        Object value = getValue();
```
ftence: Should binary values be also checked? If so, a more optimal way than using `getBinaryValue` should be used.

warp10/src/main/java/io/warp10/continuum/gts/GTSDecoder.java
```diff
@@ -528,26 +528,52 @@ public Object getBinaryValue() {
...
+        Class valClass = value.getClass();
+        // getValue could return a BigDecimal, we need to smooth the comparison so
+        // they are considered as Double
+        if (value instanceof BigDecimal) {
+          valClass = Double.class;
+        }
+        if (null != lastClass && !(valClass.equals(lastClass))) {
+          throw new RuntimeException("Non homogeneous GTS Encoder.");
```
ftence: Creating and using a subclass of RuntimeException could be nicer.

warp10/src/main/java/io/warp10/script/functions/PIVOT.java
```diff
@@ -45,19 +46,28 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+    List<Object> paramLabeling = (List<Object>) top;
+    List<Object> labeling = new ArrayList<Object>(paramLabeling.size());
     
     Set<String> classes = new HashSet<String>();
-    for (Object o: labeling) {
+    for (Object o: paramLabeling) {
+      if (o instanceof GTSEncoder) {
+        GTSEncoder encoder = (GTSEncoder) o;
+        GTSDecoder decoder = encoder.getDecoder(true);
+        o = decoder.decode(null, true);
```
ftence: Catch the RuntimeException to display a nice error message.

warp10/src/main/java/io/warp10/script/functions/PIVOT.java
```diff
@@ -70,15 +80,24 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-    List<Object> gts = (List<Object>) top;
+    List<Object> paramGts = (List<Object>) top;
+    List<Object> gts = new ArrayList<Object>(paramGts.size());
     
-    for (Object o: gts) {
+    for (Object o: paramGts) {
+      if (o instanceof GTSEncoder) {
+        GTSEncoder encoder = (GTSEncoder) o;
+        GTSDecoder decoder = encoder.getDecoder(true);
+        o = decoder.decode(null, true);
```
ftence: Ditto.

warp10/src/main/java/io/warp10/script/functions/PIVOT.java
pi-r-p: I agree. It will could be usefull for other functions too.

warp10/src/main/java/io/warp10/continuum/gts/GTSDecoder.java
hbs: The only reason it's a RuntimeException is that it avoids declaring it is thrown, no plan to subclass it.

PR #686, opened 2020-03-07
Added support for listing and aborting WarpScript executions

hbs:


warp10/src/main/java/io/warp10/script/ext/stackps/StackPSWarpScriptExtension.java
```diff
@@ -0,0 +1,61 @@
...
+   *  Name of configuration key with the stackps secret. 
+   */
+  public static final String CONF_STACKPS_SECRET = "stackps.secret";
+  
+  /**
+   * Current StackPS Secret
+   */
+  public static String STACKPS_SECRET;  
+
+  public static final String ATTRIBUTE_SESSION = "stackps.session";
```
pi-r-p: Why isn't session name an attribute of the stack, just like attribute_name ?

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
```diff
@@ -138,7 +139,8 @@ public void handle(String target, Request baseRequest, HttpServletRequest req, H
     //
     
     WarpScriptStack stack = new MemoryWarpScriptStack(this.storeClient, this.directoryClient);
-
+    stack.setAttribute(WarpScriptStack.ATTRIBUTE_NAME, "[EgressExecHandler " + Thread.currentThread().getName() + "]");
+
```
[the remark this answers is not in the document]
hbs: Why can't they add a prefix to the WarpScript code they push?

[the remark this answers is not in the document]
pi-r-p: - It is more simple than prepend lines, shift line numbers or statement number in error messages. VSCode already does macro inlining and error message transformations, studio does not.
- It does not appear in the 'warpscript.script' WarpScript logs.
- More flexible to enforce some extra policy later, for example include the user name in each request, or to add a proxy that automatically add the incoming IP or other metadata in attribute_name if a company want to find who is taking all the shared resources. (128 character is a bit too short for that)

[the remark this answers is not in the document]
hbs: You don't need to shift lines as you could very well set the session on the first line without a LF. 
Having an alternate way of setting things just for the /exec endpoint brings redundancy and lack of coherency without added value, therefore this will not be added.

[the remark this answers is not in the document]
hbs: Second thoughts, added possibility to set the session via header (but not the name), but disallowed overwriting the session in WSSESSION.

PTAL.

warp10/src/main/java/io/warp10/script/ext/stackps/WSPS.java
```diff
@@ -0,0 +1,66 @@
...
+
+    List<Object> results = new ArrayList<Object>();
+    
+    for (WarpScriptStack stck: WarpScriptStackRegistry.stacks()) {
+      List<Object> result = new ArrayList<Object>();
+      
+      result.add(stck.getUUID());
+      result.add(stck.getAttribute(WarpScriptStack.ATTRIBUTE_CREATION_TIME));
+      result.add(stck.getAttribute(WarpScriptStack.ATTRIBUTE_NAME));
+      results.add(result);
```
pi-r-p: also add session here
      result.add(stck.getAttribute(StackPSWarpScriptExtension.ATTRIBUTE_SESSION));

warp10/src/main/java/io/warp10/script/ext/stackps/WSPS.java
```diff
@@ -0,0 +1,66 @@
...
+    }
```
pi-r-p: Will be better to replace the list output by a list of map, with keys such as "session_name" "warpscript_name" and so on. It allow greater flexibility in the future, and easier javascript reading for guys who will code frontend.

warp10/src/main/java/io/warp10/script/ext/stackps/WSINFO.java
```diff
@@ -0,0 +1,96 @@
...
+
+public class WSINFO extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+  
+  public static final String[] EXPOSED_ATTRIBUTES = new String[] {
+      WarpScriptStack.ATTRIBUTE_CREATION_TIME,
+      WarpScriptStack.ATTRIBUTE_FETCH_COUNT,
+      WarpScriptStack.ATTRIBUTE_GTS_COUNT,
+      WarpScriptStack.ATTRIBUTE_MACRO_NAME,
+      WarpScriptStack.ATTRIBUTE_NAME,
+      WarpScriptStack.ATTRIBUTE_SECTION_NAME,
```
pi-r-p: Missing: 
     StackPSWarpScriptExtension.ATTRIBUTE_SESSION

(could be relocated in WarpScriptStack maybe ?)

hbs: nope since it is added by an extension it is more logical to have it in the extension itself

pi-r-p: Yes for location, but having session name among infos in WSPS is great too.

pi-r-p:
First implementation done in VScode extension. As discussed, it could be great to have a "session info" function that returns information about user scripts for one session, without knowing the server `stackps.secret`. 
For better security, session name could also have a minimum length.

warp10/src/main/java/io/warp10/script/ext/stackps/WSKILL.java
```diff
@@ -0,0 +1,58 @@
...
+        throw new WarpScriptException(getName() + " invalid secret.");
+      }        
+      top = stack.pop();
+    }      
+
+    if (!(top instanceof String)) {
+      throw new WarpScriptException(getName() + " expects a UUID.");
+    }
+    
+    stack.push(WarpScriptStackRegistry.abort(top.toString()));
```
ftence: Same comment as for `WSABORT`: do we make a specific case for own stack?

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -514,6 +518,10 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
                 throw new WarpScriptException(getName() + " exceeded limit of " + fetchLimit + " datapoints, current count is " + fetched.get());
               }
 
+              if (stack.aborted()) {
+                throw new WarpScriptException(getName() + " execution aborted.");
+              }
+
```
ftence: I think it should be inside the `while` several line above, not after. It could be added after line 511 for instance.

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -562,6 +571,10 @@ public void exec(String line) throws WarpScriptException {
       //
 
       for (int st = 0; st < statements.length; st++) {
+        if (this.aborted) {
+          throw new WarpScriptException("Execution aborted.");
```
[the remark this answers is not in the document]
hbs: Aborted stacks should never be reused, this is the whole point of aborting them. Are you suggesting that we should be able to externally STOP a stack on top of being able to abort one?

hbs:
PTAL

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -1614,14 +1617,39 @@ private String rewriteMacroSymbol(String symbol) {
   }
   
   @Override
-  public void abort() {
+  public void signal(Signal signal) {
+    // Only set the signal if the stack is not yet signaled so a KILL is not
+    // overriden by anything else for example
+    if (this.signaled) {
+      return;
+    }
```
ftence: We should still allow overriding a STOP by a KILL.

warp10/src/main/java/io/warp10/script/WarpScriptStack.java
```diff
@@ -739,14 +744,14 @@ private void ensureCapacity(int n) {
...
    */
-  public void abort();
+  public void signal(Signal signal);
   
   /**
-   * Return the state of execution abortion.
+   * Throw the exception associated with the current signal sent to the stack
    */
-  public boolean aborted();
+  public void signal() throws WarpScriptATCException;
```
ftence: I find `signal()` to be confusing, can it be called `checkSignal()`?

hbs: It's not a check, it's a triggering of a buffered signal, so maybe triggerSignal?

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
hbs: So enforce a hierarchy of signals

warp10/src/main/java/io/warp10/script/ext/stackps/WSNAME.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+    
+    String name = top.toString();
+    
+    if (name.length() > MAX_SIZE) {
+      throw new WarpScriptException(getName() + " expects a name less than " + MAX_SIZE + " characters.");
+    }
+    
+    if (session) {
+      if (null != stack.getAttribute(StackPSWarpScriptExtension.ATTRIBUTE_SESSION)) {
```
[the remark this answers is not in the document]
hbs: The idea was the opposite, if it already has a session name you cannot change the session name

warp10/src/main/java/io/warp10/script/ext/stackps/WSINFO.java
```diff
@@ -0,0 +1,97 @@
...
+    
+    Map<Object,Object> infos = null;
+
+    String uuid = top.toString();
+    
+    for (WarpScriptStack stck: WarpScriptStackRegistry.stacks()) {
+      if (uuid.equals(stck.getUUID())) {
+        infos = getInfos(stck);
+      }
+    }
```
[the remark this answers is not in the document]
hbs: It's actually the session name we use to identify the stack, not the UUID, so it is not readily available in the underlying map.

warp10/src/main/java/io/warp10/script/ext/stackps/WSKILL.java
hbs: use with caution, but no

warp10/src/main/java/io/warp10/script/ext/stackps/WSABORT.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Override
+  public Object apply(WarpScriptStack stack) throws WarpScriptException {
+    Object top = stack.pop();
+    
+    if (!(top instanceof String)) {
+      throw new WarpScriptException(getName() + " expects a session name (STRING).");
+    }
+    
+    stack.push(WarpScriptStackRegistry.abortSession(top.toString()));
```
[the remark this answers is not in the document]
hbs: use with caution, but no

warp10/src/main/java/io/warp10/script/ext/stackps/WSINFO.java
```diff
@@ -0,0 +1,99 @@
...
+//    if (null != secret) {     
+//      if (!(top instanceof String)) {
+//        throw new WarpScriptException(getName() + " expects a secret.");
+//      }
+//      if (!secret.equals(top)) {
+//        throw new WarpScriptException(getName() + " invalid secret.");
+//      }
+//      
+//      top = stack.pop();
+//    }
```
ftence: Is there a reason this comment is here?

hbs: sediment has been cleared

warp10/src/main/java/io/warp10/script/WarpScriptStackRegistry.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+  
+  public static void enable() {
+    enabled = true;
+  }
+  
+  public static void disable() {
+    // Clear the registered stacks
+    stacks.clear();
+    enabled = false;
```
ftence: Because none of the methods are synchronized, there could be stacks in `stacks` after this registry is disabled.
I think it does not matter that much but as a precaution, I would allow `unregister` even if it is disabled.

hbs:
PTAL

PR #687, opened 2020-03-09
Feat/delete/cleanseries

aurrelhebert:
This is a first proposal to enhance #687

### DIFF
--- a/warp10/src/main/thrift/io_warp10_continuum_store_thrift_data.thrift
+++ b/warp10/src/main/thrift/io_warp10_continuum_store_thrift_data.thrift
@@ -64,6 +64,12 @@ struct Metadata {
    * depends on the configuration of the activity window in Warp 10.
    */
   7: optional i64 lastActivity,  
+  
+  /**
+    * Optional field expecting a duration (in ms) corresponding to a security to clean Geo Time Series.
+    * used only on the Delete endpoint
+    */
+  8: optional i64 cleanttl,  
 }
 
 /**
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ warp10/src/main/resources/REVISION
 gradle.properties
 archive
 /CHANGELOG.md
+.vscode
--- a/build.gradle
+++ b/build.gradle
@@ -216,7 +216,6 @@ subprojects {
         repositories {
             repositories.mavenDeployer {
                 configuration = configurations.deployerJars
-                uniqueVersion = false
 
                 repository(url: "scpexe://${stagingURL}") {
                     authentication userName: stagingUser
--- a/warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
+++ b/warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
@@ -1451,6 +1451,7 @@ public class Ingress extends AbstractHandler implements Runnable {
     long gts = 0;
     
     boolean completeDeletion = false;
+    boolean cleanDeletion = null != request.getParameter(Constants.HTTP_PARAM_CLEAN_TTL);
 
     boolean dryrun = null != request.getParameter(Constants.HTTP_PARAM_DRYRUN);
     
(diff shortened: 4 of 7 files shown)


aurrelhebert:
I have run some tests about this code, everything is working find with an ingress up to date, however using an old version of the ingress with this patch will behave as follow: 
- update is ok
- update then deleting the series (normal or clean) and afterward re-pushing a point will NOT create the series 

How did you manage to handle it correctly when the lastactivity was added ?

warp10/src/main/thrift/io_warp10_continuum_store_thrift_data.thrift
```diff
@@ -64,6 +64,12 @@ struct Metadata {
    * depends on the configuration of the activity window in Warp 10.
    */
   7: optional i64 lastActivity,  
+  
+  /**
+    * Optional field expecting a duration (in ms) corresponding to a security to clean Geo Time Series.
+    * used only on the Delete endpoint
+    */
+  8: optional i64 cleanttl,
```
StevenLeRoux: Prefer Camel Case : cleanTTL

hbs:
Thanks for taking the time of submitting this PR, however it has several problems.

Any feature, except for some very specific area of the Warp 10 code should be introduced in both the standalone and distributed versions of Warp 10 so users can expect to have a smooth experience when moving back and forth between the two.

The way you handle the parameters is problematic, if the actual value of your cleanttl query parameter is not parseable as a Long then your code will fail even though the delete message has already been sent.

The Directory is designed to be a dumb soldier and execute what it is told to do through the messages it consumes off of its Kafka topic. Your PR introduces some decision into the Directory itself which is not something suitable as it would break the current assignment of responsabilities, making it more difficult to maintain. As an example your code does not correctly handle the case when a Directory plugin is defined. It also has the drawback of not returning to the user the actual list of series which will be deleted since the deletion decision is made in the Directory which does not have a public endpoint.

Lastly, even though the decision should not be made in Directory (see above), using the Metadata structure as a vehicle for the ttl parameter should be discouraged, the impacts of modifying this structure has a leverage effect which can cause vas amounts of memory to be consumed when your series are in the 10s of millions as your initial issue seems to indicate.

As it turns out, as part of the Warp 10 Accelerator PR which is currently open, we had a change ready to add the possibility to specify activeafter/quietafter parameters to /delete requests. This change,  augmented with the possibility to simply delete the metadata i.e. not the data points themselves could very well address your issue, i.e. delete series which were quiet after a given timestamp and only delete the Metadata. This change was planned to appear in a PR of its own, feel free to indicate if it needs some extra modifications to address all the points in your issue when it shows up. We were waiting for the Warp 10 Accelerator PR to get more scrutiny first but we could fast track it if necessary. The PR makes those possibilities optional so environments where deleting metadata without deleting data is not suitable will not allow it.

aurrelhebert:
Great news! Let's wait for this PR then. How will you handle the activeafter/quietafter as the last activity is only available in the directory? However I didn't understood how the delete endpoint and the Warp 10 Accelerator are related?

hbs:
The lastactivy is kept track of in Directory, what the PR does is that it adds the possibility to select GTS to delete based on activeafter/quietafter just like FIND or FETCH requests. The lastactivity is then checked at the Directory level at the time of GTS selection, the selected GTS are then returned to the caller and the /delete request will only delete those.

The introduction of those parameters was deemed necessary when we worked on the Accelerator feature, not directly because of the accelerator but because the use of the in memory store suffers from the exact same 'dead series' problem since the in-memory store basically enforces a TTL on the data, this problem popped up as we worked on the accelerator with low level deletion of LevelDB files, i.e. without calls to /delete, the persisted data would disappear, so would its in-memory counterpart after a while but the actual series remained in Directory.

hbs:
You might want to experiment with PR #690

aurrelhebert:
That's what I had in mind. I wasn't aware of the activeafter/quietafter requests methods : 
I will run some tests in the afternoon to provide some feedback's on this PR ;)

PR #690, opened 2020-03-10
Adds activeafter/quietafter params to /delete requests and a metaonly option

hbs:


warp10/src/main/java/io/warp10/continuum/store/Constants.java
```diff
@@ -376,6 +376,7 @@
   public static final String HTTP_PARAM_SAMPLE = "sample";
   public static final String HTTP_PARAM_PREBOUNDARY = "boundary.pre";
   public static final String HTTP_PARAM_POSTBOUNDARY = "boundary.post";
+  public static final String HTTP_PARAM_NODATA = "nodata";
```
aurrelhebert: I would propose `deletemeta` as HTTP parameter here instead of "nodata". IMO, this would made more sence for a user

warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
```diff
@@ -1545,6 +1545,22 @@ public void handleDelete(String target, Request baseRequest, HttpServletRequest
         throw new IOException("Parameter " + Constants.HTTP_PARAM_DELETEALL + " should be set when deleting a full range.");
```
aurrelhebert: When the `nodata` (or deletemeta) option is set, we should consider it's correspond to a `deleteall`call and not throw this error.

warp10/src/main/java/io/warp10/continuum/store/Constants.java
hbs: Renamed 'metaonly', 'deletemeta' could incorrectly lead users to think that 'deleteall' is related.

aurrelhebert:
When do you plan to merge this PR on master?

hbs:
Probably within a week, the lag in the PR processing pipeline is being reduced

warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
```diff
@@ -1772,7 +1807,10 @@ public void close() throws Exception {
                   continue;
                 }
               }
-              pushDeleteMessage(start, end, minage, metadata);
+              
+              if (!metaonly) {
+                pushDeleteMessage(start, end, minage, metadata);
+              }
```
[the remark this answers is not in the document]
hbs: It does not change much as the pushDeleteMessages in the finally clause simply flushes unsent messages, but I added a conditional. For completeDeletion, it will be set to true when metaonly is specified so the code within the conditional will be called.

PR #691, opened 2020-03-11
Added configuration of macro cache size

hbs:


warp10/src/main/java/io/warp10/script/WarpScriptMacroLibrary.java
```diff
@@ -46,8 +49,25 @@
...
+    
+    final int maxcachesize = Integer.parseInt(WarpConfig.getProperty(Configuration.WARPSCRIPT_LIBRARY_CACHE_SIZE, Integer.toString(DEFAULT_CACHE_SIZE)));
+    
+    macros = new LinkedHashMap<String,Macro>() {
+      @Override
+      protected boolean removeEldestEntry(java.util.Map.Entry<String,Macro> eldest) {
+        return this.size() > maxcachesize;
+      }
+    };
+
```
pi-r-p: What about a warning somewhere in the logs to tell user that some macro are not in the cache ?

etc/conf.templates/distributed/20-macros.conf.template
```diff
@@ -40,4 +40,8 @@
 // Maximum TTL that can be set using MACROTTL (defaults to 2**62 ms)
 #warpscript.repository.ttl.hard = 
 
+// Number of macros loaded from 'warpscript.repository.directory' to keep in memory
+warpscript.repository.cache.size = 10000
 
+// Size of macro cache for the macros loaded from the classpath
+warpscript.library.cache.size = 10000
```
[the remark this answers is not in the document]
hbs: Size of caches is measured in number of entries, not actual memory footprint

warp10/src/main/java/io/warp10/script/WarpScriptMacroLibrary.java
hbs: This warning would be written every time a macro would be loaded once you reach the limit. Setting a Sensision metric with the number of cached macros is probably a better option.

etc/conf.templates/distributed/20-macros.conf.template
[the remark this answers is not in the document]
pi-r-p: So the second comment should be "Number of macros loaded from the classpath" ?

warp10/src/main/java/io/warp10/script/WarpScriptMacroLibrary.java
pi-r-p: Indeed. With 10000 as default, I doubt this will happen too often.

etc/conf.templates/distributed/20-macros.conf.template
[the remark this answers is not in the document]
hbs: Nope, this is the maximum size of the cache, not the number of loaded macros

warp10/src/main/java/io/warp10/script/WarpScriptMacroLibrary.java
hbs: Upated.

warp10/src/main/java/io/warp10/script/WarpScriptMacroLibrary.java
```diff
@@ -185,6 +237,20 @@ public static Macro loadMacro(Object root, InputStream in, String name) throws W
...
+        macroTtl = (long) stack.getAttribute(WarpScriptStack.ATTRIBUTE_MACRO_TTL);
+      }
+      
+      if (macroTtl > hardTtl) {
+        macroTtl = hardTtl;
+      }
+
+      // Set expiry. Note using a ttl too long will wrap around the sum and will
+      // make the macro expire too early
+      macro.setExpiry(System.currentTimeMillis() + ttl);
```
ftence: Why not checking for overflow and not setting expiry if it overflows?

warp10/src/main/java/io/warp10/script/WarpScriptMacroRepository.java
```diff
@@ -102,8 +104,24 @@
...
+    //
+    // Create macro map
+    //
+    
+    final int maxcachesize = Integer.parseInt(WarpConfig.getProperty(Configuration.REPOSITORY_CACHE_SIZE, Integer.toString(DEFAULT_CACHE_SIZE)));
+    
+    macros = new LinkedHashMap<String,Macro>() {
+      @Override
+      protected boolean removeEldestEntry(java.util.Map.Entry<String,Macro> eldest) {
+        return this.size() > maxcachesize;
```
ftence: Is there a reason there is a sensision metric for the library and not for the repo?

ftence: Same question for logging evicted macros.

hbs: Macros which are evicted upon loading of a jar will never get re-added. In the case of the repo, when ondemand is true the macros will get loaded anyway. The ondemand param is true by default so we could log eviction only when it is false.

warp10/src/main/java/io/warp10/script/WarpScriptMacroLibrary.java
```diff
@@ -249,7 +251,12 @@ public static Macro loadMacro(Object root, InputStream in, String name) throws W
 
       // Set expiry. Note using a ttl too long will wrap around the sum and will
       // make the macro expire too early
-      macro.setExpiry(System.currentTimeMillis() + ttl);
+      
+      try {
+        macro.setExpiry(Math.addExact(System.currentTimeMillis(), ttl));
+      } catch (ArithmeticException ae) {
+        macro.setExpiry(Long.MAX_VALUE - 1);
```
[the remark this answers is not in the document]
hbs: To avoid having weird comparison results if checking with >

PR #692, opened 2020-03-11
Modified validation of GTS list.

hbs:
When using the 'gts' parameter of the FETCH input map, we now perform some more advanced checks to ensure that the provided READ token would actually select the given GTS. We also replace somme labels (namely owner or producer) if absent by a value extracted from the token and which makes the GTS fall within the set of GTS which could be selected with the given token.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/FETCH.java
+++ b/warp10/src/main/java/io/warp10/script/functions/FETCH.java
@@ -54,6 +54,7 @@ import io.warp10.continuum.gts.GTSDecoder;
 import io.warp10.continuum.gts.GTSHelper;
 import io.warp10.continuum.gts.GeoTimeSerie;
 import io.warp10.continuum.gts.GeoTimeSerie.TYPE;
+import io.warp10.continuum.gts.MetadataSelectorMatcher;
 import io.warp10.continuum.sensision.SensisionConstants;
 import io.warp10.continuum.store.Constants;
 import io.warp10.continuum.store.DirectoryClient;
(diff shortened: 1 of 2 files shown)


warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -235,32 +236,83 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+        //
+        // If the metadata would not get selected by the provided token
+        // force the producer/owner/app to be that of the token
+        //
+        
+        if (!matches) {
+          m.getLabels().remove(Constants.PRODUCER_LABEL);
+          m.getLabels().remove(Constants.OWNER_LABEL);
+          m.getLabels().remove(Constants.APPLICATION_LABEL);
+          m.getLabels().putAll(tokenSelectors);
```
ftence: You can't use that because it will add a `=` before each label value, which won't be overridden in case there is a producer and no owner or no producer and an owner.
At the end the labelID will be wrong.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -235,32 +236,84 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-          throw new WarpScriptException(getName() + " provided token is incompatible with '" + PARAM_GTS + "' parameter, expecting a single application.");
+        //
+        // If the metadata would not get selected by the provided token
+        // force the producer/owner/app to be that of the token
+        //
+        
+        if (!matches) {
+          m.getLabels().remove(Constants.PRODUCER_LABEL);
+          m.getLabels().remove(Constants.OWNER_LABEL);
+          m.getLabels().remove(Constants.APPLICATION_LABEL);
```
ftence: I understand that this is for clarity and safeness but as they are always overridden, this is not needed.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -235,32 +236,84 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+            //
+            m.getLabels().put(Constants.PRODUCER_LABEL, tokenSelectors.get(Constants.PRODUCER_LABEL).substring(1));
+            m.getLabels().put(Constants.OWNER_LABEL, tokenSelectors.get(Constants.OWNER_LABEL).substring(1));            
+          } else if (singleProducer && !tokenSelectors.containsKey(Constants.OWNER_LABEL)) {
+            //
+            // If the token has a single producer but no owner, use the producer as the owner, this would
+            // lead to a narrower scope than what the token would actually select so it is fine.
+            //
+            m.getLabels().put(Constants.PRODUCER_LABEL, tokenSelectors.get(Constants.PRODUCER_LABEL).substring(1));
+            m.getLabels().put(Constants.OWNER_LABEL, m.getLabels().get(Constants.PRODUCER_LABEL));
```
ftence: Using a variable containing `tokenSelectors.get(Constants.PRODUCER_LABEL).substring(1)` would be better IMHO.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -235,32 +236,84 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+            //
+            m.getLabels().put(Constants.PRODUCER_LABEL, tokenSelectors.get(Constants.PRODUCER_LABEL).substring(1));
+            m.getLabels().put(Constants.OWNER_LABEL, m.getLabels().get(Constants.PRODUCER_LABEL));                        
+          } else if (singleOwner && !tokenSelectors.containsKey(Constants.PRODUCER_LABEL)) {
+            //
+            // If the token has a single owner but no producer, use the owner as the producer, again this would
+            // lead to a narrower scope than what the token can actually access so it is fine too.
+            //
+            m.getLabels().put(Constants.OWNER_LABEL, tokenSelectors.get(Constants.OWNER_LABEL).substring(1));            
+            m.getLabels().put(Constants.PRODUCER_LABEL, m.getLabels().get(Constants.OWNER_LABEL));
```
ftence: Same thing with `tokenSelectors.get(Constants.OWNER_LABEL).substring(1)`.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -235,32 +236,86 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+            // If the token has a single producer and single owner, use them for the GTS
+            //
+            m.getLabels().put(Constants.PRODUCER_LABEL, tokenSelectors.get(Constants.PRODUCER_LABEL).substring(1));
+            m.getLabels().put(Constants.OWNER_LABEL, tokenSelectors.get(Constants.OWNER_LABEL).substring(1));            
+          } else if (singleProducer && !tokenSelectors.containsKey(Constants.OWNER_LABEL)) {
+            //
+            // If the token has a single producer but no owner, use the producer as the owner, this would
+            // lead to a narrower scope than what the token would actually select so it is fine.
+            //
+            String producer = tokenSelectors.get(Constants.PRODUCER_LABEL).substring(1);
```
ftence: Can be defined outside the loop.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -235,32 +236,86 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+            m.getLabels().put(Constants.PRODUCER_LABEL, producer);
+            m.getLabels().put(Constants.OWNER_LABEL, producer);                        
+          } else if (singleOwner && !tokenSelectors.containsKey(Constants.PRODUCER_LABEL)) {
+            //
+            // If the token has a single owner but no producer, use the owner as the producer, again this would
+            // lead to a narrower scope than what the token can actually access so it is fine too.
+            //
+            String owner = tokenSelectors.get(Constants.OWNER_LABEL).substring(1);
```
ftence: Ditto.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -235,32 +236,86 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+          } else {
+            throw new WarpScriptException(getName() + " provided token is incompatible with '" + PARAM_GTS + "' parameter, expecting a single application.");
+          }
+
+          if (singleProducer && singleOwner) {
+            //
+            // If the token has a single producer and single owner, use them for the GTS
+            //
+            m.getLabels().put(Constants.PRODUCER_LABEL, tokenSelectors.get(Constants.PRODUCER_LABEL).substring(1));
+            m.getLabels().put(Constants.OWNER_LABEL, tokenSelectors.get(Constants.OWNER_LABEL).substring(1));
```
ftence: Use the variables which will be defined because of the comments below.

PR #693, opened 2020-03-12
Added FOREACH iteration on STRINGs

hbs:


warp10/src/main/java/io/warp10/script/functions/FOREACH.java
```diff
@@ -69,12 +69,29 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
     }
 
     long index = 0;
-
+    
+    if (obj instanceof String) {
+      final String s = (String) obj;
+      
+      obj = new Iterator<String>() {        
+        int idx = 0;
```
pi-r-p: Default method 'remove' is not overridden. It would cause compilation problems with JDK 7.

Do we still care about jdk 7 ?

hbs: JDK7 will be unsupported in future releases. As of 2.4.0 I am not sure everything actually works in JDK7 anyway, and I really doubt any deployment of Warp 10 actually uses JDK7 as its JVM so I think this can be left as is.

PR #696, opened 2020-03-16
Enable OSS support in TOKENGEN and TOKENDUMP for the secret management

Giwi:


warp10/src/main/java/io/warp10/script/ext/token/TokenWarpScriptExtension.java
```diff
@@ -16,57 +16,58 @@
...
-  private static final Map<String,Object> functions = new HashMap<String,Object>();
+
+  private static final Map<String, Object> functions = new HashMap<String, Object>();
   private static final KeyStore keystore;
-  
+
   static {
     TOKEN_SECRET = WarpConfig.getProperty(CONF_TOKEN_SECRET);
+    if(null != Warp.getKeyStore().getKey(TokenWarpScriptExtension.CONF_TOKEN_SECRET)){
+      TOKEN_SECRET = new String(Warp.getKeyStore().getKey(TokenWarpScriptExtension.CONF_TOKEN_SECRET)).replaceAll("\n", "").trim();
```
hbs: Don't call getKeyStore multiple times, the whole keystore gets cloned when you do so.

Specify StandardCharsets.UTF_8 when creating the String instance.

warp10/src/main/java/io/warp10/script/ext/token/TokenWarpScriptExtension.java
```diff
@@ -16,57 +16,58 @@
...
-  
-  private static final Map<String,Object> functions = new HashMap<String,Object>();
+
+  private static final Map<String, Object> functions = new HashMap<String, Object>();
   private static final KeyStore keystore;
-  
+
   static {
     TOKEN_SECRET = WarpConfig.getProperty(CONF_TOKEN_SECRET);
+    if(null != Warp.getKeyStore().getKey(TokenWarpScriptExtension.CONF_TOKEN_SECRET)){
```
hbs: Warp.getKeyStore could return null

warp10/src/main/java/io/warp10/script/ext/token/TokenWarpScriptExtension.java
```diff
@@ -16,57 +16,58 @@
...
 
+import java.util.HashMap;
+import java.util.Map;
+
 public class TokenWarpScriptExtension extends WarpScriptExtension {
-  
+
   /*
-   *  Name of configuration key with the token secret. 
+   *  Name of configuration key with the token secret.
```
hbs: Add a comment here and in the documentation templates explaining that warp.key.token.secret will prevail if it exists.

warp10/src/main/java/io/warp10/json/JsonUtils.java
```diff
@@ -31,11 +31,6 @@
 import com.fasterxml.jackson.databind.ser.BeanSerializer;
 import com.fasterxml.jackson.databind.ser.BeanSerializerModifier;
 import com.fasterxml.jackson.databind.ser.impl.UnknownSerializer;
```
hbs: This part of the PR doesn't have much to do with the rest!

Giwi: I know, it will be a part of a second PR.

warp10/src/main/java/io/warp10/script/ext/token/TokenWarpScriptExtension.java
```diff
@@ -16,57 +16,61 @@
...
   private static final KeyStore keystore;
-  
+
   static {
     TOKEN_SECRET = WarpConfig.getProperty(CONF_TOKEN_SECRET);
 
-    if (null != TOKEN_SECRET) {
+    if (null == TOKEN_SECRET) {
       keystore = Warp.getKeyStore();
+      if (null != keystore && null != keystore.getKey(TokenWarpScriptExtension.CONF_TOKEN_SECRET)) {
```
hbs: Extract the key only once with nested ifs.

warp10/src/main/java/io/warp10/script/ext/token/TokenWarpScriptExtension.java
```diff
@@ -16,57 +16,61 @@
...
   public static String TOKEN_SECRET = null;
-  
-  private static final Map<String,Object> functions = new HashMap<String,Object>();
+
+  private static final Map<String, Object> functions = new HashMap<String, Object>();
   private static final KeyStore keystore;
-  
+
   static {
     TOKEN_SECRET = WarpConfig.getProperty(CONF_TOKEN_SECRET);
```
hbs: Use separate constants to make the code clearer, even though they have the same value. CONF_TOKEN_SECRET for the configuration key, KEY_TOKEN_SECRET for the KeyStore key. Clearly describe the behavior in the comments. As it is now it is rather hard to understand since you say "Myst be warp.key.token.secret is handled by OSS" with a constant defined below. What should I do if I want to use OSS? Recompile the code with a different value of CONF_TOKEN_SECRET?

warp10/src/main/java/io/warp10/script/ext/token/TokenWarpScriptExtension.java
```diff
@@ -16,57 +16,73 @@
...
-
-    if (null != TOKEN_SECRET) {
-      keystore = Warp.getKeyStore();
+    if (null == TOKEN_SECRET) {
+      // if OSS wrapped secret exists
+      TOKEN_SECRET = WarpConfig.getProperty(KEY_TOKEN_SECRET);
```
hbs: Unwrapping of any configuration key prefixed with Configuration.WARP_KEY_PREFIX is automatic, the only thing you have to do is to attempt to retrieve the KEY_TOKEN_SECRET key and use it if it is set.

So KEY_TOKEN_SECRET should be "token.secret"

etc/conf.templates/standalone/70--extensions.conf.template
```diff
@@ -50,9 +50,14 @@
 #warpscript.extension.sensision = io.warp10.script.ext.sensision.SensisionWarpScriptExtension
 // MUTEX, SHMLOAD, SHMSTORE
 #warpscript.extension.shm = io.warp10.script.ext.shm.SharedMemoryWarpScriptExtension
+
```
hbs: You need to modify the distributed configuration template too.

etc/conf.templates/standalone/70--extensions.conf.template
```diff
@@ -50,9 +50,14 @@
...
 // TOKENDUMP, TOKENGEN, TOKENSECRET
 #warpscript.extension.token = io.warp10.script.ext.token.TokenWarpScriptExtension
+// If you use OSS to handle this secret
```
hbs: Explain more clearly how this is taken into account vs token.secret

warp10/src/main/java/io/warp10/script/ext/token/TokenWarpScriptExtension.java
Giwi: indeed

warp10/src/main/java/io/warp10/script/ext/token/TokenWarpScriptExtension.java
```diff
@@ -35,7 +35,7 @@
    * Name of configuration key with OSS wrapped token secret
```
hbs: It's actually the name of the secret in the key store, not in the configuration.

warp10/src/main/java/io/warp10/script/ext/token/TokenWarpScriptExtension.java
```diff
@@ -16,57 +16,73 @@
...
-  
+
   static {
     TOKEN_SECRET = WarpConfig.getProperty(CONF_TOKEN_SECRET);
-
-    if (null != TOKEN_SECRET) {
-      keystore = Warp.getKeyStore();
+    if (null == TOKEN_SECRET) {
+      // if OSS wrapped secret exists
+      TOKEN_SECRET = WarpConfig.getProperty(KEY_TOKEN_SECRET);
```
hbs: This should test the existence of the key in the keystore, i.e. Warp.getKeyStore().getKey(KEY_TOKEN_SECRET) and then set TOKEN_SECRET accordingly, here this would always set TOKEN_SECRET to null since you already checked the same configuration key.

warp10/src/main/java/io/warp10/script/ext/token/TokenWarpScriptExtension.java
```diff
@@ -16,57 +16,73 @@
...
+      if (null == TOKEN_SECRET) { // if no configuration key
+        keystore =  Warp.getKeyStore();
+      } else {
+        keystore = null;
+        KeyStore keyStore = Warp.getKeyStore();
```
hbs: Rename this variable ks so it differs frankly from keystore

PR #699, opened 2020-03-22
Attribute check for tokens

hbs:
Add the ability to define a list of token attributes which are banned. Any token bearing one of those attributes will be rejected with an error by the standard token parsing code.

The ability to disable those checks is offered so authentication plugins can rely on token extraction even if the token is bearing a banned attribute.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/Tokens.java
+++ b/warp10/src/main/java/io/warp10/continuum/Tokens.java
@@ -70,7 +70,7 @@ public class Tokens {
   /**
    * If set, skip attribute checks
    */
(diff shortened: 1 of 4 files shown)


warp10/src/main/java/io/warp10/continuum/Tokens.java
```diff
@@ -65,6 +65,31 @@
   
   private static List<AuthenticationPlugin> plugins = new ArrayList<>();
   
+  private static final List<String> blockedAttributes;
+  
+  /**
+   * If set, skip attribute checks
+   */
+  private static ThreadLocal<Boolean> skipCheckAttributes = new ThreadLocal<Boolean>() {
```
ftence: Can be final.

PR #703, opened 2020-03-25
Set the source and target to Java 8, making this project only support Java 8+.

ftence:
This is to reflect the fact that already a lot of code is only compatible with Java 8+, mainly because of ` Math.*Exact()` and `Files.walk()`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -60,8 +60,8 @@ subprojects {
 
     group = 'io.warp10'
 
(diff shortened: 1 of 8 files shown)


hbs:
You could then remove the few checks that still exist to determine if the JDK is 8+ or not (SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_1_8)).

PR #707, opened 2020-03-27
Added TypeResolver support to TYPEOF

hbs:


warp10/src/main/java/io/warp10/script/functions/TYPEOF.java
```diff
@@ -176,5 +193,12 @@ private static String defaultType(Class c) {
       return "X-" + canonicalName;
     }
   }
+  
+  public synchronized static void addResolver(TypeResolver resolver) {
```
ftence: `public static synchronized void` should be preferred.

PR #714, opened 2020-04-01
SQL Support

sir-apollof:


hbs:
Pity your timing was off, we just released 2.5.0! And so far sql support is not on the roadmap.

PR #718, opened 2020-04-09
Geounpack loose

hbs:


warp10/src/main/java/io/warp10/script/functions/GEOUNPACK.java
```diff
@@ -31,6 +31,7 @@
 
 import com.geoxp.GeoXPLib;
 import com.geoxp.GeoXPLib.GeoXPShape;
+import com.geoxp.oss.jarjar.org.bouncycastle.util.Arrays;
```
ftence, line 34: Wrong import, should be `import java.util.Arrays;`

hbs:
PTAL

PR #719, opened 2020-04-10
Wrapmv nocomp

hbs:


warp10/src/main/java/io/warp10/script/functions/WRAP.java
```diff
@@ -90,7 +86,7 @@ public Object applyOnElement(Object element) throws WarpScriptException {
             wrapper = GTSWrapperHelper.fromGTSEncoderToGTSWrapper((GTSEncoder) element, compress);
           }
         } else {
-          throw new WarpScriptException(getName() + " expects a Geo Time Series™ of a GTSEncoder or a list on top of the stack");
+          throw new WarpScriptException(getName() + " expects a Geo Time Series of a GTSEncoder or a list on top of the stack");
```
ftence, line 89: Typo "of" -> "or" and you can change to "a list thereof".

PR #720, opened 2020-04-10
Avoid a copy when building List from stack

ftence:
When `Stack.popn()` is called, it copies the requested elements in the stack array in a new array. In `TOLIST`, this new array is wrapped in a `List` using `Arrays.asList`, which does not copy the data. However when using either `addAll` or the `ArrayList` constructor, it calls `toArray()` on the wrapped array which does a `clone`. 

This PR avoids this last unneeded copy by assigning directly the array produced by `popn()` to the internal array of the `ArrayList`. It also uses `System.arraycopy` which is much faster.

Perf gains expressed as % of time taken for a `->LIST` using the PR code wrt the pre-PR code ie _100 * time_pr / time_pre_pr_.
| Number of elements  | % time taken |
| ------------- | ------------- |
| 1 | 60%  |
| 10  | 65%  |
| 100 | 45%  |
| 1000  | 14%  |
| 10000 | 11%  |
| 10000 | 10%  |

This PR makes `STACKTOLIST` work on a stack which has reached the depth limit.

### DIFF
--- a/warpscript/build.gradle
+++ b/warpscript/build.gradle
@@ -46,6 +46,7 @@ jar {
     include 'io/warp10/WarpURLDecoder.class'
     include 'io/warp10/DoubleUtils.class'
     include 'io/warp10/FloatUtils.class'
+    include 'io/warp10/ListWrapper.class'
     include 'io/warp10/Revision.class'
     include 'io/warp10/ThrowableUtils.class'
     include 'io/warp10/continuum/AuthenticationPlugin.class'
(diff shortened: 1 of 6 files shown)


warpscript/build.gradle
```diff
@@ -46,6 +46,7 @@ jar {
     include 'io/warp10/WarpURLDecoder.class'
     include 'io/warp10/DoubleUtils.class'
     include 'io/warp10/FloatUtils.class'
+    include 'io/warp10/ListWrapper.class'
```
hbs, line 49: Does this include the ArrayItr class too?

ftence: Since the inner class `ArrayItr` is public, it is in the ListWrapper.class file.

PR #721, opened 2020-04-10
UNLIST to throw when not given a List/Array. …

ftence:
Add UNSET which does the same as UNLIST on sets

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/UNSET.java
+++ b/warp10/src/main/java/io/warp10/script/functions/UNSET.java
@@ -24,8 +24,7 @@ import io.warp10.script.WarpScriptStackFunction;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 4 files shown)


warp10/src/main/java/io/warp10/script/functions/UNSET.java
```diff
@@ -0,0 +1,53 @@
...
+import io.warp10.script.NamedWarpScriptFunction;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.script.WarpScriptStackFunction;
+
+import java.util.Set;
+
+/**
+ * Push onto the stack all elements of the set on top
+ * on top of a MARK
```
hbs, line 28: extraneous "on top"

PR #731, opened 2020-04-14
#Fixes 716 - Fixes edge cases of boundary fetch

hbs:


warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -572,16 +556,35 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+                  if (lastCount + dpcount < count) {
+                    GTSHelper.setValue(base, ts, location, elevation, value, false);                
+                  } else {
+                    // We are done, exit
+                    break;
+                  }
+                } else {
+                  // Consider all points returned by the underlying system
+                  GTSHelper.setValue(base, ts, location, elevation, value, false);                
+                }
```
ftence, line 584: Could be simplified by keeping `GTSHelper.setValue(base, ts, location, elevation, value, false);` here and adding 
```
if (countOnly && lastCount + dpcount >= count) {
  break;
}
```
earlier in the loop or even in the while condition.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -610,27 +613,20 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+            // It is therefore necessary to keep track of how many datapoints were already fetched and
+            // shrink the GTS so we do not return too many.
+            //
             
-            if (count >= 0 && lastCount + GTSHelper.nvalues(gts) - boundary > count) {
-              // We would add too many datapoints, we will shrink the GTS.
-              // As it it sorted in reverse order of the ticks (since the datapoints are organized
-              // this way in HBase), we just need to shrink the GTS.
-              gts = GTSHelper.shrinkTo(gts, (int) Math.max(count - lastCount + boundary, 0));
+            if (count >= 0 && 0 == postBoundary && 0 == preBoundary && lastCount + GTSHelper.nvalues(gts) > count) {
```
ftence, line 626: Use `countOnly`.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
hbs: Keeping the conditional just above the call to setValue.

hbs:
PTAL

PR #738, opened 2020-04-17
Mapper, Bucketizer and Reducer Interfaces extends Aggregator for simplicity

ftence:
Will avoid problems like those fixed with #678 in the future.


hbs:
Waving the copyright year for this time....

ftence:
Never!

PR #746, opened 2020-04-22
Simplify constant conditions and constant variables

ftence:
Each modification is explained in the diff.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
+++ b/warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
@@ -198,10 +198,10 @@ public class EgressExecHandler extends AbstractHandler {
       // Extract parameters from the path info and set their value as symbols
       //
       
-      String pathInfo = req.getPathInfo().substring(target.length());
+      String pathInfo = req.getPathInfo();
       
-      if (pathInfo.length() > 0) {
-        pathInfo = pathInfo.substring(1);
+      if (pathInfo != null && pathInfo.length() > Constants.API_ENDPOINT_EXEC.length()) {
+        pathInfo = pathInfo.substring(Constants.API_ENDPOINT_EXEC.length() + 1);
         String[] tokens = pathInfo.split("/");
 
         for (String token: tokens) {
--- a/warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
+++ b/warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
@@ -200,7 +200,7 @@ public class EgressExecHandler extends AbstractHandler {
       
       String pathInfo = req.getPathInfo().substring(target.length());
       
-      if (null != pathInfo && pathInfo.length() > 0) {
+      if (pathInfo.length() > 0) {
         pathInfo = pathInfo.substring(1);
         String[] tokens = pathInfo.split("/");
 
--- a/warp10/src/main/java/io/warp10/hadoop/Warp10InputFormat.java
+++ b/warp10/src/main/java/io/warp10/hadoop/Warp10InputFormat.java
@@ -305,6 +305,7 @@ public class Warp10InputFormat extends InputFormat<Text, BytesWritable> {
     br = new BufferedReader(new FileReader(outfile));
     
     Warp10InputSplit split = new Warp10InputSplit();
+    String lastserver = null;
     int subsplits = 0;
     
     while(true) {
@@ -318,7 +319,7 @@ public class Warp10InputFormat extends InputFormat<Text, BytesWritable> {
       
       // If the server changed or we've reached the maximum split size, flush the current split.
       
-      if (avgsplitcount == subsplits) {
+      if ((null != lastserver && !lastserver.equals(tokens[0])) || avgsplitcount == subsplits) {
         // Add fallback fetchers, shuffle them first
         Collections.shuffle(fallbacks);
         for (String fallback: fallbacks) {
@@ -331,6 +332,7 @@ public class Warp10InputFormat extends InputFormat<Text, BytesWritable> {
       }
       
       subsplits++;
+      lastserver = tokens[0];
 
       split.addEntry(fallbacksonly ? null : tokens[0], tokens[2]);
     }
--- a/warp10/src/main/java/io/warp10/hadoop/Warp10InputFormat.java
+++ b/warp10/src/main/java/io/warp10/hadoop/Warp10InputFormat.java
@@ -305,7 +305,6 @@ public class Warp10InputFormat extends InputFormat<Text, BytesWritable> {
     br = new BufferedReader(new FileReader(outfile));
     
     Warp10InputSplit split = new Warp10InputSplit();
-    String lastserver = null;
     int subsplits = 0;
     
     while(true) {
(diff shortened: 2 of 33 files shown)


warp10/src/main/java/io/warp10/standalone/InMemoryChunkSet.java
```diff
@@ -399,11 +399,6 @@ public int compare(Object[] o1, Object[] o2) {
         continue;
       }
       
-      // Chunk does not intersect the main range, so if we are not fetching a preboundary, ignore it
-      if (boundaryOnly && null == boundary) {
-        continue;
-      }
```
ftence, line 405: This never happen. In the above line
```
      if (null == chunkDecoder) {
        continue;
      }
```
and the only way for `chunkDecoder` to be non-null is 
```
        if (null != this.chunks[chunk] && !(boundaryOnly && null == boundary)) {
          chunkDecoder = this.chunks[chunk].getUnsafeDecoder(false);
        }
```
which means `!(boundaryOnly && null == boundary)` so `boundaryOnly && null == boundary` is always false.

warp10/src/main/java/io/warp10/standalone/StandaloneStoreClient.java
```diff
@@ -387,45 +387,37 @@ public boolean hasNext() {
...
-              
-            stoprow = new byte[startrow.length];
-            bb = ByteBuffer.wrap(stoprow).order(ByteOrder.BIG_ENDIAN);
-            bb.put(Constants.HBASE_RAW_DATA_KEY_PREFIX);
-            bb.putLong(metadatas.get(idx).getClassId());
-            bb.putLong(metadatas.get(idx).getLabelsId());                            
-            bb.putLong(Long.MAX_VALUE - then);
+          if (idx >= metadatas.size()) {
+            return false;
           }
```
ftence, line 405: `startrow` is always set to `null` or else `return true;` in the first part.
So the test `null == startrow` is always true and then `startrow` is reassigned to a new value.

warp10/src/main/java/io/warp10/script/filter/FilterByLabels.java
```diff
@@ -136,12 +136,12 @@ public FilterByLabels(String name, Map<String,String> selectors, boolean checkLa
           }
 
           // Now check the label or attribute value
-          if (hasLabel && checkLabels) {
```
ftence, line 139: if `hasLabel`, then `checkLabels` because `hasLabel` can be true only if `checkLabels`

warp10/src/main/java/io/warp10/script/filter/FilterByLabels.java
```diff
@@ -136,12 +136,12 @@ public FilterByLabels(String name, Map<String,String> selectors, boolean checkLa
...
+          if (hasLabel) {
             if (!matcher.reset(labels.get(label)).matches()) {
               matched = false;
               break;
             }            
-          } else if (checkAttributes) {
```
ftence, line 144: Because
```
          if (!matched) {	          if (!matched) {
            break;	            break;
          }
```
We know that it matched and it was not on labels, so it must be on attributes.

warp10/src/main/java/io/warp10/hadoop/Warp10InputFormat.java
```diff
@@ -319,7 +318,7 @@ public Warp10InputFormat() {
       
       // If the server changed or we've reached the maximum split size, flush the current split.
       
-      if (null != lastserver && !lastserver.equals(tokens[0]) || avgsplitcount == subsplits) {
```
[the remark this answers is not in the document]
hbs: lastserver should be set to tokens[0] in the loop instead of removing the check.

warp10/src/main/java/io/warp10/script/MacroHelper.java
```diff
@@ -56,7 +56,7 @@ public String toString() {
     
     @Override
     public String snapshot() {
-      if (macro instanceof Snapshotable) {
+      if (macro != null) {
```
hbs, line 59: make that null != macro

warp10/src/main/java/io/warp10/script/filter/FilterByLabels.java
hbs: Maybe, but should we check the attributes? This is what checkAttributes determines, so it is not always true.

warp10/src/main/java/io/warp10/script/functions/SNAPSHOT.java
```diff
@@ -284,7 +284,7 @@ public static void addElement(SNAPSHOT snapshot, StringBuilder sb, Object o, boo
           sb.append(WarpScriptLib.LIST_START);
           sb.append(" ");
           for (Object oo : (Vector) o) {
-            addElement(snapshot, sb, oo, readable);
```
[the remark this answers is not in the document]
hbs: Agreed, but keeping readable in the calls makes understanding the code simpler when browsing it without an IDE which makes the parameter name explicit.

warp10/src/main/java/io/warp10/standalone/StandalonePlasmaHandler.java
```diff
@@ -248,7 +248,7 @@ public void onWebSocketMessage(Session session, String message) throws Exception
             this.handler.unsubscribe(session, metadatas);
           }
         } finally {
-          if (iter instanceof MetadataIterator) {
+          if (iter != null) {
```
hbs, line 251: null != iter

warp10/src/main/java/io/warp10/standalone/StandaloneStoreClient.java
```diff
@@ -387,45 +387,37 @@ public boolean hasNext() {
...
               }
             }
-          } else {
-            startrow = null;
           }
           
-          // We need to reseek if startrow is null (it indicates we need to skip to the next GTS)
-          if (null == startrow) {
-            idx++;
+          // We need to reseek because we need to skip to the next GTS
```
hbs, line 400: Leave a comment indicating this is determined by startrow having been set to null

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
```diff
@@ -200,7 +200,7 @@ public void handle(String target, Request baseRequest, HttpServletRequest req, H
       
       String pathInfo = req.getPathInfo().substring(target.length());
       
-      if (null != pathInfo && pathInfo.length() > 0) {
```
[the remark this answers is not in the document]
ftence: So it still need to be fixed, as `req.getPathInfo().substring(target.length())` could throw a NPE.
Fixing that.

warp10/src/main/java/io/warp10/script/filter/FilterByLabels.java
ftence: In the else we have `!hasLabel && matched` so `matched != hasLabel`. This means we were not in the 
```
else {
            matched = hasLabel;
          }
```
branch but in the 
```
          if (!hasLabel && checkAttributes) {
            if (!attributes.containsKey(label)) {
              matched = false;
            }
          }
```
branch.

Which means `checkAttributes == true`.

hbs: Then add it as a comment to make the understanding easier once the clause is removed.

PR #749, opened 2020-04-30
Initial commit of ->VARINT and VARINT->

hbs:


warp10/src/main/java/io/warp10/script/functions/TOVARINT.java
```diff
@@ -0,0 +1,68 @@
...
+      
+      try {
+        for (int i = 0; i < longs.size(); i++) {
+          if (!(longs.get(i) instanceof Long)) {
+            throw new WarpScriptException(getName() + " operates on a LIST of LONGs.");
+          }
+          baos.write(Varint.encodeUnsignedLong(((Long) longs.get(i)).longValue()));
+        }
+        stack.push(baos.toByteArray());
+      } catch (Exception e) {
```
ftence, line 61: Which exception do you intend to catch?
`baos.write` can throw NPE or OutOfBound but it cannot happen because it is not given Null and baos is of sufficient length.
Also you will catch the WarpScriptException which is unnecessary.

warp10/src/main/java/io/warp10/script/functions/VARINTTO.java
```diff
@@ -0,0 +1,82 @@
...
+    
+    List<Object> values = new ArrayList<Object>();
+
+    try {
+      while(bb.hasRemaining() && count > 0) {
+        long value = Varint.decodeUnsignedLong(bb);
+        values.add(value);
+        count--;
+      }
+    } catch (Exception e) {
```
ftence, line 67: I think the only Exception you can catch here is IllegalArgument, you should change to that.

warp10/src/main/java/io/warp10/script/functions/VARINTTO.java
```diff
@@ -0,0 +1,82 @@
...
+      throw new WarpScriptException(getName() + " error while decoding values.", e);
+    }
+    
+    stack.push(values);
+    
+    //
+    // If we were count based, output the number of bytes we consumed
+    //
+    if (countbased) {
+      stack.push((long) (data.length - (int) bb.remaining()));
```
ftence, line 77: Unnecessary cast to `int`.

warp10/src/main/java/io/warp10/script/functions/VARINTTO.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.List;
+
+import io.warp10.continuum.gts.Varint;
+import io.warp10.script.NamedWarpScriptFunction;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.script.WarpScriptStackFunction;
+
+/**
+ * Converts a byte array containing varints into a list of varints
```
ftence, line 30: typo: a list of LONGs.

warp10/src/main/java/io/warp10/script/functions/TOVARINT.java
hbs: ByteArrayOutputStream.write throws IOException, so catching this.

warp10/src/main/java/io/warp10/script/functions/TOVARINT.java
```diff
@@ -58,8 +59,8 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
           baos.write(Varint.encodeUnsignedLong(((Long) longs.get(i)).longValue()));
         }
         stack.push(baos.toByteArray());
-      } catch (Exception e) {
-        throw new WarpScriptException(getName() + " error while decoding values.", e);
+      } catch (IOException ioe) {
+        throw new WarpScriptException(getName() + " error while decoding values.", ioe);
```
ftence, line 63: typo: decoding->encoding

PR #753, opened 2020-05-05
Make MIN and MAX return LONG for Counters

ftence:
Used test for Double/BigDecimal and switched the if/else branches.


warp10/src/main/java/io/warp10/script/functions/math/MAX.java
```diff
@@ -47,16 +48,16 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
         }
 
         if (null == min) {
-          if (element instanceof Long || element instanceof Integer) {
-            min = ((Number) element).longValue();
-          } else {
+          if (element instanceof Double || element instanceof BigDecimal) {
```
hbs, line 51: How about renaming min in max as part of this PR?

ftence:
No performance difference between `MIN`/`MAX` and `NumericBinaryFunction` according to JHM benchmarks.

PR #758, opened 2020-05-12
Fixed edge cases when count is 0

hbs:


warp10/src/main/java/io/warp10/standalone/StandaloneStoreClient.java
```diff
@@ -36,6 +36,8 @@
 import org.iq80.leveldb.WriteBatch;
 import org.iq80.leveldb.WriteOptions;
 
+import com.geoxp.oss.jarjar.org.apache.commons.codec.binary.Hex;
+
```
ftence, line 40: Unused.

PR #760, opened 2020-05-13
Removed calls to toCharArray

hbs:


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -3013,7 +3013,7 @@ public static Object parseValue(String valuestr) throws ParseException {
       } else if ('b' == firstChar && valuestr.startsWith("b64:")) {
         value = Base64.decodeBase64(valuestr.substring(4));
       } else if ('h' == firstChar && valuestr.startsWith("hex:")) {
-        value = Hex.decodeHex(valuestr.substring(4).toCharArray());
+        value = Hex.decode(valuestr.substring(4));
```
ftence, line 3016: The behavior of this method differs from the previous one for invalid inputs:
- The exception thrown when given a hex string with an odd number of letter is much less clear: `java.lang.StringIndexOutOfBoundsException: String index out of range: 3`
- This method does not thrown any exception when given an invalid hex string, for instance `1Z` returns `new byte[]{16}`

PR #761, opened 2020-05-13
Merge most Math functions and make them applicable on lists

ftence:
Extensively tested with JMH and perfs are the same for already existing signatures.


warp10/src/main/java/io/warp10/script/ListRecursiveStackFunction.java
```diff
@@ -0,0 +1,101 @@
...
+ * This is similar to GTSStackFunction for GTSs but it is a tad faster if the function uses some parameters as this
+ * implementation does not use a Map for parameters.
+ * This is similar to ElementOrListStackFunction but it is faster has it checks for List after the other cases which
+ * it optimal in almost all the cases where the number of elements is higher than the nesting depth of the lists.
+ */
+public abstract class ListRecursiveStackFunction extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+
+  /**
+   * Interface defining the function to be generated which is applied to each element.
+   * This method should check the type of the given element and call unhandled on it if the element is of
```
hbs, line 36: The handling of nested lists should be done by ElementStackFunction itself not left to the willingness of the applyOnElement implementation.

applyOnElement should be declared abstract, but a final apply method should be added which checks if the passed object is a list, and if so it will build the resulting list by calling apply recursively until an element is not a list in which case applyOnElement will be called.

ftence: After some benchmarks, this solution seems indeed better at the cost of little to none decrease in performance.

warp10/src/main/java/io/warp10/script/ListRecursiveStackFunction.java
```diff
@@ -0,0 +1,105 @@
...
+//   limitations under the License.
+//
+
+package io.warp10.script;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Abstract class to inherit when a function can be called on a single element or a list of elements or list of list of elements, etc.
```
hbs, line 23: list of lists or nested lists in a more general way?

warp10/src/main/java/io/warp10/script/functions/NumericalBinaryFunction.java
```diff
@@ -0,0 +1,151 @@
...
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.DoubleBinaryOperator;
+import java.util.function.LongBinaryOperator;
+
+/**
+ * Apply a double or long binary operator to two values.
+ * If the only the long operator is defined, all numbers are converted to long.
```
hbs, line 32: extraneous 'the'

warp10/src/main/java/io/warp10/script/functions/NumericalUnaryFunction.java
```diff
@@ -0,0 +1,66 @@
...
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptStack;
+
+import java.math.BigDecimal;
+import java.util.function.DoubleUnaryOperator;
+import java.util.function.LongUnaryOperator;
+
+/**
+ * Apply a double or long unary operator to one value, a list of values or nested lists of values.
+ * If the only the long operator is defined, all numbers are converted to long.
```
hbs, line 29: extraneous 'the'

warp10/src/main/java/io/warp10/script/functions/NumericalBinaryFunction.java
```diff
@@ -0,0 +1,151 @@
...
+ * If the only the double operator is defined, all numbers are converted to double.
+ * If the both long and double operator are defined, all numbers are converted long until a Double or BigDecimal is found,
+ * then they are all converted to double.
+ *
+ * The operator can also be applied to a list of values and a single operand, on top.
+ * The result is a copy of the list with values being the result of the operator applied on the initial value and the single operand.
+ *
+ * If the function is given a list on top, its behavior depends on applyOnSingleList:
+ * - if True the result is op(...op(op(op(v[0], v[1]), v[2]), v[3]), ... v[n])
```
hbs, line 41: Use 'true' and 'false', not the pythonic 'True' and 'False'

PR #771, opened 2020-06-03
Update gradle to 6.4.1

ftence:


build.gradle
```diff
@@ -212,7 +212,6 @@ subprojects {
         repositories {
             repositories.mavenDeployer {
                 configuration = configurations.deployerJars
-                uniqueVersion = false
```
hbs, line 215: What was the purpose of this flag?

ftence: `uniqueVersion` was used to make the maven plugin to append a unique identifier to snapshots.
It does not exist anymore, thus the removal.
I tagged my repo with a tag ending in SNAPSHOT and the published jar for maven ends with SNAPSHOT.jar, so all seems OK.

PR #773, opened 2020-06-06
Initial commit of GEO.BUFFER

hbs:


warp10/src/main/java/io/warp10/script/functions/GeoJSON.java
```diff
@@ -67,6 +72,23 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+    if (null != buffer) {
+      // Clear the buffer
+      stack.setAttribute(GEOBUFFER.ATTR_GEOBUFFER, null);
+      // Apply the buffer operation
+      BufferOp bop = new BufferOp(geometry, (BufferParameters) buffer.get(GEOBUFFER.KEY_PARAMS));
+      geometry = bop.getResultGeometry(((Double) buffer.get(GEOBUFFER.KEY_DIST)).doubleValue());
+      if (Boolean.TRUE.equals(buffer.get(GEOBUFFER.KEY_WKT))) {
+        stack.push(geometry.toText());
+        return stack;
+      }
```
ftence, line 89: If find this kind of hacky, I'd prefer to have a dedicated function to apply the buffer and output a WKT.

Other solution: I was planning to modify `->GEOJSON` and introduce `->WKT` as well as `->WKB` to convert between formats, maybe it would be a good place to add this logic?

hbs: Hackyness is a way of life :-)

How about we change the behavior of GEO.BUFFER so if the 'wkt', 'wkb' or 'geojson' key is present in its input parameters then the buffer is added to the value associated with the key and output instead of configuring the buffer for the next GEO.WKT/GEO.WKB/GEO.GEOJSON call?

ftence: Good idea!

hbs:
simplifyFactor is does not exist in the vividsolutions JTS. Exposing it would require a switch to location tech which has too many impacts at this stage. Will need to be reconsidered later.

hbs:
PTAL

PR #774, opened 2020-06-06
Revamp gradle files and build script for cleaner build and publish

ftence:


warp10/build.gradle
```diff
@@ -43,171 +42,147 @@ dependencies {
...
+        configurations.compileClasspath.collect {
             it.isDirectory() ? it : zipTree(it).matching {
                 exclude { it.toString().contains('/servlet-api') }
-                exclude { it.toString().contains('jmh') }
                 exclude { it.toString().contains('junit') }
-                exclude { it.toString().contains('parquet') }
             }
         }
     } {
         exclude 'META-INF/*.DSA'
```
hbs, line 151: META-INF/*.RSA signatures may appear/disappear depending on dependencies. It is important that those files be removed otherwise the JVM may fail to validate repackaged classes if they initially had a signature.

hbs:
Please resolve conflict

ftence:
To better understand the changes induced by this PR, here are several diffs made with current master (https://github.com/senx/warp10-platform/tree/3d612363f6a90e01ab4937213fe9f8481ae7251b):

Diff of the files and directories in the generated warp10-*.jar, `+` means added in the jar from this PR, `-` means removed from the jar of this PR: https://gist.github.com/ftence/147be14922eadd1d6c6356b34862f431

Diff of the "not found" classes with `jdeps` introduced by this PR. That is `jdeps -v` has been called on both jars and only the "not found" generated by the jar from this PR which were **not present** in the jar created with master are reported: https://gist.github.com/ftence/d4e676f22b1e9a99d3fc579fe07db47d

Impacted classes listed by this `jdeps` diff are:
- `org.apache.hadoop.hbase.coprocessor.example.BulkDeleteEndpoint` which is not directly used, only `BulkDeleteProtos` is used to interact with this endpoint.
- `org.apache.hadoop.hbase.coprocessor.example.RowCountEndpoint` is unused and not a dependency of another class
- `org.apache.hadoop.hbase.coprocessor.example.ZooKeeperScanPolicyObserver` is unused and not a dependency of another class
- `org.apache.hadoop.hbase.mapreduce.IndexBuilder` is unused and not a dependency of another class
- `org.apache.hadoop.hbase.mapreduce.SampleUploader` is unused and not a dependency of another class
- `org.apache.hadoop.hbase.thrift2.DemoClient` is unused and not a dependency of another class
- `org.apache.hadoop.hbase.thrift.DemoClient` is unused and not a dependency of another class
- `org.apache.hadoop.hbase.thrift.HttpDoAsClient` is unused and not a dependency of another class

This `jdeps` diff does not list implicit dependencies (injection for instance) thus the files diff can give more insight:
- `com.lmax` and several `com.sun.*` and packages have been removed. As this does not raise alarm with jdeps, it should be OK.
- Some `com.vividsolution.jts` classes have been removed because of the jts/jts-core mess. Again, no jdeps alarm.
- Probably useless files in the root of the jar and in META-INF
- Differences in `org.apache.commons.beanutils`, because 1.9.0 was forced. jdeps is OK.
- `org.apache.commons.math` is removed, jdeps OK.
- A LOT of classes from `org.apache.hadoop.*`, because hbase-server is not a dependency anymore as it was pulled by hbase-examples. Again jdeps would have noticed something.
- Multiple `org.apache.jasper` classes have been removed, this is a side effect of removing the `hbase-server` dependency which was the one one needing `jsp-2.1`. So that's OK.
- `org.fusesource.leveldbjni.All` has been removed, not sure why, but this class is empty.
- Also a side effect of removing `hbase-server`, `org.jamon` is removed.
- Same with `jetty-sslengine` which removes `org.mortbay.jetty.security` classes.

ftence:
I forgot to mention that all the modifications on the dependencies aim at avoiding duplicates when packing Warp 10.

Without this PR, there are many duplicate .class files in the jar, without any control on which version is used. It seems that the order in the jar is consistent, ie if there is a version _a_ and _b_, either _a_'s classes are always the first or _b_'s are. It seems also that if the pack is done several times, the order is consistent. However this is dangerous are there are no guarantees.

This PR drastically reduce the number of duplicates. The only duplicate .class files are because of jline 2.13, see the comment in warp10/build.gradle.

steven-gueguen (approved):

PR #775, opened 2020-06-13
Initial commit of HIDE/SHOW

hbs:


warp10/src/main/java/io/warp10/script/WarpScriptStack.java
```diff
@@ -833,5 +833,16 @@ private void ensureCapacity(int n) {
    */
   public void restore() throws WarpScriptException;
   
+  /**
+   * Hide some stack levels
```
ftence, line 837: You should clarify that the hidden levels are the deepest.

PR #776, opened 2020-06-16
Allow headers setting in URLFETCH

ftence:


warp10/src/main/java/io/warp10/script/ext/urlfetch/URLFETCH.java
```diff
@@ -69,6 +69,12 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
     Object o = stack.pop();
 
+    HashMap<String, String> properties = new HashMap<String, String>();
+    if (o instanceof Map) {
```
hbs, line 73: Instead of casting it would probably be better to iterate on the entries and call put(String.valueOf(k),String.valueOf(v));

PR #782, opened 2020-07-03
Uniformize the configuration templates

ftence:


etc/conf.templates/distributed/00-secrets.conf.template
```diff
@@ -45,22 +45,22 @@ warp.hash.token = hex:hhhhhh...
 warp.hash.app = hex:hhhhhh...
 
 //
-// 256 bits key for protecting tokens, format is hex:hhhh...hhhh
+// Key to use for protecting tokens (128/192/256 bits in hex or OSS reference)
```
hbs, line 48: What do you call an OSS reference? The OSSKeyStore has the added ability to decode a key wrapped with the master key retrieved from OSS but there is no possibility to fetch a key directly from OSS.

ftence: Yes, you're right, I took this from other configuration keys but it is clearly not the good terminology.
7c89ce4 should fix that.

PR #787, opened 2020-07-08
Use LinkedHashMap for the generation of the Accelerator report

ftence:


hbs:
Can you describe the purpose?

ftence:
Sure. There are two reasons:
- TOMAP generates a LinkedHashMap, so to be consistent with Map creation in WarpScript, this function should also push a LinkedHashMap.
- Report was not easy to read in JSON with primary information (status and accelerated) mixed with secondary one:
```
[
  {
    "cache": true,
    "defaults.read": [
      "cache",
      "persist"
    ],
    "chunkcount": 1,
    "defaults.write": [
      "cache",
      "persist"
    ],
    "persist": true,
    "accelerated": false,
    "defaults.delete": [
      "cache",
      "persist"
    ],
    "chunkspan": 9223372036854775807,
    "status": true
  }
]
```

With this PR, the JSON looks like this:
```
[
  {
    "status": true,
    "cache": true,
    "persist": true,
    "accelerated": false,
    "chunkcount": 1,
    "chunkspan": 9223372036854775807,
    "defaults.read": [
      "cache",
      "persist"
    ],
    "defaults.write": [
      "cache",
      "persist"
    ],
    "defaults.delete": [
      "cache",
      "persist"
    ]
  }
]
```

PR #789, opened 2020-07-09
Add optional macro or mapper parameter for DEDUP to merge duplicate ticks

ftence:
This optional parameter can apply more complex strategies for merging duplicate ticks than just keeping the last tick.

The implementation is based on GTSHelper.map but does not expose this parameter in MAP. This is because deduping will behave strangely with step>1, pre>0, post>0 or when using outputTicks. This is why it was included in DEDUP and not MAP.

Example:
```
NEWGTS
0 NaN NaN NaN 0 ADDVALUE
1 NaN NaN NaN 1 ADDVALUE
1 NaN NaN NaN 2 ADDVALUE
1 NaN NaN NaN 3 ADDVALUE
2 NaN NaN NaN 4 ADDVALUE
3 NaN NaN NaN 5 ADDVALUE

mapper.median DEDUP
```
Gives
```
[[{"c":"","l":{},"a":{},"la":0,"v":[[0,0.0],[1,2.0],[2,4.0],[3,5.0]]}]]
```

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/DEDUP.java
+++ b/warp10/src/main/java/io/warp10/script/functions/DEDUP.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018  SenX S.A.S.
+//   Copyright 2018-2020  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


warp10/src/main/java/io/warp10/script/functions/DEDUP.java
```diff
@@ -18,24 +18,25 @@
...
  * Remove duplicates from GTS instances.
  * <p>
- * DEDUP expects no other parameter on the stack than the GTS instances
+ * DEDUP expects an optional mapper or macro (as MAP) and GTS instances
  */
-public class DEDUP extends ElementOrListStackFunction {
+public class DEDUP extends ListRecursiveStackFunction {
 
-  private final ElementStackFunction dedup = new ElementStackFunction() {
+  private final ElementStackFunction simpleDedup = new ElementStackFunction() {
```
hbs, line 33: Put it in upper case since it is a constant

PR #795, opened 2020-07-22
Standalone Directory index by label feature

pi-r-p:
Use case : several millions of GTS with different classnames, same .app, several labels by GTS. 

User wants to find `.*{serialNumber=X}`. User only know the serialNumber, not the other labels.

This feature allows to build an index of one or several labels (serialNumber in the example).
It speeds up FIND, FETCH, DELETE operations.

### DIFF
--- a/etc/conf.templates/distributed/00-warp.conf.template
+++ b/etc/conf.templates/distributed/00-warp.conf.template
@@ -139,13 +139,6 @@ warp.components =
 #warp.labels.maxsize = 2048
 
 //
-// Default priority order for matching labels when doing a FIND/FETCH.
-// Comma separated list of label names.
-// Defaults to .producer,.app,.owner
-//
-#warp.labels.priority =
-  
-//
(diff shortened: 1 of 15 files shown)


trepca:
woohoo! 🍾

hbs:
This is too specific to be included in Warp 10 as is. I suggest this PR be closed and an issue be open to add a generic modular indexing mechanism which would allow to register classes which would handle specific indices. The behavior added by this PR can then be implemented by such a class.

pi-r-p:
Great idea! Something that work with standalone, distributed, sharded version.

Imho, a generic module should implement at least these functions:
-buildindex(metadata) (called when decoding each metadata)
-find (directoryrequest) : it can return metadata found + a new directory request to search for remaining series that where not found by the module.
-unregister(metadata)
-register(metadata) : It could be the same as buildindex

it could be great (but risky) that the module also access the current "this.metadata"  index.

hbs:
Closing this PR, foreseeing a design doc of more generic additional indices.

PR #797, opened 2020-07-22
Added pixel extraction from PImage instances

hbs:


warp10/src/main/java/io/warp10/script/processing/image/PupdatePixels.java
```diff
@@ -26,6 +26,7 @@
 import java.util.List;
```
pi-r-p, line 26: ArraylList useless import

PR #798, opened 2020-07-23
Allow extensions or plugins to register custom serializers

randomboolean:


ftence:
This functionality already exists through the `JsonUtils.addEncoder` function. It works roughly the same way `SNAPSHOT.addEncoder` works so this is homogeneous.

randomboolean:
> This functionality already exists through the `JsonUtils.addEncoder` function. It works roughly the same way `SNAPSHOT.addEncoder` works so this is homogeneous.

This is only useful for transforming an object into a string. It does not allows reusing other registered serializers in a nested way, nor using the JsonGenerator object.

PR #802, opened 2020-07-29
Modify custom JSON serialization

ftence:
Current custom JSON serialization is not working and could be more practical.

#798 proposed to expose Jackson API but it is preferable not to in case we want to change the JSON serialization/deserialization library.

This PR changes the current API to register custom serialization, making it simpler to use while not exposing Jackson.
The idea is to allow the registration of "transformers" whose role is to convert un-serializable objects to serializable ones. Typically they will create a Map corresponding to the fields of the Object.

Example of use:
```
JsonUtils.addTransformer(new JsonUtils.JsonTransformer() {
      @Override
      public TransformationResult transform(Object o) {
        if (o instanceof CustomObject) {
          HashMap<String, Object> coRepresentation = new HashMap<String, Object>();
          coRepresentation.put("CustomObject", ((CustomObject) o).children);
          return new TransformationResult(true, coRepresentation);
        } else {
          return new TransformationResult(false, null);
        }
      }
    });
```

### DIFF
--- a/warp10/src/main/java/io/warp10/json/JsonUtils.java
+++ b/warp10/src/main/java/io/warp10/json/JsonUtils.java
@@ -107,9 +107,27 @@ public class JsonUtils {
 
   public interface JsonTransformer {
     class TransformationResult {
-      public final boolean transformed;
-      public final Object result;
-      public final boolean raw;
+
+      /**
+       * Whether the transformer managed to transform the given object or not.
+       * If false, result and raw fields are ignored.
+       */
+      private final boolean transformed;
+
+      /**
+       * The result of the transformation.
+       * If raw, the result must be a String and will be added as is in the resulting JSON.
+       * If not raw, the result should be an Object which can be JSONified. Typically Maps, Lists, String, Numbers, Booleans
+       * or even other Objects which are known to have serializer (GTS, Bytes, ...) or handled by other transformers.
+       */
+      private final Object result;
+
+      /**
+       * Whether the result is to be added as is in the resulting JSON.
+       * If true, result must be a String.
+       * If false, result can be any Object which will be JSONified.
+       */
+      private final boolean raw;
 
       public TransformationResult(boolean transformed, Object result, boolean raw) {
         this.transformed = transformed;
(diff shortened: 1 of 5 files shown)


warp10/src/main/java/io/warp10/json/JsonUtils.java
```diff
@@ -57,33 +52,35 @@ public void serialize(Object value, JsonGenerator gen, SerializerProvider serial
...
+        JsonTransformer.TransformationResult transfRes = null;
+        for (JsonTransformer transformer: transformers) {
+          transfRes = transformer.transform(value);
+          if (null != transfRes && transfRes.transformed) {
             break;
           }
         }
-        if (encoded) {
-          gen.writeRaw(sb.toString());
+        if (null != transfRes && transfRes.transformed) {
```
hbs, line 82: Shouldn't we leave the possibility to return JSON content too? Because the transformer may use an existing implementation of a serializer which will produce JSON. The validity of the returned JSON would be the responsability of the called entity as it was in the previous implementation.

ftence:
I added the fonctionality you described @hbs. Here how it's used:
```
JsonUtils.addTransformer(new JsonUtils.JsonTransformer() {
  @Override
  public TransformationResult transform(Object o) {
    if (o instanceof CustomObject) {
      HashMap<String, Object> coRepresentation = new HashMap<String, Object>();
      coRepresentation.put("CustomObject", ((CustomObject) o).children);
      try {
        String rep = JsonUtils.objectToJson(coRepresentation);
        return new TransformationResult(true, rep, true);
      } catch (IOException e) {
        return new TransformationResult(false, null, false);
      }
    } else {
      return new TransformationResult(false, null, false);
    }
  }
});
```

warp10/src/main/java/io/warp10/json/JsonUtils.java
```diff
@@ -95,20 +96,41 @@ public void serialize(Object value, JsonGenerator gen, SerializerProvider serial
...
   // ObjectMapper instances are thread-safe, so we can safely use a single static instance.
   //
-  private static final ObjectMapper strictMapper;
-  private static final ObjectMapper looseMapper;
+  private static final ObjectMapper STRICT_MAPPER;
+  private static final ObjectMapper LOOSE_MAPPER;
+
+  public interface JsonTransformer {
+    class TransformationResult {
+      public final boolean transformed;
```
hbs, line 110: Add comments describing the meaning of those fields

hbs: Also add getter method so the fields can be set private

PR #803, opened 2020-07-29
Modified code so list of Metadatas is not deep copied when cloning Fe…

hbs:
…tchRequest

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/store/OptimizedSlicedRowFilterGTSDecoderIterator.java
+++ b/warp10/src/main/java/io/warp10/continuum/store/OptimizedSlicedRowFilterGTSDecoderIterator.java
@@ -50,7 +50,11 @@ public class OptimizedSlicedRowFilterGTSDecoderIterator extends GTSDecoderIterat
   public OptimizedSlicedRowFilterGTSDecoderIterator(FetchRequest req, Connection conn, TableName tableName, byte[] colfam, KeyStore keystore, boolean useBlockCache) {
     
     // We allocate a new one so the parameters are preserved, the metadatas will vary anyway
+    // Remove Metadatas from FetchRequest otherwise new FetchRequest(req) will do a deep copy
+    List<Metadata> lm = req.getMetadatas();          
+    req.unsetMetadatas();
     this.request = new FetchRequest(req);
+    req.setMetadatas(lm);
     this.now = req.getNow();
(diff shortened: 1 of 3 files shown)


ftence:
The token is also deep copied, should it be avoided too? In that case, a utility function clearly stating it is shallow copy would be nicer.

hbs:
The token is not a heavy object to copy, the list of metadata can contain hundreds of thousands of instances, making it a real hit in terms of performance and memory consumption

warp10/src/main/java/io/warp10/continuum/store/ParallelGTSDecoderIteratorWrapper.java
```diff
@@ -301,11 +301,21 @@ public ParallelGTSDecoderIteratorWrapper(FetchRequest req, boolean optimized, Ke
         GTSDecoderIterator iterator = null;
         
         if (optimized) {
+          // Remove Metadatas from FetchRequest otherwise new FetchRequest(req) will do a deep copy
```
ftence, line 304: Factorize code outside conditional.

warp10/src/main/java/io/warp10/continuum/store/ParallelGTSDecoderIteratorWrapper.java
```diff
@@ -320,11 +330,21 @@ public ParallelGTSDecoderIteratorWrapper(FetchRequest req, boolean optimized, Ke
       GTSDecoderIterator iterator = null;
       
       if (optimized) {
+        // Remove Metadatas from FetchRequest otherwise new FetchRequest(req) will do a deep copy
```
ftence, line 333: Also factorize here.

warp10/src/main/java/io/warp10/continuum/store/OptimizedSlicedRowFilterGTSDecoderIterator.java
```diff
@@ -193,7 +197,11 @@ public boolean hasNext() {
       if (null != iterator) {
         try { iterator.close(); } catch (Exception e) {}
       }
+      // Remove Metadatas from FetchRequest otherwise new FetchRequest(req) will do a deep copy
```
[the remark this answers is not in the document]
hbs: Indeed, Metadatas is unset

hbs:
PTAL

PR #810, opened 2020-07-31
Add optimization to ASREGS

ftence:
Replaces STORE on lists containing only register numbers by POPRs

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/ASREGS.java
+++ b/warp10/src/main/java/io/warp10/script/functions/ASREGS.java
@@ -16,18 +16,19 @@
 
 package io.warp10.script.functions;
 
-import java.util.ArrayList;
-import java.util.BitSet;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
 import io.warp10.script.NamedWarpScriptFunction;
-import io.warp10.script.WarpScriptStackFunction;
 import io.warp10.script.WarpScriptException;
 import io.warp10.script.WarpScriptLib;
 import io.warp10.script.WarpScriptStack;
 import io.warp10.script.WarpScriptStack.Macro;
+import io.warp10.script.WarpScriptStackFunction;
+
+import java.util.ArrayList;
+import java.util.BitSet;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Modifies a Macro so the LOAD/STORE operations for the given variables are
@@ -156,11 +157,8 @@ public class ASREGS extends NamedWarpScriptFunction implements WarpScriptStackFu
       throw new WarpScriptException(getName() + " was unable to convert variables to registers.");
     }
 
-    Map<String,Integer> varregs = new HashMap<String,Integer>();
-    
-    int regidx = 0;
-    int numregs = 0;
-    
+    Map<String, Integer> varregs = new HashMap<String, Integer>();
+
     for (Object v: vars) {
       
       if (v instanceof Long) {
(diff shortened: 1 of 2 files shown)


warp10/src/main/java/io/warp10/script/functions/ASREGS.java
```diff
@@ -242,16 +242,40 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+              statements.set(i - 1, NOOP); // replace ENDLIST
+              statements.set(i, NOOP); // replace STORE
+
+              // As we must flip the order of the list, we must store the registers first.
+              int[] regInList = new int[listLength];
+              for (int listIndex = 0; listIndex < listLength; listIndex++) {
+                regInList[listIndex] = ((Long) statements.get(idx + 1 + listIndex)).intValue();
+              }
+
+              // Replace register number by POPRs. Be careful, we flip the list order!
```
hbs, line 274: What happens if you initially had 1 2 3 [ 1 2 2 ] STORE? It seems that register 2 will end up with value 2 instead of 3.

ftence: Good question, I didn't think about this case.
Turns out
```
  1 2 3 [ 1 2 2 ] STORE
  2 LOAD
```
Leave `2` on the stack as `STORE` iterates through the array backwards poping and storing. So the behavior is the same with or without `ASREGS`.

However there is an other bug, I'm investigating.

hbs: I think we should consider fixing STORE in the process, 1 2 3 [ 1 2 2 ] STORE should really leave 3 in register 2, at least that is what would be expected.

warp10/src/main/java/io/warp10/script/functions/STORE.java
```diff
@@ -56,19 +56,22 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
     }
     
     if (count > 0) {
-      for (int i = count - 1; i >= 0; i--) {
+      Object[] objectsToStore = stack.popn(count);
```
hbs, line 59: Use get(n) in the loop then a dropn instead of calling popn which will allocate a new array.

warp10/src/main/java/io/warp10/script/functions/ASREGS.java
```diff
@@ -177,22 +175,16 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-        regidx = inuse.nextClearBit(0);
+        int regidx = inuse.nextClearBit(0);
         inuse.set(regidx);
         varregs.put(v.toString(), regidx);
       }
-    }    
+    }
 
-    numregs = regidx + 1;
+    HashMap<Integer, PUSHR> PUSHRs = new HashMap<Integer, PUSHR>();
```
hbs, line 184: Use a static set of maps instead of per call maps.

ftence:
Here is a extract from a JHM report. It shows that the code is around 3 times faster for `[] STORE` constructs:

With PR:
```
"score": 9.96892377328933,
"scoreError": 0.8550817091112984,
"scoreUnit": "ops/us",
"preinvocation": "<% 1  2  <% [  1  2  ]  STORE  %>  ASREGS  %>",
"macro": "<% EVAL  %>",
"postinvocation": "<% CLEAR  %>"
```

Without PR (equivalent):
```
"score": 3.5234077888226287,
"scoreError": 0.21148269769533423,
"scoreUnit": "ops/us",
"preinvocation": "<% 1  2  <% [  1  2  ]  STORE  %>  %>",
"macro": "<% EVAL  %>",
"postinvocation": "<% CLEAR  %>"
```

PR #811, opened 2020-07-31
Make REMOVE work on GTS using point index

ftence:


warp10/src/main/java/io/warp10/script/functions/REMOVE.java
```diff
@@ -67,8 +69,33 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+
+      for (int i = 0; i < n; i++) {
+        if (i == idx) {
+          continue;
+        }
+
+        GTSHelper.setValue(pruned, GTSHelper.tickAtIndex(gts, i), GTSHelper.locationAtIndex(gts, i), GTSHelper.elevationAtIndex(gts, i), GTSHelper.valueAtIndex(gts, i), false);
+      }
+
+      stack.push(pruned);
```
hbs, line 96: Shouldn't REMOVE behave similarly on a GTS as it does on a list or map? i.e. leave on the stack the data point tuple it just removed? With [ 0 NaN NaN NaN NULL ] if the datapoint did not exist?

ftence: You're right, fixing this.

PR #817, opened 2020-08-31
Add Pmask for masking a PImage with another

ftence:


warp10/src/main/java/io/warp10/script/processing/image/Pmask.java
```diff
@@ -0,0 +1,59 @@
...
+    top = stack.pop();
+
+    if (!(top instanceof PImage)) {
+      throw new WarpScriptException(getName() + " expects two PIMAGEs.");
+    }
+
+    PImage img = (PImage) top;
+
+    // Apply mask
+    img.mask(mask);
```
hbs, line 53: You should catch the IllegalArgumentException which can be thrown by PImage#mask and convert it to a WarpScriptException.

PR #819, opened 2020-09-01
Add RUNR to run macros from registers and make ASREGS use it

ftence:


warp10/src/main/java/io/warp10/script/functions/VARS.java
```diff
@@ -43,32 +43,42 @@ public VARS(String name) {
...
 
+  public static List<Object> getVars(Macro macro) throws WarpScriptException {
+    return getVars(macro, false);
+  }
+
   /**
    * Loop over the macro statements, in a recursive manner, extracting all variable names.
    *
    * @param macro The root macro to extract the variable names from.
+   * @param onlyStoreAndPopr Return only variables used for storage.
```
hbs, line 77: What is the typical case when this option would be used?

ftence: As stated in `ASREGS`
```
// If variables are not defined, only get variables and registers used by STORE and POPR
// to avoid global varaibles.
```
If a variable is only used by a LOAD, this must not be converted to a register because it is most certainly a global variable. Converting it to a register will only break the code.

hbs: What I meant was that this case should be described next to the parameter in the javadoc for clarity purposes.

ftence: Done.

PR #835, opened 2020-09-18
Initial commit of mapper.geo.fence, added constants for geo mappers

hbs:


warp10/src/main/java/io/warp10/script/mapper/MapperGeoFence.java
```diff
@@ -0,0 +1,107 @@
...
+    long location = locations[0];
+    long elevation = elevations[0];
+    
+    //
+    // If there is less than one value or if there is no location associated with the
+    // value, return null as the tick value.
+    //
+   
+    if (null == values[0] || GeoTimeSerie.NO_LOCATION == location || !GeoXPLib.isGeoXPPointInGeoXPShape(location, this.shape)) {
+      return new Object[] { tick, location, elevation, false };
```
ftence, line 86: The comment does not match the code.

On top of that, I'm not sure returning `false` for `GeoTimeSerie.NO_LOCATION` is what most users expect. I'd rather return `null` for no location.

warp10/src/main/java/io/warp10/script/mapper/MapperGeoFence.java
```diff
@@ -0,0 +1,107 @@
...
+      stack.push(new MapperGeoFence(getName(), (GeoXPShape) shape));
+      return stack;
+    }
+  }
+
+  private final GeoXPShape shape;
+  
+  /**
+   * Default constructor, the timezone will be UTC
+   */
```
ftence, line 60: Timezone of the GeoXPShape?

ftence (changes requested):

ftence (approved):

PR #839, opened 2020-09-18
Added conditionals so an NPE is not raised when name or labels is null

hbs:


pi-r-p:
What is the use case ? 
GTS constructors do set name, didn't they ?

hbs:
Metadata coming from unwrapped encoders may contain null name or labels. Same goes for possible functions from extensions, there is no guarantee that name and/or labels will be set.

pi-r-p (approved):

ftence (approved):

PR #842, opened 2020-09-21
Refactored HIDE/SHOW so 0 is interpreted as no levels instead of all levels

hbs:
…s not called.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/WarpScriptStack.java
+++ b/warp10/src/main/java/io/warp10/script/WarpScriptStack.java
@@ -834,11 +834,22 @@ public interface WarpScriptStack {
   public void restore() throws WarpScriptException;
   
   /**
+   * Hide all stack levels
+   * @return The number of levels actually hidden
+   */
+  public int hide();
+  
+  /**
    * Hide the deepest 'count' stack levels
    * @param count Number of levels to hide, use 0 to hide all
    * @return The number of levels actually hidden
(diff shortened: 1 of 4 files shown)


warp10/src/main/java/io/warp10/script/WarpScriptStack.java
```diff
@@ -833,13 +833,24 @@ private void ensureCapacity(int n) {
...
   
+  /**
+   * Hide all stack levels
+   * @return The number of levels actually hidden
+   */
+  public int hide();
+  
   /**
    * Hide the deepest 'count' stack levels
    * @param count Number of levels to hide, use 0 to hide all
```
ftence, line 844: Update comment.

warp10/src/main/java/io/warp10/script/WarpScriptStack.java
```diff
@@ -833,13 +833,24 @@ private void ensureCapacity(int n) {
...
   public int hide(int count);  
   
+  /**
+   * Show all previously hidden stack levels.
+   */
+  public void show();
+  
   /**
    * Show some stack levels previously hidden. Up to 'count'
    * levels previously hidden will be made visible. Those
```
ftence, line 856: Ditto a few line after.

ftence (changes requested):

ftence (approved):

PR #846, opened 2020-09-22
Set UTF-8 as default encoding and check it is at the start of Warp 10

ftence:
This should avoid any problem with functions using default encoding used throughout the code:
- new String(byte[])
- new PrintWriter
- new InputStreamReader
- new OutputStreamWriter
- etc.

### DIFF
--- a/warp10/src/main/java/io/warp10/WarpDist.java
+++ b/warp10/src/main/java/io/warp10/WarpDist.java
@@ -38,6 +38,8 @@ import io.warp10.sensision.Sensision;
 import io.warp10.warp.sdk.AbstractWarp10Plugin;
 
 import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


warp10/src/main/java/io/warp10/WarpDist.java
```diff
@@ -120,6 +122,10 @@ public static void main(String[] args) throws Exception {
     System.out.println();
     
     System.setProperty("java.awt.headless", "true");
+
+    if (Charset.defaultCharset() != StandardCharsets.UTF_8) {
```
hbs, line 126: put constant first.

warp10/src/main/java/io/warp10/standalone/Warp.java
```diff
@@ -120,6 +122,10 @@ public static void main(String[] args) throws Exception {
     System.out.println("  Revision " + Revision.REVISION);
     System.out.println();
 
+    if (Charset.defaultCharset() != StandardCharsets.UTF_8) {
```
hbs, line 125: ditto

hbs (changes requested):

hbs:
Default encoding should be changed and checked in io.warp10.WarpConfig, io.warp10.worf.TokenGen as well as they will read the configuration files.

PR #849, opened 2020-09-23
Implement Hyndman's 9 types of percentile computation

ftence:


warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
```diff
@@ -138,65 +150,99 @@ public int compare(Integer idx1, Integer idx2) {
...
+    double pn = p * nonNullLength;
 
     if (!this.interpolate) {
-      if (n >= nonNullLength) {
-        n--;
+      // Type 1 in Hyndman, Rob & Fan, Yanan. (1996). Sample Quantiles in Statistical Packages. The American Statistician.
+      int j = (int) Math.floor(pn);
+      if (pn > j) {
+        j++;
       }
```
randomboolean, line 167: Can't you use just Math.ceil() instead ?

warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
```diff
@@ -138,65 +150,99 @@ public int compare(Integer idx1, Integer idx2) {
...
-
-        long tick = (long) (ticks[indices[m]] + factor * (ticks[indices[n]] - ticks[indices[m]]));
-        double v;
-
-        if (values[indices[m]] instanceof Long) {
-          v = ((Number) values[indices[m]]).longValue() + factor * (((Number) values[indices[n]]).longValue() - ((Number) values[indices[m]]).longValue());
-        } else {
-          v = ((Number) values[indices[m]]).doubleValue() + factor * (((Number) values[indices[n]]).doubleValue() - ((Number) values[indices[m]]).doubleValue());
-        }
+      // Type 8 in Hyndman, Rob & Fan, Yanan. (1996). Sample Quantiles in Statistical Packages. The American Statistician.
```
randomboolean, line 176: why do you choose Type 8 ?

warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
```diff
@@ -57,26 +57,33 @@ public Builder(String name, boolean forbidNulls) {
 
     @Override
     public Object apply(WarpScriptStack stack) throws WarpScriptException {
-      Object value = stack.pop();
+      Object top = stack.pop();
 
-      if (!(value instanceof Number)) {
+      boolean interpolate = false;
```
randomboolean, line 62: The naming of this variable is confusing. The behaviour with interpolate = false is a type of interpolation.

randomboolean:
Percentile is using Hyndman's type 1 for interpolation per default. You are adding type 8. (If we use the same naming of types than for R cran quantile() function).

These type of interpolation are based on formulas that are not easy to grasp for a WarpScript users. Maybe we can adopt intuitive interpolations ? Midpoint or nearest point for example.

warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
randomboolean: Actually I think it's cleaner to do
Math.floor(p*(nonNullLength -1))
so there is no need to reindex from 1-indexed to 0-indexed

warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
ftence: I disagree. Even if it isn't the term used by Hyndman (discontinuous and continuous is), interpolation make the behavior very clear: we're interpolating between GTS points or not. This is equally true for users which may not be experts in percentile but want to choose between a point in the original GTS or an interpolated point.

ftence:
We do not expose the types to the user, and usually they don't care which formula is used underneath as long as the result is correct.
I don't see the point in switching to simpler formulas as they will have fewer "good" properties than the two formulae proposed here.

warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
ftence: That would lead to a different result. For instance with `nonNullLength = 4` and `p = 0.9`:
```
Math.ceil(p * nonNullLength) - 1 = 3
Math.floor(p * (nonNullLength - 1)) = 2
```

ftence:
Comparison with R.

R code:
```
data <- seq(1, 100)
for(t in 1:9) {print(quantile(data, probs = c(0.0,0.01,0.015,0.1,0.45789,0.991,0.9991,0.999,1.0), type=t))}
```

WarpScript code:
```
[] 1 100 <% +! %> FOR
[] 1 100 <% 10.0 / +! %> FOR
[] 1 100 <% 10.0 / +! %> FOR
[] 1 100 <% +! %> FOR
[] 1 100 <% +! %> FOR
MAKEGTS
'gts' STORE

[ 0.0 0.01 0.015 0.1 0.45789 0.991 0.9991 0.999 1.0 ]
'quantiles' STORE

[]
1 9
<%
  't' STORE
  $quantiles
  <%
    'p' STORE
    [
      $gts
      $p 100 * 'type' $t TOSTRING + mapper.percentile
      MAXLONG
      MAXLONG
      1
      1
      T
    ]
    MAP
    
    VALUES
  %>
  F LMAP

  FLATTEN
  +!
%>
FOR
```

Comparison of the result, column are quantile and rows are langage(type):
||0.0|0.01|0.015|0.1|0.45789|0.991|0.9991|0.999|1.0|
|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|
|R(1)|1|1|2|10|46|100|100|100|100|
|WS(1)|1|1|2|10|46|100|100|100|100|
|R(2)|1|1.5|2|10.5|46|100|100|100|100|
|WS(2)|1.0|1.5|2.0|10.5|46.0|100.0|100.0|100.0|100.0|
|R(3)|1|1|2|10|46|99|100|100|100|
|WS(3)|1|1|2|10|46|99|100|100|100|
|R(4)|1|1|1.5|10|45.789|99.1|99.91|99.9|100|
|WS(4)|1.0|1.0|1.5|10.0|45.789|99.1|99.91|99.9|100.0|
|R(5)|1|1.5|2|10.5|46.289|99.6|100|100|100|
|WS(5)|1.0|1.5|2.0|10.5|46.289|99.6|100.0|100.0|100.0|
|R(6)|1|1.01|1.515|10.1|46.24689|100|100|100|100|
|WS(6)|1.0|1.01|1.515|10.1|46.24689|100.0|100.0|100.0|100.0|
|R(7)|1|1.99|2.485|10.9|46.33111|99.109|99.9109|99.901|100|
|WS(7)|1.0|1.99|2.485|10.9|46.33111|99.109|99.9109|99.90100000000001|100.0|
|R(8)|1|1.33666666666667|1.83833333333333|10.3666666666667|46.2749633333333|99.7636666666667|100|100|100|
|WS(8)|1.0|1.3366666666666667|1.8383333333333334|10.366666666666667|46.27496333333333|99.76366666666667|100.0|100.0|100.0|
|R(9)|1|1.3775|1.87875|10.4|46.2784725|99.72275|100|100|100|
|WS(9)|1.0|1.3775|1.87875|10.4|46.2784725|99.72274999999999|100.0|100.0|100.0|

randomboolean (approved):

warp10/src/main/java/io/warp10/script/aggregator/Percentile.java
```diff
@@ -138,73 +163,199 @@ public int compare(Integer idx1, Integer idx2) {
...
+
+    return new Object[] {tick, location, elevation, value};
   }
 
   @Override
   public String toString() {
     StringBuilder sb = new StringBuilder();
     sb.append(StackUtils.toString(this.percentile));
     sb.append(" ");
+    sb.append("type" + type);
```
hbs, line 357: This will lead to invalid WarpScript, the type needs to be enclosed in single quotes.

PR #851, opened 2020-09-23
FILLMAPPER

randomboolean:
Wrap a Mapper so empty buckets are filled by the mapper and non-empty buckets are left untouched.

Unseen buckets by MAP (for example if step > 1) are discarded


warp10/src/main/java/io/warp10/script/mapper/FILLMAPPER.java
```diff
@@ -0,0 +1,90 @@
...
+
+import io.warp10.script.NamedWarpScriptFunction;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptLib;
+import io.warp10.script.WarpScriptMapperFunction;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.script.WarpScriptStackFunction;
+
+/**
+ * Wrap a Mapper so empty buckets are filled by the mapper and non-empty buckets are left untouched
```
hbs, line 27: This comment deserves a rewrite, bucket is a term used in BUCKETIZE, not so in MAP, so the comment deserves to explain more thoroughly the intent, i.e. to generate values for ticks which fall on bucket ends with no associated values when MAP is called on a bucketized GTS. The comment must enable someone diving into the code to fully understand what it does.

warp10/src/main/java/io/warp10/script/mapper/FILLMAPPER.java
```diff
@@ -0,0 +1,90 @@
...
+      super(name);
+      this.mapper = mapper;
+    }
+
+    @Override
+    public Object apply(Object[] args) throws WarpScriptException {
+
+      long[] ticks = (long[]) args[3];
+      long tick = (long) args[0];
+      int tick_index = (int) ((long[]) args[7])[4];
```
hbs, line 48: Explain what this is, args[7][4] is kind of cryptic for someone not knowing what args contains.

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -5069,6 +5069,11 @@ public static void reset(GeoTimeSerie gts) {
 
   public static List<GeoTimeSerie> map(GeoTimeSerie gts, Object mapper, long prewindow, long postwindow, long occurrences, boolean reversed, int step, boolean overrideTick, WarpScriptStack stack,
                                        List<Long> outputTicks, boolean dedup) throws WarpScriptException {
+    return map(gts, mapper, prewindow, postwindow, occurrences, reversed, step, overrideTick, stack, outputTicks, false, false);
+  }
+
+  public static List<GeoTimeSerie> map(GeoTimeSerie gts, Object mapper, long prewindow, long postwindow, long occurrences, boolean reversed, int step, boolean overrideTick, WarpScriptStack stack,
+                                       List<Long> outputTicks, boolean dedup, boolean fillOnly) throws WarpScriptException {
```
hbs, line 5076: Why add a fillOnly parameter? This value can be computed in map by checking if mapper is an instance of FILLMAPPER.

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -5208,74 +5213,78 @@ public static void reset(GeoTimeSerie gts) {
...
-          if (null == outputTicks || !reversed) {
-            start = prewindow <= mapped.bucketcount ? tick - prewindow * mapped.bucketspan : Long.MIN_VALUE;
-          } else {
-            start = prewindow - 1 <= mapped.bucketcount ? tick - (prewindow - 1) * mapped.bucketspan : Long.MIN_VALUE;
-          }
-        } else {
-          if (null == outputTicks) {
-            if (reversed) {
-              start = idx + prewindow < ticks.length ? (ticks[idx + (int) prewindow]) : Long.MIN_VALUE;
+      // we don't need to extract more than a single point if we use a FillerMapper on a bucketized GTS on a non-empty bucket
```
hbs, line 5216: Rewrite the comment, the fact that the GTS is bucketized is not directly visible in the condition you inserted. Think about clarity for people reading the comment.

hbs (changes requested):

warp10/src/main/java/io/warp10/script/mapper/FILLMAPPER.java
```diff
@@ -0,0 +1,92 @@
...
+import io.warp10.script.NamedWarpScriptFunction;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptLib;
+import io.warp10.script.WarpScriptMapperFunction;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.script.WarpScriptStackFunction;
+
+/**
+ * The BUCKETIZE function separate the time axis into buckets of time to process Geo Time Series (gts).
+ * A gts that has been bucketized by the function BUCKETIZE have ticks at regular timestamps (bucketends) separated by the same time span (bucketspan).
```
hbs, line 28: ... has ticks at regular timestamps

warp10/src/main/java/io/warp10/script/mapper/FILLMAPPER.java
```diff
@@ -0,0 +1,92 @@
...
+ * Some of those ticks are not associated with a value. They are gap, aka missing values, in the bucketized gts.
```
hbs, line 29: They are gaps, ...

hbs (changes requested):

warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -24,6 +24,7 @@
 import io.warp10.script.WarpScriptStackFunction;
 import io.warp10.script.WarpScriptException;
 import io.warp10.script.WarpScriptStack;
+import io.warp10.script.mapper.FILLMAPPER;
```
ftence, line 27: Unused import.

warp10/src/main/java/io/warp10/script/mapper/FILLMAPPER.java
```diff
@@ -0,0 +1,92 @@
...
+    public FillerMapper(String name, WarpScriptMapperFunction mapper) {
+      super(name);
+      this.mapper = mapper;
+    }
+
+    @Override
+    public Object apply(Object[] args) throws WarpScriptException {
+
+      // args[7] contains the window parameters [ prewindow, postwindow, start, stop, tick index ]
+      int tick_index = (int) ((long[]) args[7])[4];
```
ftence, line 50: Rename variable to follow the Java naming convention.

warp10/src/main/java/io/warp10/script/mapper/FILLMAPPER.java
```diff
@@ -0,0 +1,92 @@
...
+
+      // if the tick is missing in the window, apply mapper to fill it
+      if (-1 == tick_index) {
+        return mapper.apply(args);
+      }
+
+      // if tick is present, set back the previous value
```
ftence, line 57: "previous"  is ambiguous, prefer something along the lines "return/keep the current datapoint untouched".

randomboolean:
The behavior on duplicate ticks should be discussed. 
Since it doesn't touch other non-duplicate ticks, right now it will just discard all duplicates except the first tick.
Maybe it should apply the mapper to deduplicate ?

ftence (changes requested):

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -5101,10 +5102,10 @@ public static void reset(GeoTimeSerie gts) {
     GeoTimeSerie mapped = gts.clone();
 
     //
-    // Do nothing if there are no values and gts was not bucketized
+    // Do nothing if gts was not bucketized and either there are no value or the mapper is a filler mapper
```
ftence, line 5105: Better thrown an exception if a macrofiller is used and the GTS is not bucketized.

ftence (changes requested):

ftence:
I fail to see how that function can be used effectively.

Most mappers return values that are not the same unit as the initial values (rate, pow, tostring, dotproduct, var, minute, hdist, log, ...), may not fill (eq, lt, gt, finite, ...) or must be applied on a single value (abs, round, ...).

Maybe this is meant to be used in conjunction with MACROMAPPER or only with mapper.sum, mean, median? Even though, I think there is a major flaw in twisting MAP to do the job of FILL: missing values in buckets must be approximately even spaced and the ratio missing values to present values must be known. This is because when you specify a window in MAP, it takes into account empty buckets.

For instance, if you bucketize a GTS with buckets of 1 second:
- If you have a present value every 2 buckets, you must have a window of at least 2 points or 2 seconds, pre or post.
- If you have a present value every 10 buckets, you must have a window of at least 10 points or 10 seconds, pre or post.

The problem is if you have a GTS with sometimes 10 missing values in a row or 2 missing values in a row, you have to chose a window size. We could go with a window a size 10 pre or post but that would probably introduce artifacts when there are few missing values.

Here is an example showing the issue:
```
0 't' STORE

NEWGTS
'raw' RENAME
50 80
<%
  'i' STORE
  $t RANDOM $i LOG * + 't' STORE
  $t s NaN NaN NaN $t ADDVALUE
%>
FOR
'raw' STORE

[
  $raw
  bucketizer.mean
  0
  1 s
  0
]
BUCKETIZE
0 GET
'bucketized' RENAME
'bucketized' STORE

$bucketized
[ NaN NaN NaN -1 ] FILLVALUE 
0 < [ SWAP T mapper.replace 0 0 0 ] MAP 
0 GET
'holes' RENAME
'holes' STORE

[
  $bucketized
  mapper.mean
  FILLMAPPER
  1 1 0
] MAP
0 GET
[ NaN NaN NaN -1 ] FILLVALUE // to visualize non-filled missing values
'fillmapper.mean.1p' RENAME
'fillmapper.mean.1p' STORE


[
  $bucketized
  mapper.mean
  FILLMAPPER
  2 2 0
] MAP
0 GET
[ NaN NaN NaN -1 ] FILLVALUE // to visualize non-filled missing values
'fillmapper.mean.2p' RENAME
'fillmapper.mean.2p' STORE


$bucketized
$bucketized [ NaN NaN NaN -1 ] FILLVALUE
filler.interpolate
FILL
DROP
[ NaN NaN NaN -1 ] FILLVALUE // to visualize non-filled missing values
'FILL' RENAME
'FILL' STORE

$FILL
$fillmapper.mean.2p
$fillmapper.mean.1p
$holes
$bucketized
// $raw
```

FILLMAPPER with a window of 1 pre, 1 post cannot fill all the values because its window is too narrow. It also creates some artifacts because sometimes the window is too wide.
FILLMAPPER with a window of 2 pre, 2 post does not leave missing values but creates a lot of artifacts because the window is too wide.

FILL reproduces exactly the behavior of INTERPOLATE in this case and the graph exactly superimposes the one of the bucketized GTS, but without any missing value.

image

There is certainly room for improvement to make FILL on bucketized GTS easier to use and also more optimized by filling only one GTS but I think this is the framework to use if you need to fill a bucketized GTS.

randomboolean:
The other solution is to add another syntax `[ [ $gts1 $gts2 ... ] $filler ] FILL [ $result_gts1 $result_gts2 ... ]`, which is similiar to that of other frameworks, and to implement more fillers.
I dismissed that option because this syntax can make the usage of FILL confusing with its original syntax (e.g. ` [ [ $a $b ] $f ]  FILL` is quite different from `$a $b $f FILL`).
In light of the points you bring I now think it's the correct thing to do. What do you think ?

ftence:
So the original goal of the the PR was to make `FILL` work on more than 2 GTSs, is that right?
That's indeed a nice idea because it would make FILL + REDUCE really compatible.

What do you mean by "[ [ $a $b ] $f ] FILL is quite different from $a $b $f FILL"? It is syntactically different, yes, but it would give the same GTSs, the former in a list, the later with 2 elements on the stack or am I missing something?

I think it's OK to introduce the `[ [ $gts1 $gts2 ... ] $filler ] FILL [ $result_gts1 $result_gts2 ... ]` signature while keeping the old one to make sure not to break anything. Moreover, I think this signature, on top of being nice to combine with REDUCE, will borrow a lot of code (sweeping line) to REDUCE. So I think it would be nice to introduce partitions in this signature. Thus the old and new signatures would be quite different, avoiding confusion:  `[ [ $gts1 $gts2 ... ] [ 'label1' ... ] $filler ] FILL [ $result_gts1 $result_gts2 ... ]`. Look also at the possibility and relevance of adding PFILL, in the same way PREDUCE exists for REDUCE.

@hbs: it would be nice to have you opinion on that as you introduced the FILL framework. Any reason to limit to 2 GTSs?

randomboolean:
No the initial goal was to fill empty buckets of a GTS while looping through its buckets. That is, extend FILLVALUE, INTERPOLATE, FILLNEXT and FILLPREVIOUS to a more general framework.
This is a task commonly named as missing value imputation.

This PR started as making a function that would convert a mapper into a filler so that the work can be done by MAP. But as you pointed out, this is flawed since the windows parameters of MAP are not suited for this task. It would be better to use as arguments the numbers of previous and next non empty buckets.

FILL was suggested as a better framework than MAP for this task. But FILL needs at least two GTS and modify one according to the other. Hence why I suggested the other syntax since the implemented logic is quite different.

Another solution could be to make another function that is not related neither to MAP nor to FILL.

hbs:
Closing this issue as another option is being considered.

PR #852, opened 2020-09-25
Rolling back signature change.

hbs:


ftence (approved):
A bit of context: this rollback keeps the method signature the same for pre-2.7.0- and post-2.7.0 (understand 2.7.0 excluded). Thus, for an extension using this method, the same jar can be deployed on those pre and post 2.7.0 versions.
Extensions compiled with pre/post 2.7.0 Warp 10 jar will not work with 2.7.0 and vice versa.

PR #853, opened 2020-09-28
Add a limit parameter to SPLIT

ftence:
Performance is the same when used without limit.

With limit, the performance can be increased, for instance:
```
UUID '-' 2 SPLIT 0 GET // Faster
UUID '-'   SPLIT 0 GET // Slower
```

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/SPLIT.java
+++ b/warp10/src/main/java/io/warp10/script/functions/SPLIT.java
@@ -54,15 +54,11 @@ public class SPLIT extends NamedWarpScriptFunction implements WarpScriptStackFun
       o = stack.pop();
     }
 
-    if (!(o instanceof String)) {
+    if (!(o instanceof String) || 1 != ((String) o).length()) {
       throw new WarpScriptException(getName() + " expects a string delimiter of length 1.");
     }
 
-    if (1 != o.toString().length()) {
-      throw new WarpScriptException(getName() + " expects a string delimiter of length 1.");
-    }
-
-    char delimiter = o.toString().charAt(0);
+    char delimiter = ((String) o).charAt(0);
 
     o = stack.pop();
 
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/functions/SPLIT.java
```diff
@@ -16,56 +16,94 @@
...
+   * @return A List of splits.
+   */
+  public static List<String> split(String input, char delim, int limit) {
+    ArrayList<String> l = new ArrayList<String>();
+    int offset = 0;
+    int splits = 1;
+
+    while (splits < limit) {
+      int index = input.indexOf(delim, offset);
+      if (index == -1) {
```
hbs, line 92: -1 == index

hbs (changes requested):

warp10/src/main/java/io/warp10/script/functions/SPLIT.java
```diff
@@ -16,56 +16,94 @@
...
     char delimiter = o.toString().charAt(0);
-    
+
     o = stack.pop();
 
     if (!(o instanceof String)) {
       throw new WarpScriptException(getName() + " operates on a String.");
     }
 
+    stack.push(split(o.toString(), delimiter, limit));
```
hbs, line 73: Could use (String) o since you checked that o was indeed an instance of String.

warp10/src/main/java/io/warp10/script/functions/SPLIT.java
```diff
@@ -16,56 +16,94 @@
...
+
+      o = stack.pop();
+    }
+
     if (!(o instanceof String)) {
       throw new WarpScriptException(getName() + " expects a string delimiter of length 1.");
     }
-    
+
     if (1 != o.toString().length()) {
```
hbs, line 61: Can use (String) o

hbs (changes requested):

PR #854, opened 2020-09-28
Allow GET on String and make path-like parameter work on all accepted types

ftence:


warp10/src/main/java/io/warp10/script/functions/GET.java
```diff
@@ -17,74 +17,93 @@
...
+   * Get a value from a List, Map, byte[] or String using respectively a Long, Object, Long or Long key.
+   * @param key Either on Object for a Map or a Long for List, byte[] or String. Negative indexing is possible, in that case the corresponding index is size + index.
+   * @param collection Either a List, Map, byte[] or String instance.
+   * @return The value at the given key.
+   * @throws WarpScriptException If the collection type cannot be handled or the key is invalid for the collection type.
+   */
+  public static Object get(Object key, Object collection) throws WarpScriptException {
+    if (collection instanceof List) {
+      if (!(key instanceof Long)) {
+        throw new WarpScriptException("Getting on List requires a Long.");
```
hbs, line 73: LIST

warp10/src/main/java/io/warp10/script/functions/GET.java
```diff
@@ -17,74 +17,93 @@
...
       }
 
-      value = nestedGet((List) coll, (List<Long>) key);
-    }
-    
-    stack.push(value);
+      int idx = ((Long) key).intValue();
+      int size = ((byte[]) collection).length;
+      idx = computeAndCheckIndex(idx, size);
+      return ((byte[]) collection)[idx] & 0xFFL;
```
hbs, line 102: It would probably be better to return a byte array of length 1 rather than a LONG, it would be more coherent with the way GET behaves for STRINGs.

hbs (changes requested):

warp10/src/main/java/io/warp10/script/functions/GET.java
ftence: That would change the current behavior.

hbs:
Actually I think the behavior you are introducing is counter intuitive, if 'GET'ting a list of indices I would expect the result to be a list of values. I think it would probably be better to introduce another function to do what you implemented in your PR.

ftence:
I do agree, problem is it is already the current behavior, so I'm just extending it.

hbs:
The documentation for GET does not describe this behavior when the index parameter is a list, so it is really highly probable nobody actually expects that behavior.

ftence:
Removing the "traversing" list parameter in GET is possible as it is not in the documentation, although @randomboolean may be already using it in his code.

ftence:
For reference, we're keeping the List syntax because it is in fact already documented and we don't want to break any script a user might have coded.
However, as I already mentioned, the List syntax cannot be used on a Map because Maps can have a List key. The proposed solution limits this syntax to top-level List ie:
`[ [ 1 2 ] [ 3 4 ] ] [ -1 0 ] GET` returns `3` because the top-level collection is a List.
`{ [ 1 2 ] '1,2' [ 3 4 ] '3,4' } [ 1 2 ] GET` returns `'1,2'` because the top-level collection is a Map. 
`{ 1 { 2 3 } 4 { 5 6 } } [ 1 2 ] GET` return `NULL` because `[ 1 2 ]` is not a key in the Map.
`[ { 1 { 2 3 } 4 { 5 6 } } ] [ 0 1 2 ] GET` returns `3` because the top level collection is  a List.
`[ { [ 1 2 ] '1,2' [ 3 4 ] '3,4' } ] [ 0 [ 1 2 ] ] GET` returns `'1,2'` because the top-level collection is a List.

warp10/src/main/java/io/warp10/script/functions/GET.java
```diff
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018  SenX S.A.S.
+//   Copyright 2018-2020  SenX S.A.S.
```
hbs, line 2: Fix Copyright notice ;-)

warp10/src/main/java/io/warp10/script/functions/GET.java
```diff
@@ -17,74 +17,95 @@
...
+    stack.push(value);
 
-        idx = computeAndCheckIndex(idx, size);
+    return stack;
+  }
 
-        value = ((List) coll).get(idx);
+  /**
+   * Get a value from a List, Map, byte[] or String using respectively a Long, Object, Long or Long key.
+   * @param key Either on Object for a Map or a Long for List, byte[] or String. Negative indexing is possible, in that case the corresponding index is size + index.
```
hbs, line 67: an Object

hbs (changes requested):

PR #855, opened 2020-09-28
Add streamDelimiter parameter to HTTP definitions to stream the POST payload to the macro

ftence:
On 100s MB files, there is no noticeable difference in speed between the previous method (all the data in the request payload) and the stream method. However, the memory consumption of the stream method is, as expected, lower.

Example of a script to read [IBTrACS](https://www.ncdc.noaa.gov/ibtracs/) CSV files by splitting the stream on new lines.

```
{
  'path' '/ibtracs'
  'prefix' false
  'streamDelimiter' '%0A' 'UTF-8' ->BYTES
  'macro' 
  <% 
    'payload' STORE
    'request' STORE

    $payload ISNULL NOT
    <% 
    $payload 'UTF-8' BYTES->
    'row' STORE

      <%
        $row ',' 13 SPLIT
        'row_split' STORE
        
        $row_split 0 GET 'id' STORE

        $row_split 6 GET TRIM ' ' 'T' REPLACE 'z' +
        TOTIMESTAMP 'ts' STORE

        $row_split 8 GET TODOUBLE 'lat' STORE
        $row_split 9 GET TODOUBLE 'lon' STORE

        $row_split 10 GET TRIM
        DUP SIZE 0 >
        <% TODOUBLE %>
        <% DROP -1 %>
        IFTE
        'wind' STORE

        $row_split 11 GET TRIM
        DUP SIZE 0 >
        <% TODOUBLE %>
        <% DROP -1 %>
        IFTE
        'pressure' STORE

        T 'valid_input' STORE
      %>
      <%
        // Ingore invalid input
        F 'valid_input' STORE
      %>
      <% %>
      TRY
      
      $valid_input
      <%
        // Last id is stored in register 0
        PUSHR0 $id !=
        <%
          // No previous encoders or current encoder is for another id
          PUSHR0 ISNULL NOT
          <%
            // Pool of encoders stored in register 1.
            PUSHR1 ISNULL
            <% [] POPR1 %>
            IFT

            PUSHR1
            $wind_encoder +! 
            $pressure_encoder +!
            
            SIZE 1000 > // UPDATE every 1000 encoders
            <%
              PUSHR1 $request 'params' GET 'token' GET 0 GET UPDATE
              [] POPR1
            %>
            IFT
          %>
          <%
            $request 'params' GET 'token' GET 0 GET
            'token' STORE
          %>
          IFTE

          $row_split 5 GET 'name' STORE 

          NEWENCODER
          'stream.wind' RENAME
          { 'id' $id } RELABEL
          { 'name' $name } SETATTRIBUTES
          'wind_encoder' STORE
          NEWENCODER
          'stream.pressure' RENAME
          { 'id' $id } RELABEL
          { 'name' $name } SETATTRIBUTES
          'pressure_encoder' STORE

          $id POPR0
        %>
        IFT

        $wind_encoder
        $ts $lat $lon NaN $wind ADDVALUE
        DROP

        $pressure_encoder
        $ts $lat $lon NaN $pressure ADDVALUE
        DROP
      %>
      IFT
    %>
    <%
      PUSHR1 $request 'params' GET 'token' GET 0 GET UPDATE

      {
        'status' 200
      } 
    %>
    IFTE
  %>
  ASREGS
}
```

### DIFF
--- a/warp10/src/main/java/io/warp10/plugins/http/HTTPWarp10Plugin.java
+++ b/warp10/src/main/java/io/warp10/plugins/http/HTTPWarp10Plugin.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018  SenX S.A.S.
+//   Copyright 2018-2020  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
--- a/warp10/src/main/java/io/warp10/plugins/http/HTTPWarp10Plugin.java
+++ b/warp10/src/main/java/io/warp10/plugins/http/HTTPWarp10Plugin.java
@@ -112,7 +112,7 @@ public class HTTPWarp10Plugin extends AbstractWarp10Plugin implements Runnable {
   /**
    * Map of uri to stream delimiter.
    */
-  private Map<String, Byte> streamDelimiter = new HashMap<String, Byte>();
+  private Map<String, Byte> streamDelimiters = new HashMap<String, Byte>();
 
   /**
    * Map of filename to uri
@@ -248,7 +248,7 @@ public class HTTPWarp10Plugin extends AbstractWarp10Plugin implements Runnable {
           String uri = uris.remove(spec);
           this.macros.remove(uri);
           this.parsePayloads.remove(uri);
-          this.streamDelimiter.remove(uri);
+          this.streamDelimiters.remove(uri);
           this.sizes.remove(spec);
           this.prefixes.remove(uri);
         }
@@ -259,7 +259,7 @@ public class HTTPWarp10Plugin extends AbstractWarp10Plugin implements Runnable {
         for (String uri: inactiveURIs) {
           this.macros.remove(uri);
           this.parsePayloads.remove(uri);
-          this.streamDelimiter.remove(uri);
+          this.streamDelimiters.remove(uri);
           this.prefixes.remove(uri);
         }        
       } catch (Throwable t) {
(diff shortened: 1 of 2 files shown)


warp10/src/main/java/io/warp10/plugins/http/HTTPWarp10Plugin.java
```diff
@@ -108,6 +109,11 @@
    */
   private Map<String, Boolean> parsePayloads = new HashMap<String, Boolean>();
 
+  /**
+   * Map of uri to stream delimiter.
+   */
+  private Map<String, Byte> streamDelimiter = new HashMap<String, Byte>();
```
hbs, line 115: field should use plura to conform to the other field names.

warp10/src/main/java/io/warp10/plugins/http/WarpScriptHandler.java
```diff
@@ -103,14 +112,29 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
...
+      } else {
+        // The input stream will be splitted and given to the macro later.
+        if (null != request.getHeader("Content-Type") && "application/gzip".equals(request.getHeader("Content-Type"))) {
+          inputStream = new GZIPInputStream(request.getInputStream());
+        } else {
+          inputStream = request.getInputStream();
         }
       }
 
+      // Get the parameters in the URL and payload if we want to parse the x-www-form-urlencoded payload.
```
hbs, line 137: The lines below are unclear, they are executed even if isParsePayload is false, the fact that it works should be described in a comment.

warp10/src/main/java/io/warp10/plugins/http/WarpScriptHandler.java
```diff
@@ -166,10 +233,19 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
...
     } finally {
       WarpConfig.clearThreadProperties();
       WarpScriptStackRegistry.unregister(stack);
-    }    
+    }
+  }
+
+  private static int indexOf(byte[] array, byte target, int start, int end) {
+    for (int i = start; i < end; i++) {
+      if (array[i] == target) {
```
hbs, line 245: make target final and use target == array[i]

hbs (changes requested):

PR #856, opened 2020-10-06
Ecdh

hbs:


ftence (approved):
A quick comment on the purpose of this function would be welcome ;).

hbs:
The ECDH function performs an Elliptic-Curve Diffie-Hellman key agreement procedure as per https://en.wikipedia.org/wiki/Elliptic-curve_Diffie%E2%80%93Hellman

PR #858, opened 2020-10-09
Make DTW work on locations and elevations and add window parameter

ftence:


warp10/src/main/java/io/warp10/script/functions/DTW.java
```diff
@@ -16,186 +16,372 @@
...
+          for (int i = 0; i < dimVal1.length; i++) {
+            if (dimVal1[i] < min) {
+              min = dimVal1[i];
+            }
+            if (dimVal1[i] > max) {
+              max = dimVal1[i];
+            }
+          }
+
+          if (max - min < 0.000001) {
```
hbs, line 310: should make this a constant with an attached comment. Also why 10E-6?

warp10/src/main/java/io/warp10/script/functions/DTW.java
```diff
@@ -16,186 +16,372 @@
...
+          for (int i = 0; i < dimVal2.length; i++) {
+            if (dimVal2[i] < min) {
+              min = dimVal2[i];
+            }
+            if (dimVal2[i] > max) {
+              max = dimVal2[i];
+            }
+          }
+
+          if (max - min < 0.000001) {
```
hbs, line 332: ditto

warp10/src/main/java/io/warp10/script/functions/DTW.java
```diff
@@ -205,77 +391,159 @@ public double compute(double[] values1, int offset1, int len1, double[] values2,
...
+
+  public static double euclidean(double[][] values1, int index1, double[][] values2, int index2) throws WarpScriptException {
+    double d = 0D;
+    for (int dimension = 0; dimension < values1.length; dimension++) {
+      d += Math.pow(values1[dimension][index1] - values2[dimension][index2], 2.0D);
     }
-    
-    if (!manhattan) {
-      dtwDist = Math.sqrt(dtwDist);
+    return Math.sqrt(d);
```
hbs, line 521: if you do not need the actual distance you could remove the call to sqrt, it would not change the result

warp10/src/main/java/io/warp10/script/functions/DTW.java
```diff
@@ -205,77 +391,159 @@ public double compute(double[] values1, int offset1, int len1, double[] values2,
...
+    if (2 != values1.length) {
+      throw new WarpScriptException("Loxodromic distance must be given two dimensions: lat, lon.");
+    }
+
+    double lat1 = values1[0][index1];
+    double lat2 = values2[0][index2];
+    double lon1 = values1[1][index1];
+    double lon2 = values2[1][index2];
+
+    return GeoXPLib.loxodromicDistance(GeoXPLib.toGeoXPPoint(lat1, lon1), GeoXPLib.toGeoXPPoint(lat2, lon2));
```
hbs, line 534: You should re-implement the distance computation directly on lat/lon instead of converting to HHCode then letting loxodromicDistance split the HHCode. Same goes for the orthodromic distance below.

warp10/src/main/java/io/warp10/script/functions/OPTDTW.java
```diff
@@ -53,6 +51,20 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
     int count = ((Number) o).intValue();
     
     o = stack.pop();
+
+    // Optional window parameter.
+    int window = Integer.MAX_VALUE;
+
+    if (o instanceof Number) {
```
hbs, line 58: Should check that it is a LONG rather than a number

hbs (changes requested):

warp10/src/main/java/io/warp10/script/functions/DTW.java
ftence: This affects how the threshold works and the final result, so it's better to keep the sqrt.

I'll add euclideansq as a new distance to speed up computation while making sure the user knows the value is squared.

hbs (approved):

PR #864, opened 2020-11-03
Add missing configuration keys to configuration templates

ftence:


etc/conf.templates/distributed/10-ingress.conf.template
```diff
@@ -293,7 +293,12 @@ ingress.kafka.metadata.maxsize = 900000
 //
 // Identification of Ingress Metadata Update endpoint source
 //
-#ingress.metadata.update = 
+#ingress.metadata.update =
+
+//
+// Identification of Ingress Metadata Update endpoint source when doing a delta update of attributes
+//
+#ingress.metadata.update.delta =
```
hbs, line 301: This is not a configuration key

hbs (changes requested):

etc/conf.templates/distributed/10-ingress.conf.template
ftence: Neither `ingress.metadata.update`, removing both.

PR #865, opened 2020-11-05
Add SHMDEFINED which tells if a symbol is defined in SHM

ftence:
Currently, to check if a SHM symbol exists, we have to use `SHMLOAD` and `TRY` to catch the exception if the symbol is not defined. This strategy has to be used everywhere as the symbol may have expired because of the TTL.

This "better ask for forgiveness than permission" method is usually not the way to go in Java and thus in WarpScript. `SHMDEFINED` used in conjunction with `IFT` fixes that. When the symbol does not exist, this method is ~5x faster, when it exists, it is marginally slower.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/ext/shm/SHMDEFINED.java
+++ b/warp10/src/main/java/io/warp10/script/ext/shm/SHMDEFINED.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2020  SenX S.A.S.
+//   Copyright 2020  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


warp10/src/main/java/io/warp10/script/ext/shm/SHMDEFINED.java
```diff
@@ -0,0 +1,43 @@
+//
+//   Copyright 2018-2020  SenX S.A.S.
```
hbs, line 2: 2020

warp10/src/main/java/io/warp10/script/ext/shm/SHMDEFINED.java
```diff
@@ -0,0 +1,43 @@
...
+  public SHMDEFINED(String name) {
+    super(name);
+  }
+
+  @Override
+  public Object apply(WarpScriptStack stack) throws WarpScriptException {
+    Object top = stack.pop();
+    
+    if (!(top instanceof String)) {
+      throw new WarpScriptException(getName() + " expects a symbol name on top of the stack.");
```
hbs, line 34: Avoid mention of the stack, "expects a symbol name"

hbs (changes requested):

PR #866, opened 2020-11-06
Fix concurrency in StandaloneDirectoryClient

ftence:
`classids` was a `HashMap` which is not thread safe although it is accessed by several threads. In particular, it resulted in `CLASS_COMPARATOR` to fail with NPE because the `get` couldn't find the key because of a concurrent `put`.

Also `unregister` is not synchronized anymore because it was the only synchronized method and no synchronization on `this` was done. On the contrary, critical parts are synchronized on `metadatas`, so the critical part in `unregister` is now synchronized on `metadatas` as well.

### DIFF
--- a/warp10/src/main/java/io/warp10/standalone/StandaloneDirectoryClient.java
+++ b/warp10/src/main/java/io/warp10/standalone/StandaloneDirectoryClient.java
@@ -119,10 +119,20 @@ public class StandaloneDirectoryClient implements DirectoryClient {
   private static final Comparator<String> CLASS_COMPARATOR = new Comparator<String>() {
     @Override
     public int compare(String o1, String o2) {
-      long id1 = classids.get(o1);
-      long id2 = classids.get(o2);
+      Long id1 = classids.get(o1);
+      Long id2 = classids.get(o2);
 
-      return Directory.ID_COMPARATOR.compare(id1, id2);
+      // A key may be missing during a find if a concurrent unregister is done.
+      if (null == id1) {
+        if (null == id2) {
+          return 0;
+        }
+        return -1;
+      } else if (null == id2) {
+        return 1;
+      } else {
+        return Directory.ID_COMPARATOR.compare(id1, id2);
+      }
     }
   };
 
--- a/warp10/src/main/java/io/warp10/standalone/StandaloneDirectoryClient.java
+++ b/warp10/src/main/java/io/warp10/standalone/StandaloneDirectoryClient.java
@@ -605,7 +605,7 @@ public class StandaloneDirectoryClient implements DirectoryClient {
     } else {
       return (List<Metadata>) metadatas;
     }
-  };
+  }
 
   public void register(Metadata metadata) throws IOException {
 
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/standalone/StandaloneDirectoryClient.java
```diff
@@ -680,19 +681,22 @@ public void register(Metadata metadata) throws IOException {
     }
   }
 
-  public synchronized void unregister(Metadata metadata) {
-    if (!classids.containsKey(metadata.getName())) {
-      return;
-    }
+  public void unregister(Metadata metadata) {
     // 128BITS
     long labelsId = GTSHelper.labelsId(this.labelsLongs, metadata.getLabels());
```
hbs, line 686: Why compute the labels id before it is needed? This is a costly operation.

hbs (changes requested):

warp10/src/main/java/io/warp10/standalone/StandaloneDirectoryClient.java
```diff
@@ -680,19 +681,22 @@ public void register(Metadata metadata) throws IOException {
...
+      if (!classids.containsKey(metadata.getName())) {
+        return;
+      }
+      if (!metadatas.get(metadata.getName()).containsKey(labelsId)) {
+        return;
+      }
+      metadatas.get(metadata.getName()).remove(labelsId);
+      if (metadatas.get(metadata.getName()).isEmpty()) {
+        metadatas.remove(metadata.getName());
+        classids.remove(metadata.getName());
```
hbs, line 698: On line 632 (in register), the call to metadatas.get() could throw an NPE if the entry for that class name is removed on line 598. The solution would be to protect the part in register with a syncrhonized block on metadatas

hbs: So register and unregister might have to be mutually exclusive, hence maybe both synchronized

ftence: I tried to reduce at the minimum synchronization, avoiding making method synchronized.

warp10/src/main/java/io/warp10/standalone/StandaloneDirectoryClient.java
```diff
@@ -680,19 +706,23 @@ public void register(Metadata metadata) throws IOException {
...
-    if (metadatas.get(metadata.getName()).isEmpty()) {
-      metadatas.remove(metadata.getName());
-      classids.remove(metadata.getName());
+    long labelsId;
+
+    synchronized(metadatas) {
+      if (!classids.containsKey(metadata.getName())) {
+        return;
+      }
+      labelsId = GTSHelper.labelsId(this.labelsLongs, metadata.getLabels());
```
hbs, line 717: Actually second thoughts about where labelsId is computed. If it is occasional to call unregister for a GTS whose class is  no longer known it might be better to always compute labelsId where labelsId is declared rather than in the synchronized block since the computation is CPU heavy and therefore would extend the time spent in the syncrhonized block which would block all threads attempting to synchronize on metadatas. So probably better to always compute labelsId and do so outside of the synchronized block.

PR #867, opened 2020-11-09
Modified safe delta mode to allow for fast merges in more cases

hbs:


warp10/src/main/java/io/warp10/continuum/gts/GTSEncoder.java
```diff
@@ -378,22 +386,26 @@ public synchronized int addValue(long timestamp, long location, long elevation,
...
-      if (binaryString.equals(lastStringValue)) {
+      if (validLastStringValue && binaryString.equals(lastStringValue)) {
         tsTypeFlag |= FLAGS_VALUE_IDENTICAL;
       }
     } else if (value instanceof Double || value instanceof Float) {
       tsTypeFlag |= FLAGS_TYPE_DOUBLE;
       // Only compare to the previous double value if the last floating point value was NOT encoded as a BigDecimal
-      if (null == lastBDValue && lastDoubleValue == ((Number) value).doubleValue()) {
+      if (validLastDoubleValue
+          && null == lastBDValue
```
ftence, line 405: I think `validLastDoubleValue` and `validLastBDValue` could also be used to tell if the last floating-point value is encoded as BigDecimal or doubleinstead of setting `lastBDValue` to `null`:
- `validLastDoubleValue && !validLastBDValue` -> last value encoded as double
- `!validLastDoubleValue && validLastBDValue` -> last value encoded as BigDecimal
- `!validLastDoubleValue && !validLastBDValue` -> unknown last value floating-point value
- `validLastDoubleValue && validLastBDValue` -> cannot happen

Thus `validLastDoubleValue && null == lastBDValue` could be replaced by `validLastDoubleValue`.

warp10/src/main/java/io/warp10/continuum/gts/GTSEncoder.java
```diff
@@ -405,7 +417,7 @@ public synchronized int addValue(long timestamp, long location, long elevation,
       // Strip trailing zero so we optimize the representation
       doubleValue = doubleValue.stripTrailingZeros();
 
-      if (null != lastBDValue && 0 == lastBDValue.compareTo(doubleValue)) {
+      if (validLastBDValue && null != lastBDValue && 0 == lastBDValue.compareTo(doubleValue)) {
```
ftence, line 420: Ditto, only testing `validLastBDValue` should be enough to be sure `lastBDValue` is non-null. It can be argued that the nullity check prevents a NPE though.

warp10/src/main/java/io/warp10/continuum/gts/GTSEncoder.java
```diff
@@ -460,10 +472,10 @@ public synchronized int addValue(long timestamp, long location, long elevation,
         }
       } else {
         // Do nothing, implicitly we will encode location as raw GeoXPPoint
-        noDeltaMetaLocation = false;
       }
     } else {
       lastGeoXPPoint = GeoTimeSerie.NO_LOCATION;
+      validLastGeoXPPoint = false;
```
ftence, line 478: `lastGeoXPPoint = GeoTimeSerie.NO_LOCATION;` is unnecessary since `validLastGeoXPPoint` is set to `false`.

warp10/src/main/java/io/warp10/continuum/gts/GTSEncoder.java
```diff
@@ -496,10 +508,10 @@ public synchronized int addValue(long timestamp, long location, long elevation,
         if (Math.abs(elevation) < (1L << 48)) {
           locElevFlag |= FLAGS_ELEVATION_ZIGZAG;
         }
-        noDeltaMetaElevation = false;
       }
     } else {
       lastElevation = GeoTimeSerie.NO_ELEVATION;
+      validLastElevation = false;
```
ftence, line 514: `lastElevation = GeoTimeSerie.NO_ELEVATION;` is unnecessary since `validLastElevation` is set to `false`.

warp10/src/main/java/io/warp10/continuum/gts/GTSEncoder.java
```diff
@@ -702,6 +717,7 @@ public synchronized int addValue(long timestamp, long location, long elevation,
             this.stream.write(buf, 0, 8);
             // Clear the last BDValue otherwise we might incorrectly encode the next value specified as a BigDecimal
             lastBDValue = null;
+            validLastDoubleValue = true;
```
ftence, line 720: I think setting `validLastBDValue` to `false` is clearer than setting `lastBDValue` to `null`. This change goes with the first comment.

warp10/src/main/java/io/warp10/continuum/gts/GTSEncoder.java
```diff
@@ -715,6 +731,7 @@ public synchronized int addValue(long timestamp, long location, long elevation,
             this.stream.write(buf10, 0, l);
             // Keep track of last value
             lastBDValue = dvalue;
+            validLastBDValue = true;
```
ftence, line 734: Also sets `validLastDoubleValue` to `false`. Also goes with first comment.

warp10/src/main/java/io/warp10/continuum/gts/GTSEncoder.java
```diff
@@ -1129,10 +1144,14 @@ public synchronized void merge(GTSEncoder encoder) throws IOException {
       this.lastBDValue = encoder.lastBDValue;
       this.lastDoubleValue = encoder.lastDoubleValue;
       this.lastStringValue = encoder.lastStringValue;
+
+      // Disable delta encoding, this allows us to have a wider use of the fast path
+      safeDelta();
```
ftence, line 1149: Instead copy last loc/elev/values according to valid* booleans.

ftence (changes requested):

warp10/src/main/java/io/warp10/continuum/gts/GTSEncoder.java
hbs: Added the copying of the valid* flags, but left the noDeltaMetaTimestamp forced to true by leaving the call to safeDelta as a first step

ftence (approved):

PR #871, opened 2020-11-10
More detailed error messages for /update

ftence:
More descriptive parse errors, including precise index of the error.

Updated http status codes:
500 error for general server-side error
400 for bad input/request
403 for bad token
429 for exceeded throttling limits

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -2463,25 +2463,29 @@ public class GTSHelper {
     }
     
     //idx = str.indexOf("/");
-    idx = UnsafeString.indexOf(str, '/');
+    idx = UnsafeString.indexOf(str, '/', tsoffset);
 
     if (-1 == idx){
-      throw new ParseException("Missing timestamp separator.", idx);
+      throw new ParseException("Missing timestamp separator.", tsoffset);
     }
     
     long timestamp;
-    
-    if (tsoffset == idx) {
-      // No timestamp provided, use 'now'
-      timestamp = null != now ? (long) now : TimeSource.getTime();
-    } else {
-      if ('T' == str.charAt(tsoffset)) {
-        // Support T-XXX to record timestamps which are relative to 'now', useful for
-        // devices with no time reference but only relative timestamps
-        timestamp = (null != now ? (long) now : TimeSource.getTime()) + Long.parseLong(str.substring(1 + tsoffset, idx));
+
+    try {
+      if (tsoffset == idx) {
+        // No timestamp provided, use 'now'
+        timestamp = null != now ? (long) now : TimeSource.getTime();
       } else {
-        timestamp = Long.parseLong(str.substring(tsoffset,  idx));
+        if ('T' == str.charAt(tsoffset)) {
+          // Support T-XXX to record timestamps which are relative to 'now', useful for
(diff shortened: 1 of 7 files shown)


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -2535,14 +2543,22 @@ public static GTSEncoder parse(GTSEncoder encoder, String str, Map<String,String
     idx2 = UnsafeString.indexOf(str, ' ', idx);
     
     if (-1 == idx2){
-      throw new ParseException(str, idx);
+      if(0 == tsoffset) {
+        throw new ParseException("Missing GTS name, labels and value.", idx);
```
hbs, line 2547: The actual incorrect line is no longer shown.

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -2574,7 +2590,7 @@ public static GTSEncoder parse(GTSEncoder encoder, String str, Map<String,String
       // No class+labels, assume same class+labels as those in encoder, except
       // if encoder is null in which case we throw a parse exception
       if (null == encoder) {
-        throw new ParseException(str, idx);
+        throw new ParseException("Missing or invalid GTS name and labels.", idx);
```
hbs, line 2593: Ditto, dropping the invalid line.

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -2591,14 +2607,20 @@ public static GTSEncoder parse(GTSEncoder encoder, String str, Map<String,String
       idx2 = UnsafeString.indexOf(str, '}', idx);
       
       if (-1 == idx2){
-        throw new ParseException(str, idx);
+        throw new ParseException("Missing end of labels '}'.", str.length() - 1);
```
hbs, line 2610: Invalid line no longer shown.

hbs: Now that we have a mechanism to control the error message size I think it would be a good idea to include the invalid line whenever possible, it makes the diagnosis of the issue way easier.

hbs (changes requested):

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
ftence: Ingress and its declination already include the offending line when catching the ParseException.

I think it's a good way of doing things: the ParseException only tells where on the line there is a problem and what is the problem and the caller knows which lines caused the problem, so it's up to the caller to include the line in the Exception or error message.

I updated the error messages to include the offset where is was not and better handled the errors in PARSE.

hbs: GTSHelper#parse is used in PARSE also where the offending line is not reported in case of error, this could be a nice addition to this PR. Same with StreamingUpdate endpoints if that is not already the case.

ftence: That's exactly the changes in a433a5d.

hbs: My bad, sorry.

PR #875, opened 2020-11-19
Precise error message when conf files contain invalid lines, fixes #872

### LINKED ISSUE #872: Config file parsing error messages not very useful
I was setting up a Warp10 instance on Debian 10. Ran `bootstrap`, edited config files and ran `start` but instead of starting, got this message back:
```bash
warp10@warp10:~$ ./bin/warp10-standalone.sh start
Invalid syntax on line 1531, will force an abort.
Aborting due to 1 error.
```
First of all, it doesn't say where the syntax error is coming from. I thought I might be using `zsh` or smth before I realized it's talking about config files. Once I reset my changes to config files it went through. Line number doesn't help either as it concatenates all files together and then parses them (I think).

Thanks!
/cc @pi-r-p

ftence:
When a configuration file contain errors (multiple `=` or invalid url-encoding), starting Warp 10 produces a message similar to this:
`Found errors while reading /opt/warp10/etc/conf.d/70--extensions.conf. (Malformed lines [61, 80].)`

Fixes #872.

### DIFF
--- a/warp10/src/main/java/io/warp10/WarpConfig.java
+++ b/warp10/src/main/java/io/warp10/WarpConfig.java
@@ -24,15 +24,20 @@ import io.warp10.script.WarpScriptMacroRepository;
 
 import com.google.common.annotations.Beta;
 import com.google.common.io.Files;
+import com.google.common.io.LineProcessor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.StringReader;
+import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
@@ -40,6 +45,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


ftence:
Fixes #872.

warp10/src/main/java/io/warp10/WarpConfig.java
```diff
@@ -122,62 +126,63 @@ public static void clearThreadProperties() {
     threadProperties.remove();
   }
   
-  public static void safeSetProperties(String file) throws IOException {
+  public static void safeSetProperties(String... file) throws IOException {
```
hbs, line 129: Rename parameter since it can now contain multiple files.

warp10/src/main/java/io/warp10/WarpConfig.java
```diff
@@ -316,13 +327,16 @@ public static Properties readConfig(Reader reader, Properties properties, boolea
 
     br.close();
 
-    if (errorcount > 0) {
-      System.err.println("Aborting due to " + errorcount + " error" + (errorcount > 1 ? "s" : "") + ".");
-      System.exit(-1);
+    if (!linesInError.isEmpty()) {
+      throw new IOException("Malformed lines " + linesInError.toString() + ".");
```
hbs, line 331: Removing the System.exit call changes the behavior for callers other than the main method of WarpConfig, this will most likely break the behavior of many environment which initialize the Warp 10 config through this method, including the integration in Spark/Flink/Pig/KafkaStreams/NiFi and probably others.

hbs (changes requested):

warp10/src/main/java/io/warp10/WarpConfig.java
ftence: Changed the methods to take an extra parameter to do a `System.exit` or throw. If not specified, defaults to `System.exit` for backward-compatibility.
This parameter is the first in the methods because of the varargs and consistency between methods.

hbs (approved):

hbs:
Waiting for @steven-gueguen to review too.

steven-gueguen (approved):

PR #876, opened 2020-11-19
ALMOSTEQ (~=): Make conditions on parameters less strict

ftence:
One of the operand had to be a double, which is not required by the code. Removed that.

**Be careful**: the comparison with `lambda` was strict, which is not what users could expect. For instance, `1 1 0 ~=` would return `false`. Changed that also.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/ALMOSTEQ.java
+++ b/warp10/src/main/java/io/warp10/script/functions/ALMOSTEQ.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018  SenX S.A.S.
+//   Copyright 2018-2020  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/functions/ALMOSTEQ.java
```diff
@@ -40,26 +39,22 @@ public ALMOSTEQ(String name) {
...
 
-    // At least one of the operands must be a Double and both operands need to be numeric types
-    if  (!(op2 instanceof Number)  || !(op1 instanceof Number) || !(op2 instanceof Double || op1 instanceof Double)) {
-      throw new WarpScriptException(getName() + " can only operate on numeric types and at least one of them must be a Double.");  
-    }
-    
-    if (Double.isNaN(((Number) op1).doubleValue()) || Double.isNaN(((Number) op2).doubleValue())) {
-      stack.push(Double.isNaN(((Number) op1).doubleValue()) && Double.isNaN(((Number) op2).doubleValue()));
+    double lambda = Math.abs(((Number) lambdaParam).doubleValue());
+    double op1 = Math.abs(((Number) op1Param).doubleValue());
```
hbs, line 51: Why use Math.abs for computing values of op1 and op2?

ftence: My bad, fixed.

PR #878, opened 2020-11-26
Add capabilities to tokens

hbs:


warp10/src/main/java/io/warp10/script/ext/capabilities/CAPADD.java
```diff
@@ -0,0 +1,81 @@
...
+        capabilities = (Capabilities) stack.getAttribute(CapabilitiesWarpScriptExtension.CAPABILITIES_ATTR);
+      }
+
+      for (Entry<String,String> entry: attributes.entrySet()) {
+        if (entry.getKey().startsWith(CapabilitiesWarpScriptExtension.CAPABILITIES_PREFIX)) {
+          if (null == capabilities) {
+            capabilities = new Capabilities();
+            stack.setAttribute(CapabilitiesWarpScriptExtension.CAPABILITIES_ATTR, capabilities);
+          }
+          capabilities.capabilities.putIfAbsent(entry.getKey().substring(CapabilitiesWarpScriptExtension.CAPABILITIES_PREFIX.length()), entry.getValue());
```
ftence, line 74: Why it does not override already present capabilities?

warp10/src/main/java/io/warp10/script/ext/capabilities/CAPCHECK.java
```diff
@@ -0,0 +1,50 @@
...
+      Capabilities capabilities = null;
+      
+      if (stack.getAttribute(CapabilitiesWarpScriptExtension.CAPABILITIES_ATTR) instanceof Capabilities) {
+        capabilities = (Capabilities) stack.getAttribute(CapabilitiesWarpScriptExtension.CAPABILITIES_ATTR);
+        stack.push(capabilities.capabilities.containsKey((String) top));
+      } else {
+        stack.push(false);
+      }
+    } else {
+      stack.push(false);
```
ftence, line 45: A WarpScriptException would be helpful here.

warp10/src/main/java/io/warp10/script/ext/capabilities/CAPDEL.java
```diff
@@ -0,0 +1,65 @@
...
+      capabilities = (Capabilities) stack.getAttribute(CapabilitiesWarpScriptExtension.CAPABILITIES_ATTR);
+    }
+
+    if (top instanceof String) {
+      if (null != capabilities) {
+        capabilities.capabilities.remove((String) top);
+      }
+    } else if (top instanceof List) {
+      if (null != capabilities) {
+        if (((List) top).isEmpty()) {
```
ftence, line 49: I would find `NULL CAPDEL` to clear all capabilities to be less error-prone than `[] CAPDEL`: having an empty list could happen, having a NULL capability key is much less likely.

warp10/src/main/java/io/warp10/script/ext/capabilities/CAPGET.java
```diff
@@ -0,0 +1,50 @@
...
+    super(name);
+  }
+
+  @Override
+  public Object apply(WarpScriptStack stack) throws WarpScriptException {
+    Object top = stack.pop();
+
+    if (top instanceof String) {
+      stack.push(Capabilities.get(stack, (String) top));
+    } else if (top instanceof List) {
```
ftence, line 42: `top` could also be `null`, `Capabilities.get` would behave the same.

warp10/src/main/java/io/warp10/script/ext/capabilities/Capabilities.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Map;
+
+import io.warp10.script.WarpScriptStack;
+
+/**
+ * This class is there only to prevent the capabilities map from being accessible
+ * on the stack after a STACKATTRIBUTE call so capabilities cannot be added manually.
+ *
+ */
+public class Capabilities {
```
ftence, line 31: You should also add a `put` method, currently in `CAPADD` to be consistent.

ftence (changes requested):

warp10/src/main/java/io/warp10/script/ext/capabilities/CAPADD.java
hbs: Because if you override you cannot selectively add capabilities from a given token, wherease by combining CAPADD with the current behavior and CAPDEL you can select which capabilities will be (artificially since you actually delete them first) overridden.

warp10/src/main/java/io/warp10/script/ext/capabilities/Capabilities.java
hbs: Getting capabilities is something meant to be done by functions, adding (put) capabilities is only performed by CAPADD.

warp10/src/main/java/io/warp10/script/ext/capabilities/CAPGET.java
hbs: Changed the behavior so only null returns the full set of capabilities, if a list is passed, only those in the list are.

ftence (approved):

PR #879, opened 2020-11-26
Support dumping file based tokens when custom keys are not in use

hbs:


warp10/src/main/java/io/warp10/script/ext/token/TOKENDUMP.java
```diff
@@ -102,6 +104,7 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
       AESKey = (byte[]) top;
 
+      customKeys = true;
```
ftence, line 107: Thus `customKeys == multikey`. You could replace every use of `customKeys` by `multikey`.

warp10/src/main/java/io/warp10/script/ext/token/TOKENDUMP.java
```diff
@@ -85,6 +85,8 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
     Object top = null;
 
+    boolean customKeys = false;
```
ftence, line 88: Update copyright notice.

warp10/src/main/java/io/warp10/script/ext/token/TOKENDUMP.java
```diff
@@ -131,22 +134,34 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      QuasarTokenDecoder dec = decoder;
 
-    if (null == dec) {
-      long[] lkey = SipHashInline.getKey(SipHashKey);
-      dec = new QuasarTokenDecoder(lkey[0], lkey[1], AESKey);
-    }
+      if (null == dec) {
+        long[] lkey = SipHashInline.getKey(SipHashKey);
+        dec = new QuasarTokenDecoder(lkey[0], lkey[1], AESKey);
+      }
```
ftence, line 155: In case `customKeys` and thus `multikey`, `decoder` is always `null` so you can replace this block with 
```
long[] lkey = SipHashInline.getKey(SipHashKey);
QuasarTokenDecoder dec = new QuasarTokenDecoder(lkey[0], lkey[1], AESKey);
```

ftence (changes requested):

ftence:
Approved if and when https://github.com/hbs/warp10-platform/pull/9 is merged.

PR #882, opened 2020-12-04
Allow TOKENGEN and TOKENDUMP to be used always on custom keys without secret

ftence:
This PR allows the signatures
```
$map $aes $hash TOKENGEN
$token $aes $hash TOKENDUMP
```
to be used in any configuration, without secret.

If AES and SipHash keys are not specified, the behavior is the same as before, ie if keystore is that of a Warp/WarpDist instance, a secret is mandatory both in the conf and in the signatures, else no secret is necessary.

Also fixed a potential bug with non-thread safe `TSerializer` and improved error messages.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/ext/token/TOKENDUMP.java
+++ b/warp10/src/main/java/io/warp10/script/ext/token/TOKENDUMP.java
@@ -57,6 +57,12 @@ public class TOKENDUMP extends NamedWarpScriptFunction implements WarpScriptStac
    */
   private final boolean warpKeystore;
 
+  /**
+   * Create the TOKENDUMP function.
+   * @param name The name of the function.
+   * @param keystore The keystore containing the AES and SipHash keys to decode tokens when no such keys are given when applying this function.
+   * @param warpKeystore Whether the given keystore is that of a Warp/WarpDist instance. If true, a secret is needed to access the keystore keys.
+   */
   public TOKENDUMP(String name, KeyStore keystore, boolean warpKeystore) {
     super(name);
     if(null != keystore) {
--- a/warp10/src/main/java/io/warp10/script/ext/token/TOKENDUMP.java
+++ b/warp10/src/main/java/io/warp10/script/ext/token/TOKENDUMP.java
@@ -48,85 +48,93 @@ public class TOKENDUMP extends NamedWarpScriptFunction implements WarpScriptStac
 
   public static final String KEY_PARAMS = "params";
 
-  private byte[] tokenAESKey = null;
-  private byte[] tokenSipHashKey = null;
+  private final byte[] keystoreTokenAESKey;
+  private final byte[] keystoreTokenSipHashKey;
 
-  private boolean multikey = false;
+  /**
+   * Whether the keystore used to initialize the keys is that of a Warp/WarpDist instance.
+   * If true, a secret is needed to access the keystore keys.
+   */
+  private final boolean warpKeystore;
 
-  public TOKENDUMP(String name) {
+  public TOKENDUMP(String name, KeyStore keystore, boolean warpKeystore) {
     super(name);
-  }
-
-  public TOKENDUMP(String name, KeyStore keystore) {
-    super(name);
-    tokenAESKey = keystore.getKey(KeyStore.AES_TOKEN);
-    tokenSipHashKey = keystore.getKey(KeyStore.SIPHASH_TOKEN);
-  }
-
-  public TOKENDUMP(String name, boolean multikey) {
-    super(name);
-    this.multikey = multikey;
+    if(null != keystore) {
+      keystoreTokenAESKey = keystore.getKey(KeyStore.AES_TOKEN);
+      keystoreTokenSipHashKey = keystore.getKey(KeyStore.SIPHASH_TOKEN);
+    } else {
+      keystoreTokenAESKey = null;
+      keystoreTokenSipHashKey = null;
+    }
+    this.warpKeystore = warpKeystore;
   }
 
   @Override
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
+    byte[] tokenAESKey = null;
+    byte[] tokenSipHashKey = null;
+    boolean customKeys = false;
 
-    byte[] AESKey = tokenAESKey;
-    byte[] SipHashKey = tokenSipHashKey;
+    // First, check if SipHash and AES keys are explicitly defined. In that case, no need for secret.
+    Object top = stack.pop();
+    if (top instanceof byte[]) {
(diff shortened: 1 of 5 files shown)


warp10/src/main/java/io/warp10/script/ext/token/TOKENDUMP.java
```diff
@@ -48,85 +48,93 @@
...
-      //
-      String secret = TokenWarpScriptExtension.TOKEN_SECRET;
+    if (null == tokenAESKey) { // in that case we have also null == tokenSipHashKey
+      // SipHash and AES keys are not explicitly defined, so we fall back to those of the keystore.
+      // Check the secret if needed before the fallback.
+      if (warpKeystore) {
+        String secret = TokenWarpScriptExtension.TOKEN_SECRET;
 
-      top = stack.pop();
+        if (null == secret) {
```
hbs, line 102: You are implicitly making the definition of the secret mandatory, which was not the case previously, any reason for this change?

warp10/src/main/java/io/warp10/script/ext/token/TOKENDUMP.java
```diff
@@ -48,85 +48,93 @@
...
 
-  private boolean multikey = false;
+  /**
+   * Whether the keystore used to initialize the keys is that of a Warp/WarpDist instance.
+   * If true, a secret is needed to access the keystore keys.
+   */
+  private final boolean warpKeystore;
 
-  public TOKENDUMP(String name) {
+  public TOKENDUMP(String name, KeyStore keystore, boolean warpKeystore) {
```
hbs, line 60: Describe the purpose of warpKeystore

warp10/src/main/java/io/warp10/script/ext/token/TOKENGEN.java
```diff
@@ -57,79 +57,76 @@
...
-  private byte[] tokenSipHashKey = null;
-  private boolean multikey = false;
+  /**
+   * Whether the keystore used to initialize the keys is that of a Warp/WarpDist instance.
+   * If true, a secret is needed to access the keystore keys.
+   */
+  private final boolean warpKeystore;
 
-  public TOKENGEN(String name) {
+  public TOKENGEN(String name, KeyStore keystore, boolean warpKeystore) {
```
hbs, line 69: Describe the purpose of warpKeystore

warp10/src/main/java/io/warp10/script/ext/token/TOKENGEN.java
```diff
@@ -139,10 +136,10 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
     Map<Object, Object> params = (Map<Object, Object>) top;
 
     try {
-      TBase token = tokenFromMap(params);
+      TBase token = tokenFromMap(params, getName(), 0L);
```
hbs, line 139: You should leave the DEFAULT_TTL constant in the class and use it here instead of 0L

warp10/src/main/java/io/warp10/script/ext/token/TOKENGEN.java
```diff
@@ -241,23 +238,23 @@ public TBase tokenFromMap(Map params) throws WarpScriptException {
 
       if (null != params.get(KEY_ATTRIBUTES)) {
         if (!(params.get(KEY_ATTRIBUTES) instanceof Map)) {
-          throw new WarpScriptException(getName() + " expects '" + KEY_ATTRIBUTES + "' to be a map.");
+          throw new WarpScriptException(name + " expects '" + KEY_ATTRIBUTES + "' to be a map.");
         }
-        for (Entry<Object, Object> entry: ((Map<Object, Object>) params.get(KEY_ATTRIBUTES)).entrySet()) {
+        for (Entry<Object, Object> entry : ((Map<Object, Object>) params.get(KEY_ATTRIBUTES)).entrySet()) {
```
hbs, line 243: extraneous ' ' before ':'

warp10/src/main/java/io/warp10/script/ext/token/TOKENGEN.java
```diff
@@ -241,23 +238,23 @@ public TBase tokenFromMap(Map params) throws WarpScriptException {
...
         }
       }
 
       if (null != params.get(KEY_LABELS)) {
         if (!(params.get(KEY_LABELS) instanceof Map)) {
-          throw new WarpScriptException(getName() + " expects '" + KEY_LABELS + "' to be a map.");
+          throw new WarpScriptException(name + " expects '" + KEY_LABELS + "' to be a map.");
         }
-        for (Entry<Object, Object> entry: ((Map<Object, Object>) params.get(KEY_LABELS)).entrySet()) {
+        for (Entry<Object, Object> entry : ((Map<Object, Object>) params.get(KEY_LABELS)).entrySet()) {
```
hbs, line 255: extraneous ' ' before ':'

warp10/src/main/java/io/warp10/script/ext/token/TOKENGEN.java
```diff
@@ -191,17 +188,17 @@ public TBase tokenFromMap(Map params) throws WarpScriptException {
         }
       } else if (null != params.get(KEY_EXPIRY)) {
         rtoken.setExpiryTimestamp(((Number) params.get(KEY_EXPIRY)).longValue());
-      } else if (0 == DEFAULT_TTL) {
-        throw new WarpScriptException(getName() + " missing '" + KEY_TTL + "' or '" + KEY_EXPIRY + "'.");
+      } else if (0 == defaultTtl) {
```
hbs, line 191: 0L == defaultTTL (as later in the file)

hbs (changes requested):

warp10/src/main/java/io/warp10/script/ext/token/TOKENDUMP.java
ftence: This was already the case before but if was not clearly stated:

In TokenWarpScriptExtension.java when there was no secret in the conf nor OSS, the `keystore` was `null`:
`if (null == TOKEN_SECRET) { // if no configuration key` and `if (null != ossHandledSecret) {...} else { keystore = null; }`.

When the `keystore` was `null`, the `TokenWarpScriptExtension` constructor added the `TOKENGEN` and `TOKENDUMP` without keystore. It resulted in both keys to be `null` and `multikey` to be left to `false`. Then in `TOKENDUMP` we had
```
if ((null == AESKey || null == SipHashKey) && !this.multikey) {
      throw new WarpScriptException(getName() + " cannot be used in this context.");	
}
```
which resulted in an difficult-to-debug error message which in fact required the secret to be set.

PR #888, opened 2020-12-09
Only use on demand macro loading if refresh is set to 0

hbs:


ftence (approved):
I understand why WarpScriptMacroRepository is not instanciated anymore, the only minor downside is that this log message
```

hbs:
This is not really a downside since this message is only important when not using ondemand loading, because it could mean that some macros will not be available. Since ondemand is forced to true when refresh is set to 0, this situation cannot happen in this case.

PR #889, opened 2020-12-09
Fix current tick for mappers when GTS has duplicate ticks

ftence:


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -5484,11 +5493,17 @@ public static void reset(GeoTimeSerie gts) {
 
         for (int j = 0; j < subgts.values; j++) {
           ((Object[]) parms[6])[j] = valueAtIndex(subgts, j);
-          if (-1 == tickidx && tick == tickAtIndex(subgts, j)) {
+          if ((-1 == tickidx || reversed) && tick == tickAtIndex(subgts, j)) {
```
hbs, line 5496: Maybe add a comment why reversed is checked here, from the code it is not obvious.

hbs (changes requested):

randomboolean:
When using the parameter post>0 with duplicate ticks, there is a bug on the first ticks when the window is not yet at its widest.
For example, in the following warpscript, the first window has 2 "post ticks" with parameter post = 1.

```
NEWGTS
1 NaN NaN NaN 5 ADDVALUE
2 NaN NaN NaN 10 ADDVALUE
2 NaN NaN NaN 11 ADDVALUE
4 NaN NaN NaN 15 ADDVALUE
[ SWAP <%  VALUES ->JSON [ SWAP ] %> 1 1 0  ] MAP

[TOP] [{"c":"","l":{},"a":{},"la":0,"v":[[1,"[5,10,11]"],[2,"[5,10,11]"],[2,"[10,11,15]"],[4,"[10,11,15]"]]}]
```

There can be an argument for saying that this is not a bug, since there is a duplicate and that we can consider that ticks with the same timestamp do not count for the value of post. However if we follow that logic, the second window should also include the tick at timestamp 3, which is not the case.

There is a similar behavior for pre>0  at the end of the range.

Since this bug is about duplicate ticks and their ordering, maybe it should be discussed and fixed in this PR ?

ftence:
Yes, there's a bug for pre>0 or post>0 with duplicate ticks. It is not restricted to the first or last window. However, it is unclear how MAP should work on duplicate ticks for those parameters, it's something we have to discuss. I suggest we keep this PR on the subject of the current tick index and find a way to fix the other bug later.

PR #891, opened 2020-12-18
Http extension

randomboolean:
Add the functions HTTP, MAXURLCOUNT, MAXDOWNLOADSIZE

{
  'method' 'POST'
  'url' 'http://localhost:8080/api/v0/exec'
  'header' {} // optional, default to empty map
  'body' '1 2 3 4' // optional, default to empty string
}
HTTP

HTTP is based on URLFETCH. It adds the other methods and possibility to configure a capability to control the usage of the HTTP function.

### DIFF
--- a/etc/conf.templates/distributed/70--extensions.conf.template
+++ b/etc/conf.templates/distributed/70--extensions.conf.template
@@ -58,6 +58,8 @@
 #token.secret = ....
 // URLFETCH, MAXURLFETCHCOUNT, MAXURLFETCHSIZE
 #warpscript.extension.urlfetch = io.warp10.script.ext.urlfetch.UrlFetchWarpScriptExtension
+// HTTP, MAXURLCOUNT, MAXDOWNLOADSIZE
+warpscript.extension.http = io.warp10.script.ext.http.HttpWarpScriptExtension
 // WSPS WSINFO WSKILL STACKPSSECRET
 #warpscript.extension.stackps = io.warp10.script.ext.stackps.StackPSWarpScriptExtension
 #stackps.secret = ....
--- a/etc/conf.templates/standalone/70--extensions.conf.template
+++ b/etc/conf.templates/standalone/70--extensions.conf.template
@@ -58,6 +58,8 @@
 #token.secret = ....
 // URLFETCH, MAXURLFETCHCOUNT, MAXURLFETCHSIZE
 #warpscript.extension.urlfetch = io.warp10.script.ext.urlfetch.UrlFetchWarpScriptExtension
+// HTTP, MAXURLCOUNT, MAXDOWNLOADSIZE
+warpscript.extension.http = io.warp10.script.ext.http.HttpWarpScriptExtension
 // DEVAL
 #warpscript.extension.sharding = io.warp10.script.ext.sharding.ShardingWarpScriptExtension
 // WSPS WSINFO WSKILL STACKPSSECRET
--- a/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
+++ b/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
@@ -29,6 +29,7 @@ import org.apache.commons.codec.binary.Base64;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -188,8 +189,14 @@ public class HTTP extends FormattedWarpScriptFunction {
       }
 
       conn.setDoInput(true);
-      conn.setDoOutput(false);
-      conn.setRequestMethod("GET");
+      conn.setDoOutput(body.length() > 0);
+      conn.setRequestMethod(method.toUpperCase());
+
+      if (body.length() > 0) {
+        try (OutputStream os = conn.getOutputStream()) {
+          os.write(body.getBytes());
+        }
+      }
 
       byte[] buf = new byte[8192];
 
@@ -248,7 +255,7 @@ public class HTTP extends FormattedWarpScriptFunction {
       res.add(Base64.encodeBase64String(baos.toByteArray()));
 
     } catch (IOException ioe) {
-      throw new WarpScriptException(getName() + " encountered an error while fetching '" + url + "'", ioe);
+      throw new WarpScriptException(getName() + " encountered an error while making an HTTP " + method + " request over '" + url + "'", ioe);
     } finally {
       if (null != conn) {
         conn.disconnect();
--- a/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
+++ b/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
@@ -22,6 +22,9 @@ import io.warp10.script.WarpScriptStack;
 import io.warp10.script.WebAccessController;
 import io.warp10.script.formatted.FormattedWarpScriptFunction;
 import io.warp10.standalone.StandaloneWebCallService;
+import io.warp10.warp.sdk.Capabilities;
+import io.warp10.script.ext.http.HttpWarpScriptExtension;
+
 import org.apache.commons.codec.binary.Base64;
 
 import java.io.ByteArrayOutputStream;
@@ -83,8 +86,11 @@ public class HTTP extends FormattedWarpScriptFunction {
   @Override
   public WarpScriptStack apply(Map<String, Object> formattedArgs, WarpScriptStack stack) throws WarpScriptException {
 
-    if (!stack.isAuthenticated()) {
-      throw new WarpScriptException(getName() + " requires the stack to be authenticated.");
+    if (stack.getAttribute(WarpScriptStack.CAPABILITIES_ATTR) instanceof Capabilities) {
+      Capabilities capabilities = (Capabilities) stack.getAttribute(WarpScriptStack.CAPABILITIES_ATTR);
+      if(!(capabilities.containsKey(HttpWarpScriptExtension.HTTP_CAPABILITY))) {
+        throw new WarpScriptException("Capability " + HttpWarpScriptExtension.HTTP_CAPABILITY + " is required by function " + getName());
+      }
     }
 
     Object o = stack.pop();
--- a/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
+++ b/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
@@ -17,7 +17,6 @@
 package io.warp10.script.ext.http;
 
 import io.warp10.WarpConfig;
-import io.warp10.continuum.Configuration;
 import io.warp10.script.WarpScriptException;
 import io.warp10.script.WarpScriptStack;
 import io.warp10.script.WebAccessController;
--- a/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
+++ b/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
@@ -17,6 +17,7 @@
 package io.warp10.script.ext.http;
 
 import io.warp10.WarpConfig;
+import io.warp10.continuum.Configuration;
 import io.warp10.script.WarpScriptException;
 import io.warp10.script.WarpScriptStack;
 import io.warp10.script.WebAccessController;
@@ -38,6 +39,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -107,8 +109,10 @@ public class HTTP extends FormattedWarpScriptFunction {
     //
 
     if (stack.getAttribute(WarpScriptStack.CAPABILITIES_ATTR) instanceof Capabilities) {
+      Properties props = WarpConfig.getProperties();
       Capabilities capabilities = (Capabilities) stack.getAttribute(WarpScriptStack.CAPABILITIES_ATTR);
-      if(!(capabilities.containsKey(HttpWarpScriptExtension.HTTP_CAPABILITY))) {
+
+      if(props.containsKey(HttpWarpScriptExtension.HTTP_CAPABILITY) && !(capabilities.containsKey(props.getProperty(HttpWarpScriptExtension.HTTP_CAPABILITY)))) {
         throw new WarpScriptException("Capability " + HttpWarpScriptExtension.HTTP_CAPABILITY + " is required by function " + getName());
       }
     }
--- a/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
+++ b/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
@@ -17,11 +17,10 @@
 package io.warp10.script.ext.http;
 
 import io.warp10.WarpConfig;
-import io.warp10.script.NamedWarpScriptFunction;
 import io.warp10.script.WarpScriptException;
 import io.warp10.script.WarpScriptStack;
-import io.warp10.script.WarpScriptStackFunction;
 import io.warp10.script.WebAccessController;
+import io.warp10.script.formatted.FormattedWarpScriptFunction;
 import io.warp10.standalone.StandaloneWebCallService;
 import org.apache.commons.codec.binary.Base64;
 
@@ -42,7 +41,18 @@ import java.util.concurrent.locks.ReentrantLock;
 /**
  * Apply an HTTP method over an url
  */
-public class HTTP extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+public class HTTP extends FormattedWarpScriptFunction {
+
+
+  private final Arguments args;
+  private final Arguments output;
+
+  protected Arguments getArguments() {
+    return args;
+  }
+  protected Arguments getOutput() {
+    return output;
+  }
 
   private final ReentrantLock stackCountersLock = new ReentrantLock();
   private final WebAccessController webAccessController;
@@ -58,10 +68,20 @@ public class HTTP extends NamedWarpScriptFunction implements WarpScriptStackFunc
     } else {
       webAccessController = new WebAccessController(patternConf);
     }
+
+    getDocstring().append("Apply an HTTP method over an url and fetch response.");
+
+    args = new ArgumentsBuilder()
+      //todo
+      .build();
+
+    output = new ArgumentsBuilder()
+      //todo
+      .build();
   }
 
   @Override
-  public Object apply(WarpScriptStack stack) throws WarpScriptException {
+  public WarpScriptStack apply(Map<String, Object> formattedArgs, WarpScriptStack stack) throws WarpScriptException {
 
     if (!stack.isAuthenticated()) {
       throw new WarpScriptException(getName() + " requires the stack to be authenticated.");
--- a/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
+++ b/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
@@ -14,7 +14,7 @@
 //   limitations under the License.
 //
 
-package io.warp10.script.ext.urlfetch;
+package io.warp10.script.ext.http;
 
 import io.warp10.WarpConfig;
 import io.warp10.script.NamedWarpScriptFunction;
@@ -40,17 +40,17 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReentrantLock;
 
 /**
- * Fetch content from a URL
+ * Apply an HTTP method over an url
  */
-public class URLFETCH extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+public class HTTP extends NamedWarpScriptFunction implements WarpScriptStackFunction {
 
   private final ReentrantLock stackCountersLock = new ReentrantLock();
   private final WebAccessController webAccessController;
 
-  public URLFETCH(String name) {
+  public HTTP(String name) {
     super(name);
 
-    String patternConf = WarpConfig.getProperty(UrlFetchWarpScriptExtension.WARPSCRIPT_URLFETCH_HOST_PATTERNS);
+    String patternConf = WarpConfig.getProperty(HttpWarpScriptExtension.WARPSCRIPT_HTTP_HOST_PATTERNS);
 
     // If not defined, use already existing StandaloneWebCallService webAccessController which uses Configuration.WEBCALL_HOST_PATTERNS
     if (null == patternConf) {
@@ -104,27 +104,27 @@ public class URLFETCH extends NamedWarpScriptFunction implements WarpScriptStack
     }
 
     //
-    // Check that we do not exceed the maxurlfetch limit
+    // Check that we do not exceed the limits
     //
 
     // Get the current counters in the stack and initialize them if not present.
-    AtomicLong urlfetchCount;
-    AtomicLong urlfetchSize;
+    AtomicLong urlCount;
+    AtomicLong downloadSize;
 
     try {
       stackCountersLock.lockInterruptibly();
 
-      Object ufCount = stack.getAttribute(UrlFetchWarpScriptExtension.ATTRIBUTE_URLFETCH_COUNT);
-      Object ufSize = stack.getAttribute(UrlFetchWarpScriptExtension.ATTRIBUTE_URLFETCH_SIZE);
+      Object ufCount = stack.getAttribute(HttpWarpScriptExtension.ATTRIBUTE_HTTP_COUNT);
+      Object ufSize = stack.getAttribute(HttpWarpScriptExtension.ATTRIBUTE_HTTP_SIZE);
 
       if (null == ufCount || null == ufSize) {
-        urlfetchCount = new AtomicLong();
-        urlfetchSize = new AtomicLong();
-        stack.setAttribute(UrlFetchWarpScriptExtension.ATTRIBUTE_URLFETCH_COUNT, urlfetchCount);
-        stack.setAttribute(UrlFetchWarpScriptExtension.ATTRIBUTE_URLFETCH_SIZE, urlfetchSize);
+        urlCount = new AtomicLong();
+        downloadSize = new AtomicLong();
+        stack.setAttribute(HttpWarpScriptExtension.ATTRIBUTE_HTTP_COUNT, urlCount);
+        stack.setAttribute(HttpWarpScriptExtension.ATTRIBUTE_HTTP_SIZE, downloadSize);
       } else {
-        urlfetchCount = (AtomicLong) ufCount;
-        urlfetchSize = (AtomicLong) ufSize;
+        urlCount = (AtomicLong) ufCount;
+        downloadSize = (AtomicLong) ufSize;
       }
     } catch (InterruptedException ie) {
       throw new WarpScriptException(getName() + " thread has been interrupted", ie);
@@ -134,16 +134,16 @@ public class URLFETCH extends NamedWarpScriptFunction implements WarpScriptStack
       }
     }
 
-    if (urlfetchCount.get() + urls.size() > (long) UrlFetchWarpScriptExtension.getLongAttribute(stack, UrlFetchWarpScriptExtension.ATTRIBUTE_URLFETCH_LIMIT)) {
-      throw new WarpScriptException(getName() + " is limited to " + UrlFetchWarpScriptExtension.getLongAttribute(stack, UrlFetchWarpScriptExtension.ATTRIBUTE_URLFETCH_LIMIT) + " calls.");
+    if (urlCount.get() + urls.size() > (long) HttpWarpScriptExtension.getLongAttribute(stack, HttpWarpScriptExtension.ATTRIBUTE_HTTP_LIMIT)) {
+      throw new WarpScriptException(getName() + " is limited to " + HttpWarpScriptExtension.getLongAttribute(stack, HttpWarpScriptExtension.ATTRIBUTE_HTTP_LIMIT) + " calls.");
     }
 
     List<Object> results = new ArrayList<Object>();
 
     for (URL url: urls) {
       // Recheck the count here in case of concurrent runs
-      if (urlfetchCount.addAndGet(1) > (long) UrlFetchWarpScriptExtension.getLongAttribute(stack, UrlFetchWarpScriptExtension.ATTRIBUTE_URLFETCH_LIMIT)) {
-        throw new WarpScriptException(getName() + " is limited to " + UrlFetchWarpScriptExtension.getLongAttribute(stack, UrlFetchWarpScriptExtension.ATTRIBUTE_URLFETCH_LIMIT) + " calls.");
+      if (urlCount.addAndGet(1) > (long) HttpWarpScriptExtension.getLongAttribute(stack, HttpWarpScriptExtension.ATTRIBUTE_HTTP_LIMIT)) {
+        throw new WarpScriptException(getName() + " is limited to " + HttpWarpScriptExtension.getLongAttribute(stack, HttpWarpScriptExtension.ATTRIBUTE_HTTP_LIMIT) + " calls.");
       }
 
       HttpURLConnection conn = null;
@@ -183,14 +183,14 @@ public class URLFETCH extends NamedWarpScriptFunction implements WarpScriptStack
             break;
           }
 
-          if (urlfetchSize.get() + baos.size() + len > (long) UrlFetchWarpScriptExtension.getLongAttribute(stack, UrlFetchWarpScriptExtension.ATTRIBUTE_URLFETCH_MAXSIZE)) {
-            throw new WarpScriptException(getName() + " would exceed maximum size of content which can be retrieved via URLFETCH (" + UrlFetchWarpScriptExtension.getLongAttribute(stack, UrlFetchWarpScriptExtension.ATTRIBUTE_URLFETCH_MAXSIZE) + " bytes)");
+          if (downloadSize.get() + baos.size() + len > (long) HttpWarpScriptExtension.getLongAttribute(stack, HttpWarpScriptExtension.ATTRIBUTE_HTTP_MAXSIZE)) {
+            throw new WarpScriptException(getName() + " would exceed maximum size of content which can be retrieved via this function (" + HttpWarpScriptExtension.getLongAttribute(stack, HttpWarpScriptExtension.ATTRIBUTE_HTTP_MAXSIZE) + " bytes)");
           }
 
           baos.write(buf, 0, len);
         }
 
-        urlfetchSize.addAndGet(baos.size());
+        downloadSize.addAndGet(baos.size());
 
         List<Object> res = new ArrayList<Object>();
 
--- a/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
+++ b/warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
@@ -23,7 +23,6 @@ import io.warp10.script.WebAccessController;
 import io.warp10.script.formatted.FormattedWarpScriptFunction;
 import io.warp10.standalone.StandaloneWebCallService;
 import io.warp10.warp.sdk.Capabilities;
-import io.warp10.script.ext.http.HttpWarpScriptExtension;
 
 import org.apache.commons.codec.binary.Base64;
 
@@ -46,6 +45,15 @@ import java.util.concurrent.locks.ReentrantLock;
  */
 public class HTTP extends FormattedWarpScriptFunction {
 
+  //
+  // Arguments
+  //
+
+  public static final String METHOD = "method";
+  public static final String URL = "url";
+  public static final String HEADER = "header";
+  public static final String BODY = "body";
+  public static final String RESPONSE = "response";
 
   private final Arguments args;
   private final Arguments output;
@@ -56,6 +64,10 @@ public class HTTP extends FormattedWarpScriptFunction {
   protected Arguments getOutput() {
     return output;
   }
+
+  //
+  // Control
+  //
 
   private final ReentrantLock stackCountersLock = new ReentrantLock();
   private final WebAccessController webAccessController;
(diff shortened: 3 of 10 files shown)


randomboolean:
Network loops are not prevented in the current implementation. Maybe we should store authentication tokens in a shared collection to avoid a token being used to make nested HTTP calls ?

etc/conf.templates/distributed/70--extensions.conf.template
```diff
@@ -58,6 +58,8 @@
 #token.secret = ....
 // URLFETCH, MAXURLFETCHCOUNT, MAXURLFETCHSIZE
 #warpscript.extension.urlfetch = io.warp10.script.ext.urlfetch.UrlFetchWarpScriptExtension
+// HTTP, MAXURLCOUNT, MAXDOWNLOADSIZE
+warpscript.extension.http = io.warp10.script.ext.http.HttpWarpScriptExtension
```
hbs, line 62: Must be commented by default.

etc/conf.templates/standalone/70--extensions.conf.template
```diff
@@ -58,6 +58,8 @@
 #token.secret = ....
 // URLFETCH, MAXURLFETCHCOUNT, MAXURLFETCHSIZE
 #warpscript.extension.urlfetch = io.warp10.script.ext.urlfetch.UrlFetchWarpScriptExtension
+// HTTP, MAXURLCOUNT, MAXDOWNLOADSIZE
+warpscript.extension.http = io.warp10.script.ext.http.HttpWarpScriptExtension
```
hbs, line 62: Must be commented by default.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
+//
+//   Copyright 2021  SenX S.A.S.
```
hbs, line 2: 2020

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * Apply an HTTP method over an url
+ */
+public class HTTP extends FormattedWarpScriptFunction {
```
hbs, line 48: Use extends NamedWarpScriptFunction implements WarpScriptStackFunction for clarity. The FormattedWarpScriptFunction approach does cripple the overall readability and may be removed in future versions of Warp 10.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+      webAccessController = StandaloneWebCallService.getWebAccessController();
+    } else {
+      webAccessController = new WebAccessController(patternConf);
+    }
+
+    getDocstring().append("Apply an HTTP method over an url and fetch response.");
+
+    args = new ArgumentsBuilder()
+      .addArgument(String.class, METHOD, "The http method.")
+      .addArgument(String.class, URL, "The URL to send the GET request to. Must begin with http:// or https://.")
```
hbs: The documentation will be elsewhere but this is no longer limited to the GET method.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+      .addOptionalArgument(Map.class, HEADER, "An optional header.", new HashMap<>())
+      .addOptionalArgument(String.class, BODY, "An optional body", "")
+      .build();
+
+    output = new ArgumentsBuilder()
+      .addArgument(List.class, RESPONSE, "A list containing, for each URL, a 4-element list. These 4-element lists contain, in this order, a LONG status code, a STRING status message or an empty STRING if not available, a MAP of headers and a STRING representing a bytes array encoded as base 64.")
```
hbs: The content must be returned as BYTES, not base 64 encoded as STRING. And returning a MAP is probably wiser.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  @Override
+  public WarpScriptStack apply(Map<String, Object> formattedArgs, WarpScriptStack stack) throws WarpScriptException {
+
+    //
+    // Check capability if set in configuration, or else check stack authentication
+    //
+
+    Properties props = WarpConfig.getProperties();
```
hbs, line 110: This must not be called at each invocation. Extract the capability name once at instantiation time and refer to that in apply.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+
+    if (props.containsKey(HttpWarpScriptExtension.HTTP_CAPABILITY)) {
+      if (stack.getAttribute(WarpScriptStack.CAPABILITIES_ATTR) instanceof Capabilities) {
```
hbs, line 113: Capabilities are checked using Capabilities#get

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+        Capabilities capabilities = (Capabilities) stack.getAttribute(WarpScriptStack.CAPABILITIES_ATTR);
+
+        if (!(capabilities.containsKey(props.getProperty(HttpWarpScriptExtension.HTTP_CAPABILITY)))) {
+          throw new WarpScriptException("Capability " + HttpWarpScriptExtension.HTTP_CAPABILITY + " is required by function " + getName());
+        }
+      }
+
+    } else if (!stack.isAuthenticated()) {
```
hbs, line 121: The authorization type should be configurable, either none, i.e. the function can be used by anyone if the extension is loaded, authenticated, AUTHENTICATE must have been called, and capability to inspect a capability.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+    //
+    // Check that we do not exceed the limits
+    //
+
+    // Get the current counters in the stack and initialize them if not present.
+    AtomicLong urlCount;
+    AtomicLong downloadSize;
+
+    try {
+      stackCountersLock.lockInterruptibly();
```
hbs, line 153: We don't need synchronization, the rare cases where the count would be incorrect (use in CEVAL) don't need to be handled as their consequence is not important.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+      if (stackCountersLock.isHeldByCurrentThread()) {
+        stackCountersLock.unlock();
+      }
+    }
+
+    if (urlCount.get() + 1 > (long) HttpWarpScriptExtension.getLongAttribute(stack, HttpWarpScriptExtension.ATTRIBUTE_HTTP_LIMIT)) {
+      throw new WarpScriptException(getName() + " is limited to " + HttpWarpScriptExtension.getLongAttribute(stack, HttpWarpScriptExtension.ATTRIBUTE_HTTP_LIMIT) + " calls.");
+    }
+
+    // Recheck the count here in case of concurrent runs
```
hbs, line 179: Simply do the check with  addAndGet(1), it is an AtomicLong

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    HttpURLConnection conn = null;
+
+    List<Object> res = new ArrayList<Object>();
+
+    try {
+      conn = (HttpURLConnection) url.openConnection();
+
+      if (null != url.getUserInfo()) {
```
hbs, line 191: I think it it better to disallow user infos in the URL and to provide a macro to encode basic authentication which can then be passed in the headers.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+    } else if (!stack.isAuthenticated()) {
+      throw new WarpScriptException(getName() + " requires the stack to be authenticated.");
+    }
+
+    //
+    // Retrieve arguments
+    //
+
+    String method = (String) formattedArgs.get(METHOD);
+    Map<Object, Object> properties = (Map) formattedArgs.get(HEADER);
```
hbs, line 130: Call them headers

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+    String body = (String) formattedArgs.get(BODY);
```
hbs, line 131: the body parameter could contain a STRING or BYTES to let the user choose the charset used.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+        String basicAuth = "Basic " + new String(Base64.encodeBase64String(url.getUserInfo().getBytes(StandardCharsets.UTF_8)));
+        properties.put("Authorization", basicAuth);
+      }
+
+      for (Map.Entry<Object, Object> prop: properties.entrySet()) {
+        conn.setRequestProperty(String.valueOf(prop.getKey()), String.valueOf(prop.getValue()));
+      }
+
+      conn.setDoInput(true);
+      conn.setDoOutput(body.length() > 0);
```
hbs, line 201: body could be a STRING or BYTES

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+      conn.setRequestMethod(method.toUpperCase());
+
+      if (body.length() > 0) {
+        try (OutputStream os = conn.getOutputStream()) {
+          os.write(body.getBytes());
```
hbs, line 206: Use StandardCharsets.UTF8 explicitely

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,278 @@
...
+      //
+
+      hdrs = new HashMap<String, List<String>>(hdrs);
+      hdrs.remove(null);
+
+      res.add(hdrs);
+      res.add(Base64.encodeBase64String(baos.toByteArray()));
+
+    } catch (IOException ioe) {
+      throw new WarpScriptException(getName() + " encountered an error while making an HTTP " + method + " request over '" + url + "'", ioe);
```
hbs, line 267: request to

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
```diff
@@ -0,0 +1,172 @@
...
+import io.warp10.WarpConfig;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.warp.sdk.WarpScriptExtension;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Extension for HTTP and the associated function to change limits: MAXURLCOUNT and MAXDOWNLOADSIZE
```
hbs, line 28: These functions do not need to exist. Rely on capabilities to modify the limits.

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
```diff
@@ -0,0 +1,172 @@
...
+  public static final String HTTP_CAPABILITY = "warpscript.http.capability";
+
+  //
+  // CONFIGURATION
+  //
+
+  /**
+   * Maximum number of calls to HTTP in a session
+   */
+  public static final String WARPSCRIPT_HTTP_LIMIT = "warpscript.http.limit";
```
hbs, line 48: change the name to maxrequests

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
```diff
@@ -0,0 +1,172 @@
...
+  public static final String WARPSCRIPT_HTTP_LIMIT_HARD = "warpscript.http.limit.hard";
```
hbs, line 49: remove the hard limit and use token capabilities.

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
```diff
@@ -0,0 +1,172 @@
...
+
+  /**
+   * Maximum cumulative size of content retrieved via calls to HTTP in a session
+   */
+  public static final String WARPSCRIPT_HTTP_MAXSIZE = "warpscript.http.maxsize";
+  public static final String WARPSCRIPT_HTTP_MAXSIZE_HARD = "warpscript.http.maxsize.hard";
```
hbs, line 55: remove hard limit

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
```diff
@@ -0,0 +1,172 @@
...
+  /**
+   * Maximum number of calls to HTTP in a session
+   */
+  public static final String ATTRIBUTE_HTTP_LIMIT = "http.limit";
+  public static final String ATTRIBUTE_HTTP_LIMIT_HARD = "http.limit.hard";
+
+  /**
+   * Number of calls to HTTP so far in the sessions
+   */
+  public static final String ATTRIBUTE_HTTP_COUNT = "http.count";
```
hbs, line 75: rename to http.requests

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
```diff
@@ -0,0 +1,172 @@
...
+  public static final String WARPSCRIPT_HTTP_HOST_PATTERNS = "warpscript.http.host.patterns";
+
+  //
+  // STACK
+  //
+
+  /**
+   * Maximum number of calls to HTTP in a session
+   */
+  public static final String ATTRIBUTE_HTTP_LIMIT = "http.limit";
```
hbs, line 69: rename to http.maxrequests

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
```diff
@@ -0,0 +1,172 @@
...
+  /**
+   * Number of calls to HTTP so far in the sessions
+   */
+  public static final String ATTRIBUTE_HTTP_COUNT = "http.count";
+
+  /**
+   * Maximum cumulative size of content retrieved via calls to HTTP in a session
+   */
+  public static final String ATTRIBUTE_HTTP_MAXSIZE = "http.maxsize";
+  public static final String ATTRIBUTE_HTTP_MAXSIZE_HARD = "http.maxsize.hard";
```
hbs, line 81: remove hard limit

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
```diff
@@ -0,0 +1,172 @@
...
+  /**
+   * Current  HTTP so far in the sessions
+   */
+  public static final String ATTRIBUTE_HTTP_SIZE = "http.size";
+
+  //
+  // DEFAULTS
+  //
+
+  public static final long DEFAULT_HTTP_LIMIT = 64L;
```
hbs, line 92: this is too high, limit to 1 so users explicitly set the configuration.

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
```diff
@@ -0,0 +1,172 @@
...
+  public static final long DEFAULT_HTTP_MAXSIZE = 1000000L;
```
hbs, line 93: Lower that to 65536

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
```diff
@@ -0,0 +1,172 @@
...
+
+  /**
+   * Get the value of the attribute in the stack, if not present get the value in the configuration and it not present either, get the default value.
+   * Also set the attribute in the stack once the value is found.
+   *
+   * @param stack     The stack the get the attribute from, if present.
+   * @param attribute The attribute name to get.
+   * @return The first available value in the list: attribute value, configuration value, default.
+   */
+  public static Long getLongAttribute(WarpScriptStack stack, String attribute) {
```
hbs, line 148: Remove this. We can consider adding a method to StackUtils to enforce a similar logic but with a simpler behavior, i.e. no attributeToConf structure but simply an extraneous parameter to the call.

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
```diff
@@ -0,0 +1,172 @@
+//
+//   Copyright 2021  SenX S.A.S.
```
hbs, line 2: 2020

hbs (changes requested):

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
randomboolean: I think after knowing how to use the FormattedWarpScriptFunction class it is the opposite. I find it handy to declare functions with multiple arguments to retrieve from the stack and handle defaults, while also hiding the stack syntax particularity. Also it is convenient to generate documentation files.

Maybe the logic of this class can be transferred to a stack method: stack.retrieveArguments(args)
 And documentation generation to a function of args, output and function.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
randomboolean: I was thinking of enforcing authentication in each case (capability or not), since it allows for checking if another thread in the jvm has a stack that is authenticated with the same token and also try to use HTTP, so that we can prevent nested call of HTTP and network loops.

randomboolean:
The name of the extension might be conflicting with the plugin with the same name. Also it also support https protocol.
For these reasons, we may want to change its name. Http-requests ? requests ? Or we can also keep the current name.

warp10/src/main/java/io/warp10/script/ext/http/HttpWarpScriptExtension.java
randomboolean: Limit has been set to 1L. But since auth_macro may need to make requests as well (e.g. digest authentication), should we raise this default to a few more ?

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,437 @@
...
+          if (len < 0) {
+            break;
+          }
+
+          if (downloadSize.addAndGet(len) > maxsize) {
+            throw new WarpScriptException(getName() + " would exceed maximum size of content which can be retrieved via this function (" + maxsize + " bytes)");
+          }
+
+          if (len == chunkSize) {
+            chunkRes.put(CONTENT, buf);
```
ftence, line 407: You should always make a copy.
This isn't great performance-wise but if someone leave the byte[] on the stack, it will be modified and will confuse the user.
It may be OK to reuse the same Map but I think this should be discussed.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,437 @@
...
+            for (int i = 0; i < buf2.length; i++) {
+              buf2[i] = buf[i];
+            }
+            chunkRes.put(CONTENT, buf2);
+          }
+          chunkRes.put(CHUNK_NUMBER, new Long(chunkNumber));
+          stack.push(chunkRes);
+          if (null != chunkMacro) {
+            stack.exec(chunkMacro);
+          }
```
ftence, line 419: If you make the macro optional, the content Map must be a copy, like discussed above.

ftence (changes requested):

ftence:
Nice PR and nice use of capabilities. :+1:

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,430 @@
...
+
+        headers.putAll(additionalHeaders);
+      }
+
+      for (Map.Entry<Object, Object> prop: headers.entrySet()) {
+        conn.setRequestProperty(String.valueOf(prop.getKey()), String.valueOf(prop.getValue()));
+      }
+
+      conn.setDoInput(true);
+      conn.setRequestMethod(method.toUpperCase());
```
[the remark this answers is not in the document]
randomboolean: IMO basic authentication is so common that saving a few lines and hiding the encoding is worth it

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,442 @@
...
+    // Retrieve arguments
+    //
+
+    String method = (String) params.get(METHOD);
+    if (null == method) {
+      throw new WarpScriptException(getName() + " expects an http method.");
+    }
+
+    Map<Object, Object> headers = (Map) params.getOrDefault(HEADERS, new HashMap<>());
+    Object body = params.getOrDefault(BODY,"");
```
ftence, line 194: You can simply use `get`, which will default to `null`, thus skipping both `instanceof` conditionals.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,442 @@
...
+      // Encode userinfo and set headers
+      //
+
+      if (null != authInfo) {
+
+        Map additionalHeaders;
+        if (null != authMacro) {
+          stack.push(authInfo);
+          stack.exec(authMacro);
+          additionalHeaders = (Map) stack.pop();
```
ftence, line 266: I think the macro.auth mechanism will not add any value to the HTTP method. For instance, it will not help in generating the digest auth authorization header, especially when doing multiple authenticated requests.

ftence (changes requested):
It could be nice to be able to change the user-agent either in the configuration or with the appropriate authorization (authenticated of capability) to do so.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
randomboolean: I don't see why it can not be used to generate digest auth. But I see it can be painful to write the auth macro for digest auth without another extension (one could wraps another java function, like for example [DigestScheme.authenticate](https://www.javadoc.io/static/org.apache.httpcomponents/httpclient/4.5.3/org/apache/http/impl/auth/DigestScheme.html#authenticate(org.apache.http.auth.Credentials,%20org.apache.http.HttpRequest,%20org.apache.http.protocol.HttpContext)))..

randomboolean:
> It could be nice to be able to change the user-agent either in the configuration or with the appropriate authorization (authenticated of capability) to do so.

It can be set in the headers.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Send an HTTP request to an url
```
hbs, line 45: to a URL

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+ *
+ * To raise maximum number of calls and download size limit, use these capabilities:
+ * http.requests
+ * http.size
+ *
+ * Params:
+ * METHOD The http method
```
hbs, line 52: The param names are in lower case

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+ * URL The URL to send the request to. Must begin with http:// or https://
+ * HEADERS An optional header
```
hbs, line 54: header map

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+ * BODY An optional body. STRING or BYTES
```
hbs, line 55: Specify the charset used if the body is a STRING

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+ * AUTH_INFO Authentication arguments. For example for basic authentication, provide [username, password]
```
hbs, line 56: Since this ends up being a header, maybe it is better to provide a helper function or macro to pack username and password as a value suitable for a header

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+ * AUTH_MACRO A macro that expects " + AUTH_INFO + " on the stack, and returns a map to be appended with the headers. Default to basic authentication
```
hbs, line 57: see above

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+ * CHUNK_SIZE Chunk size
```
hbs, line 58: The chunk size should have a configurable upper limit to avoid giant allocations

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+  // Limits
+  //
+
+  private final long baseMaxRequests;
+  private final long baseMaxSize;
+
+  public HTTP(String name) {
+    super(name);
+
+    String patternConf = WarpConfig.getProperty(HttpWarpScriptExtension.WARPSCRIPT_HTTP_HOST_PATTERNS);
```
hbs, line 114: Parameter extraction should be done in a static way so other instances of HTTP do not repeat it.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+    //
+    // Check authorization
+    //
+
+    if (auth && !stack.isAuthenticated()) {
+      throw new WarpScriptException(getName() + " requires the stack to be authenticated.");
+    }
+
+    if (null != capName && null == Capabilities.get(stack, capName)) {
+      throw new WarpScriptException("Capability " + capName + " is required by function " + getName());
```
hbs, line 164: missing function name

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+    } else {
+      maxsize = baseMaxSize;
+    }
+
+    //
+    // Retrieve arguments
+    //
+
+    String method = (String) params.get(METHOD);
+    if (null == method) {
```
hbs, line 190: could default to GET

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+      throw new WarpScriptException(getName() + " expects an http method.");
+    }
+
+    Map<Object, Object> headers = (Map) params.getOrDefault(HEADERS, new HashMap<>());
+    Object body = params.get(BODY);
+
+    List authInfo = (List) params.get(AUTH_INFO);
+    WarpScriptStack.Macro authMacro = (WarpScriptStack.Macro) params.get(AUTH_MACRO);
```
hbs, line 198: You are not checking if the parameter is a Macro, this will throw an exception if that is not the case

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+    Long chunkSize = (Long) params.getOrDefault(CHUNK_SIZE, -1L);
+    WarpScriptStack.Macro chunkMacro = (WarpScriptStack.Macro) params.get(CHUNK_MACRO);
+
+    //
+    // Check URL
+    //
+
+    Object urlParam = params.get(URL);
+    if (null == urlParam) {
+      throw new WarpScriptException(getName() + " expects an url.");
```
hbs, line 208: a URL

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+      downloadSize = new AtomicLong();
+      stack.setAttribute(HttpWarpScriptExtension.ATTRIBUTE_HTTP_REQUESTS, urlCount);
+      stack.setAttribute(HttpWarpScriptExtension.ATTRIBUTE_HTTP_SIZE, downloadSize);
+    } else {
+      urlCount = (AtomicLong) ufCount;
+      downloadSize = (AtomicLong) ufSize;
+    }
+
+    if (urlCount.addAndGet(1) > maxrequests) {
+      throw new WarpScriptException(getName() + " is limited to " + maxrequests + " calls.");
```
hbs, line 248: per script execution

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+      conn.setDoInput(true);
+      conn.setRequestMethod(method.toUpperCase());
+
+      //
+      // Set body
+      //
+
+      if (body instanceof String) {
+        String bodyS = (String) body;
+        conn.setDoOutput(bodyS.length() > 0);
```
hbs, line 305: This will probably produce a weird behavior if method is not POST as setting doOutput to true normally forces the method to be POST

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+        if (bodyS.length() > 0) {
+          try (OutputStream os = conn.getOutputStream()) {
+            os.write(bodyS.getBytes(StandardCharsets.UTF_8));
+          }
+        }
+
+      } else if (body instanceof byte[]) {
+        byte[] bodyB = (byte[]) body;
+        conn.setDoOutput(bodyB.length > 0);
```
hbs, line 314: ditto, body should be ignored if method is GET or another method which does not allow a body

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+        }
+      } else {
+        res.put(STATUS_MESSAGE, "");
+      }
+
+      //
+      // Make the headers map modifiable
+      //
+
+      hdrs = new LinkedHashMap<String, List<String>>(hdrs);
```
hbs, line 347: are the lists inside of the original hdrs map modifiable?

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+          ByteArrayOutputStream baos = new ByteArrayOutputStream();
+
+          while (true) {
+            int len = in.read(buf);
+            if (len < 0) {
+              break;
+            }
+
+            if (downloadSize.get() + baos.size() + len > maxsize) {
+              throw new WarpScriptException(getName() + " would exceed maximum size of content which can be retrieved via this function (" + maxsize + " bytes)");
```
hbs, line 378: per script execution

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+      InputStream in = null;
+
+      // When there is an error (response code is 404 for instance), body is in the error stream.
+      try {
+        in = conn.getInputStream();
+      } catch (IOException ioe) {
+        in = conn.getErrorStream();
+      }
+
+      if (chunkSize <= 0) {
```
hbs, line 365: what semantics does a negatice chunkSize have?

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,449 @@
...
+        if (null != in) {
+
+          int chunkNumber = 0;
+          while (true) {
+            chunkNumber++;
+
+            Map<String, Object> chunkRes = new LinkedHashMap<>(res);
+            byte[] buf = new byte[chunkSize.intValue()];
+            int len = in.read(buf);
+            if (len < 0) {
```
hbs, line 402: If EOF is reached while attempting to read buf.length bytes you will exit the loop and miss the last part of the data

hbs (changes requested):

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
randomboolean: none other than not using chunk option

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,533 @@
...
+          }
+
+        } else if (body instanceof byte[]) {
+          byte[] bodyB = (byte[]) body;
+          conn.setDoOutput(bodyB.length > 0);
+          if (bodyB.length > 0) {
+            try (OutputStream os = conn.getOutputStream()) {
+              os.write(bodyB);
+            }
+          }
```
ftence, line 381: Don't duplicate the code, simply convert the String to byte[], with `bodyS.getBytes(StandardCharsets.UTF_8)`, then do the logic on the body as byte[].

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,533 @@
...
+      }
+
+      conn.setDoInput(true);
+      conn.setRequestMethod(method.toUpperCase());
+
+      //
+      // Set body
+      //
+
+      if ("GET" != method && "DELETE" != method && "TRACE" != method && "OPTIONS" != method && "HEAD" != method) {
```
ftence, line 363: In fact, only TRACE cannot have a body, see the source code of [HttpURLConnection](https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#L1319). So either check for TRACE (using `equals`, not `==`) or handle the IOException it may throw.

The method is changed from GET to POST if there is a body by the getOutputStream. My opinion is to let that happen. This is coherent with defaulting to "POST" is no method is defined and there is a body.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,533 @@
...
+
+  //
+  // Arguments
+  //
+
+  public static final String METHOD = "method";
+  public static final String URL = "url";
+  public static final String HEADERS = "headers";
+  public static final String BODY = "body";
+  public static final String HEADERS_MACRO = "headers.macro";
```
ftence, line 77: I'm, again, not convinced by the usefulness of this parameter. This does not offer more feature or readability than coding that in WarpScript:
```
{
  /* All your HTTP parameters */
}
@your_macro_updating_headers
HTTP
```

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,533 @@
...
+    HttpURLConnection conn = null;
+
+    try {
+      conn = (HttpURLConnection) url.openConnection();
+
+      //
+      // Set headers
+      //
+
+      if (null == headersMacro) {
```
ftence, line 315: I don't see why basic auth and header modification should be exclusive.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,533 @@
...
+
+      } else {
+
+        //
+        // Chunk processing
+        //
+
+        if (null != in) {
+
+          int chunkNumber = 0;
```
ftence, line 460: `chunk.number` is 1-indexed. I'm wondering if that's coherent with other indexes in WarpScript.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -0,0 +1,533 @@
...
+        stack.push(chunkRes);
+        if (null != chunkMacro) {
+          stack.exec(chunkMacro);
+        }
+
+        //
+        // End chunk processing
+        //
+
+        res.remove(CHUNK_NUMBER);
```
ftence, line 518: You can also remove `CONTENT`.

ftence (changes requested):

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
randomboolean: TIMESPLIT for example is 1-indexed, but CHUNK is indexed with the first tick of the partial GTS.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
randomboolean: I think it can be confusing if a method is explicit in the arguments and it gets changed to POST.

ftence: You're right. It that case, just check for `TRACE` and `GET`.

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
ftence: OK, that's fine as long as it will be documented ;).

ftence (changes requested):

ftence (approved):

PR #892, opened 2021-01-04
Disable WarpInit with in memory mode to avoid IO error with LevelDB.

Giwi:
Do not call WarpInit when starting Warp 10 in memory.

### DIFF
--- a/warp10/src/main/sh/warp10-standalone.sh
+++ b/warp10/src/main/sh/warp10-standalone.sh
@@ -416,10 +416,10 @@ start() {
   # Extract configuration keys
   # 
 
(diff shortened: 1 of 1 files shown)


warp10/src/main/sh/warp10-standalone.sh
```diff
@@ -432,11 +432,13 @@ start() {
     fi
   fi
 
-  if [ "$(find -L ${LEVELDB_HOME} -maxdepth 1 -type f | wc -l)" -eq 0 ]; then
-    echo "Init leveldb"
-    # Create leveldb database
-    echo \"Init leveldb database...\" >> ${WARP10_HOME}/logs/warp10.log
-    ${JAVACMD} ${JAVA_OPTS} -cp ${WARP10_CP} ${WARP10_INIT} ${LEVELDB_HOME} >> ${WARP10_HOME}/logs/warp10.log 2>&1
+  if [ "${IN_MEMORY:-}" != "true" ]; then
```
ftence, line 435: You can also include the creation of the directory above in the if block.

ftence (changes requested):
2021 copyright notice, happy new year! ;)

warp10/src/main/sh/warp10-standalone.sh
Giwi: done

ftence (approved):

PR #893, opened 2021-01-11
Fix DEDUP signature so that it reduces list nesting

ftence:
When `DEDUP` is used with an aggregator or a macro (new in 2.7.0), it always generates a list of GTSs for each GTS. This is because this function is very similar to `MAP` so aggregators/macros can generate several GTSs. However, in the case the aggregator/macro returns a single GTS, `MAP` does not nest lists for simplicity.

This PR aims at replicating this behavior in `DEDUP`, avoiding list nesting when possible.

This does change the signature of `DEDUP`, however as this functionality is rather new and is already [documented as avoiding nesting](http://warp10.io/doc/DEDUP), this change may be considered as acceptable.

Signatures without this PR:
`GTS AGGREGATOR/MACRO` **`DEDUP`** `LIST<GTS>`
`LIST<GTS> AGGREGATOR/MACRO` **`DEDUP`** `LIST<LIST<GTS>>`

Signatures with this PR:
For an aggregator/macro generating a single GTS (most common case):
`GTS AGGREGATOR/MACRO` **`DEDUP`** `GTS`
`LIST<GTS> AGGREGATOR/MACRO` **`DEDUP`** `LIST<GTS>`
For an aggregator/macro generating multiple GTSs:
`GTS AGGREGATOR/MACRO` **`DEDUP`** `LIST<GTS>`
`LIST<GTS> AGGREGATOR/MACRO` **`DEDUP`** `LIST<LIST<GTS>>`

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/DEDUP.java
+++ b/warp10/src/main/java/io/warp10/script/functions/DEDUP.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2020  SenX S.A.S.
+//   Copyright 2018-2021  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -22,6 +22,8 @@ import io.warp10.script.ListRecursiveStackFunction;
 import io.warp10.script.WarpScriptException;
 import io.warp10.script.WarpScriptMapperFunction;
 import io.warp10.script.WarpScriptStack;
+
+import java.util.List;
 
 /**
  * Remove duplicates from GTS instances.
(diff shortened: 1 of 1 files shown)


hbs:
The documentation for DEDUP could go more into the details of what the aggregator/mapper can output and what effect it would have on the final result, especially on the number of GTS.

ftence:
I agree, but I wouldn't change the doc without knowing first if the change introduced by this PR is deemed acceptable or not.

hbs:
While removing the nesting when all inner lists have a single value is indeed a good idea, having DEDUP output a GTS instead of a LIST<GTS> does not seem wise since it will force the calling code to perform a test to check if the result is a list or not prior to extracting the resulting GTS. So I would rather see the following new signatures

ftence:
I understand you want to use the same rule as `MAP` but I think it could confuse users because `GTS DEDUP` will leave a single `GTS` but `GTS AGGREGATOR/MACRO DEDUP` will leave a `LIST`.

If your code has to test whether the result is a list or not would mean you don't know if the parameters are `LIST<GTS>` or simply `GTS` or if your mapper/macro generates a single `GTS` of several `LIST<GTS>`. I'm not sure the former often happens, correct me if I'm wrong, and for the latter you will also have to check if it's a `LIST<GTS>` or `LIST<LIST<GTS>>` or do a potentially useless `FLATTEN`.

hbs:
If you are the one providing the aggregator/macro then it is reasonable to assume you know what you are doing, but if you are writing a macro which accepts as a parameter an aggregator then you might not know.

hbs:
A useless FLATTEN on a LIST with no nesting works, but a FLATTEN on a GTS fails, hence my suggestion to always return a LIST

ftence:
I feel like we complicating things for most of the cases for a rare case which can be solved by `[ SWAP ] FLATTEN`.

Also you may want a different logic if your mapper return several GTSs, which is a case you may not be able to detect with your suggestion.

PR #894, opened 2021-01-13
Initial commit of MacroResolver

hbs:
The MacroResolver class allows to register custom macro resolution mechanism. This will allow for example the definition of macros directly in FLoWS.


pi-r-p (approved):
Great !

ftence (approved):

PR #900, opened 2021-01-14
Add DELETE dryrun option returning a list of would-be deleted GTSs

ftence:


warp10/src/main/java/io/warp10/script/functions/DELETE.java
```diff
@@ -52,19 +57,22 @@ public DELETE(String name) {
...
+    boolean onlyDryRun = false;
+    long expected = Long.MAX_VALUE;
+
     Object o = stack.pop();
-    
-    if (!(o instanceof Long)) {
-      throw new WarpScriptException(getName() + " expects a count on top of the stack.");
+
+    if (o instanceof Long) {
+      expected = (long) o;
```
hbs, line 69: This will fail, use ((Long) o).longValue();

hbs, line 69: Use ((Long) o).longValue(), autoboxing makes things harder to read and could lead to failing code.

PR #901, opened 2021-01-19
Processing.pencode

hbs:
Add support for calling Pencode on an IMAGE instance.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/processing/Pencode.java
+++ b/warp10/src/main/java/io/warp10/script/processing/Pencode.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2020  SenX S.A.S.
+//   Copyright 2018-2021  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -15,11 +15,6 @@
 //
 
 package io.warp10.script.processing;
-
-import io.warp10.script.NamedWarpScriptFunction;
-import io.warp10.script.WarpScriptStackFunction;
-import io.warp10.script.WarpScriptException;
-import io.warp10.script.WarpScriptStack;
 
 import java.awt.image.BufferedImage;
 import java.io.BufferedOutputStream;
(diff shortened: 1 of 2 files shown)


warp10/src/main/java/io/warp10/script/processing/image/Presize.java
```diff
@@ -77,7 +79,13 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
     if (top instanceof PImage) { // PGraphics extends PImage
       PImage img = (PImage) top;
+      if (top instanceof PGraphics) {
```
ftence, line 82: You indeed found an issue with Presize on a Pgraphics but it seems this fix it not enough. 
```
1000 1000 '2D8' PGraphics
200 200 Presize
```
Produces
`sun.java2d.SunGraphics2D cannot be cast to java.awt.image.BufferedImage`

warp10/src/main/java/io/warp10/script/processing/Pencode.java
```diff
@@ -39,36 +34,47 @@
...
+
     if (top instanceof Map) {
       chunks = (Map<Object,Object>) top;
       top = stack.pop();
     }
-    
-    if (!(top instanceof processing.core.PGraphics)) {
-      throw new WarpScriptException(getName() + " operates on a PGraphics instance.");
+
+    if (!(top instanceof processing.core.PImage)) {
```
ftence, line 64: You could use `PImage` instead of `processing.core.PImage`.
Same with `PGraphics` and `processing.core.PGraphics` x3.

ftence (changes requested):

ftence (approved):
Tested using
```
1000 1000 '2D8' PGraphics

500 PstrokeWeight
500 500 Ppoint
'pg' STORE

$pg Pencode 'original' STORE

$pg
200 200 Presize

20 PstrokeWeight
10 10 Ppoint

PtoImage
Pencode

$original
```

Which produces
Warp 10-Processing Image-2021-01-20T08_49_46 172Z
Warp 10-Processing Image-2021-01-20T08_49_48 492Z

PR #904, opened 2021-01-20
Add LevelDB createIfMissing Warp 10 configuration

ftence:
This can simplify tests, for instance, when tests have to be done on an empty LevelDB.

I'm not sure it should replace WarpInit however, because WarpInit makes additional checks (verifyChecksums and paranoidChecks).


hbs:
Can you validate it works also with the pure java implementation of LevelDB.

ftence:
It works. 
The only difference is that the pure java implementation creates the directory and LOCK file (but not the LOG file) when `leveldb.create.if.missing = false`. I updated the comment accordingly.

PR #909, opened 2021-01-22
Fixed NPE when deleting with start == 0 in the presence of uninitialized chunks

hbs:


warp10/src/main/java/io/warp10/standalone/InMemoryChunkSet.java
```diff
@@ -807,80 +807,80 @@ public long clean(long now) {
...
-            continue;
-          }
-          try {
+        try {
+          while (decoder.next()) {
+            if (decoder.getTimestamp() >= start && decoder.getTimestamp() <= end) {
+              deleted = true;
+              count++;
+              continue;
+            }
```
ftence, line 879: Only `encoder.addValue` can throw an `IOException`, why did you enclose the whole loop?

ftence (approved):

warp10/src/main/java/io/warp10/standalone/InMemoryChunkSet.java
hbs: Ridiculous change on my side :-), rolling back

PR #914, opened 2021-02-04
Make ADD handle secure macros

ftence:


ftence:
This last version makes it possible the use of MACROCONFIG in both macros because they keep their names.
They also keep their secure flags independently.

ftence:
After giving it more thoughts, I reverted back to the original solution so that the the idea of an addition is kept. The fix to the secureness it still added by this PR though.

If anyone wants to combine two macros keeping the names, it can be done easily with `<% @op1 @op2 %>`.

PR #917, opened 2021-02-10
Added support for FETCH to return GTSEncoders

hbs:


warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -566,11 +582,77 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+                while(copy > 0 && decoder.next()) {
+                  encoder.addValue(decoder.getTimestamp(), decoder.getLocation(), decoder.getElevation(), decoder.getBinaryValue());
+                  copy--;
+                }
+              }
+
+              lastCount += encoder.getCount();
+
+              if (merge) {
+                if (null == lastEncoder || lastEncoder.getName() != encoder.getName() || !lastEncoder.getLabels().equals(encoder.getLabels())) {
```
ftence, line 636: Should be `lastEncoder.getName().equals(encoder.getName())`

ftence (changes requested):

ftence (approved):

PR #921, opened 2021-02-22
Uniformize and improve error messages for ATC Exceptions

ftence:
For a script containing only `CONTINUE` the error message is currently `CONTINUE`.
This PR changes it to `Exception at CONTINUE: must be used in a one of FOR, FOREACH, FORSTEP, WHILE, UNTIL loop.`

This is more informative than what we have currently but less than what we get for other WarpScriptException. Having the same level of information will probably have a negative impact on the performance on the use of WarpScriptATCException so this PR might be the right balance.

`CONTINUE` and `BREAK` now use a single Exception instance which is better in terms of performance but makes the display of the stack trace impossible. I'm not sure if it is problematic or not.

`RETURN` and `NRETURN` has also been merged and the `NRETURN` class is marked as deprecated, but not the function.

Some useless imports have been removed.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/NRETURN.java
+++ b/warp10/src/main/java/io/warp10/script/functions/NRETURN.java
@@ -35,7 +35,7 @@ public class NRETURN extends NamedWarpScriptFunction implements WarpScriptStackF
 
   public NRETURN(String name) {
     super(name);
-    ex = new WarpScriptReturnException("name");
+    ex = new WarpScriptReturnException(name);
   }
 
   @Override
@@ -43,7 +43,7 @@ public class NRETURN extends NamedWarpScriptFunction implements WarpScriptStackF
     Object top = stack.pop();
 
     if (!(top instanceof Long)) {
-      throw new WarpScriptException(getName() + " expects a number of levels on top of the stack.");
+      throw new WarpScriptException(getName() + " expects a number of levels.");
     }
 
     stack.getCounter(WarpScriptStack.COUNTER_RETURN_DEPTH).set(((Number) top).longValue());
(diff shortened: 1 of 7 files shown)


warp10/src/main/java/io/warp10/script/functions/NRETURN.java
```diff
@@ -17,30 +17,35 @@
...
+@Deprecated
 public class NRETURN extends NamedWarpScriptFunction implements WarpScriptStackFunction {
-  
+
+  // The stack trace is not used so it can be instantiated once.
+  private final WarpScriptReturnException ex;
+
   public NRETURN(String name) {
     super(name);
+    ex = new WarpScriptReturnException("name");
```
hbs, line 38: I guess you meant (name)

warp10/src/main/java/io/warp10/script/functions/NRETURN.java
```diff
@@ -17,30 +17,35 @@
...
-  private final WarpScriptReturnException ex = new WarpScriptReturnException("NRETURN");
-  
+
   @Override
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
     Object top = stack.pop();
-    
+
     if (!(top instanceof Long)) {
       throw new WarpScriptException(getName() + " expects a number of levels on top of the stack.");
```
hbs, line 46: Remove  mention of 'stack'

warp10/src/main/java/io/warp10/script/functions/TIMESPLIT.java
```diff
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018  SenX S.A.S.
+//   Copyright 2018-2021  SenX S.A.S.
```
hbs, line 2: This is unrelated, but no big deal.

hbs (changes requested):

PR #924, opened 2021-02-23
Fix javadoc warnings

ftence:


hbs (approved):
No copyright notice updates but it would not really make sense since the code did not change, only the javadoc.

PR #925, opened 2021-02-26
Change publishing pipeline so that it does not rely on Bintray anymore

ftence:


hbs:
Could you have a look at the current conflicts so we can merge this PR.

PR #927, opened 2021-03-02
Shamir Secret Sharing Scheme functions

hbs:


trepca:
oh, wow, how could this be applied? I mean, I assume multiple clients would have to simultaneously come together to execute a secure script?

hbs:
The use cases are plentiful, the two added functions simply enable the splitting and recovery of data with a K out of N scheme, your creativity will do the rest ;-)

ftence (changes requested):

hbs:
PTAL

ftence (approved):

PR #935, opened 2021-03-19
PcreateFont: Fix parameter handling to allow for default 12 size

ftence:
The documentation specifies that the size is optional and the online font case (line 92) does handle that case with a default of 12 size font.


warp10/src/main/java/io/warp10/script/processing/typography/PcreateFont.java
```diff
@@ -134,7 +134,9 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
         if (null != in) { try { in.close(); } catch (Exception e) {} }
       }      
     } else {
-      if (3 == params.size()) {      
+      if (2 == params.size()) {
+        font = pg.parent.createFont(params.get(1).toString(), 12.0F);
```
hbs, line 138: Maybe this should be explicitely set in a constant in the class.

PR #937, opened 2021-03-22
Make more types snapshotable

ftence:
This PR adds:
- PGraphics
- PImage
- RealVector
- RealMatrix
- Matcher
- PShape (using reflection)

Types that can be created with core WarpScript that are still unhanded:
- PFont because it does not contain enough information to recreate the font with `PcreateFont` like the URI.
- StackContext because it could expose defined functions.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/SNAPSHOT.java
+++ b/warp10/src/main/java/io/warp10/script/functions/SNAPSHOT.java
@@ -35,9 +35,11 @@ import org.apache.commons.math3.linear.RealVector;
 import processing.awt.PGraphicsJava2D;
 import processing.core.PGraphics;
 import processing.core.PImage;
+import processing.core.PShapeSVG;
 import processing.opengl.PGraphics3D;
 
 import java.io.UnsupportedEncodingException;
+import java.lang.reflect.Field;
 import java.nio.charset.StandardCharsets;
 import java.security.interfaces.RSAPrivateKey;
 import java.security.interfaces.RSAPublicKey;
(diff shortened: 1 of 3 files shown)


warp10/src/main/java/io/warp10/script/functions/SNAPSHOT.java
```diff
@@ -407,6 +416,90 @@ public static void addElement(SNAPSHOT snapshot, StringBuilder sb, Object o, boo
...
+          for (int j = 0; j < matrix.getRowDimension(); j++) {
+            sb.append(matrix.getEntry(i, j));
+            sb.append(" +! ");
+          }
+          sb.append(" +! ");
+        }
+        sb.append(WarpScriptLib.TOMAT);
+        sb.append(" ");
+      } else if (o instanceof Matcher) {
+        sb.append(((Matcher) o).pattern());
```
hbs, line 499: Probably missing quotes around the actual pattern.

warp10/src/main/java/io/warp10/script/functions/SNAPSHOT.java
```diff
@@ -407,6 +416,90 @@ public static void addElement(SNAPSHOT snapshot, StringBuilder sb, Object o, boo
       } else if (o instanceof NamedWarpScriptFunction) {
         sb.append(o.toString());
         sb.append(" ");
+      } else if (o instanceof PGraphics) {
```
hbs, line 419: This is subject to caution as this snapshot does not preserve the state of the PGraphics (selected font, current path, current attributes such as widtht, color, etc.)

hbs (changes requested):

ftence:
I removed PGraphics shapshot altogether. I will probably do another PR to introduce PFont and PGraphics snapshots be it will require quite a bit of discussion and new code so I'm not putting that in this PR.

warp10/src/main/java/io/warp10/script/functions/SNAPSHOT.java
```diff
@@ -407,6 +416,56 @@ public static void addElement(SNAPSHOT snapshot, StringBuilder sb, Object o, boo
...
+          sb.append(" ");
+        } catch (NoSuchFieldException | IllegalAccessException e) {
+          sb.append("'UNSUPPORTED:" + WarpURLEncoder.encode(o.getClass().toString(), StandardCharsets.UTF_8) + "' ");
+        }
+      } else if (o instanceof RealVector) {
+        RealVector vector = (RealVector) o;
+        sb.append("[] ");
+        for (int i = 0; i < vector.getDimension(); i++) {
+          sb.append(vector.getEntry(i));
+          sb.append(" +! ");
```
hbs, line 446: Use WarpScriptLib.INPLACEADD

warp10/src/main/java/io/warp10/script/functions/SNAPSHOT.java
```diff
@@ -407,6 +416,56 @@ public static void addElement(SNAPSHOT snapshot, StringBuilder sb, Object o, boo
...
+          sb.append(element.toString());
+          sb.append("' ");
+          sb.append(WarpScriptLib.PLOADSHAPE);
+          sb.append(" ");
+        } catch (NoSuchFieldException | IllegalAccessException e) {
+          sb.append("'UNSUPPORTED:" + WarpURLEncoder.encode(o.getClass().toString(), StandardCharsets.UTF_8) + "' ");
+        }
+      } else if (o instanceof RealVector) {
+        RealVector vector = (RealVector) o;
+        sb.append("[] ");
```
hbs, line 443: Make a constants EMPTYLIST in WarpScriptLib and use it here and for RealMatrix.

PR #940, opened 2021-03-26
Added white space checks in snapshot to avoid missing/repeated spaces

hbs:


warp10/src/main/java/io/warp10/script/WarpScriptStack.java
```diff
@@ -431,6 +431,18 @@ public String snapshot(boolean hideSecure) {
             } else {
               SNAPSHOT.addElement(sb, o);
             }
+            int idx = sb.length() - 1;
+            if (' ' != sb.charAt(idx)) {
```
ftence, line 435: This will fail if the snapshot is empty.

warp10/src/main/java/io/warp10/script/WarpScriptStack.java
```diff
@@ -431,6 +431,18 @@ public String snapshot(boolean hideSecure) {
...
+              // The last element does not end with a space, add one
+              sb.append(" ");
+            } else {
+              // Check if we only have a single space
+              idx--;
+              while (idx >= 0 && ' ' == sb.charAt(idx)) {
+                idx--;
+              }
```
ftence, line 443: Use a do while instead?

ftence (changes requested):

hbs:
PTAL

ftence (approved):

PR #941, opened 2021-03-30
Small optimization avoiding asList

ftence:


warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -961,7 +960,7 @@ public void exec(Macro macro) throws WarpScriptException {
       for (i = 0; i < n; i++) {
         handleSignal();
 
-        Object stmt = stmts.get(i);
```
hbs, line 964: Need to change copyright notice

hbs: My bad, was looking at wrong file!

PR #942, opened 2021-03-31
Added Keccak/SHA-3 digest functions

hbs:


hbs:
Note that the change of BouncyCastle dependency may break third party extensions/plugins which expect Warp 10 to provide the dependency.

hbs:
Those exposed have foreseeable use cases, the others not yet, but they could be added later if needed now that BC has been bumped.

PR #946, opened 2021-04-06
ECRECOVER - Recovering ECC public key from curve, hash and signature

hbs:


warp10/src/main/java/io/warp10/script/functions/ECRECOVER.java
```diff
@@ -0,0 +1,234 @@
...
+
+    z = new BigInteger(1, hash);
+
+    if (nbits < hash.length * 8) {
+      z = z.shiftRight(hash.length * 8 - nbits);
+    }
+
+    List<Object> candidates = new ArrayList<Object>();
+
+    for (int j = 0; j < H.intValue(); j++) {
```
ftence, line 164: It can take more than 3 minutes to execute this loop (tested with c2pnb368w1 and SHA1). We should protect this function with either capabilities or authentication to avoid abuse.

ftence (changes requested):

warp10/src/main/java/io/warp10/script/functions/ECRECOVER.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    Map<Object,Object> params = (Map<Object,Object>) top;
+
+    if (!(params.get(KEY_CURVE) instanceof String)) {
+      throw new WarpScriptException(getName() + " missing ECC curve name under '" + KEY_CURVE + "'.");
+    }
+    final ECNamedCurveParameterSpec spec = ECNamedCurveTable.getParameterSpec((String) params.get(KEY_CURVE));
+
+    ECFieldElement N = spec.getCurve().fromBigInteger(spec.getN().mod(spec.getCurve().getField().getCharacteristic()));
```
ftence, line 72: Unused

warp10/src/main/java/io/warp10/script/functions/ECRECOVER.java
```diff
@@ -0,0 +1,238 @@
...
+    BigInteger H = spec.getH();
+    ECFieldElement A = spec.getCurve().getA();
+    ECFieldElement B = spec.getCurve().getB();
```
ftence, line 75: Unused

ftence (changes requested):

warp10/src/main/java/io/warp10/script/functions/ECRECOVER.java
```diff
@@ -194,7 +212,6 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
           //𝑟−1(𝑠𝑅−𝑧𝐺)  and 𝑟−1(𝑠𝑅′−𝑧𝐺)
```
ftence, line 213: Is that comment still valid? Should it be moved with the `BigInteger rinv` line?

warp10/src/main/java/io/warp10/script/functions/ECRECOVER.java
```diff
@@ -163,9 +165,25 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      Map<String,String> capabilities = Capabilities.get(stack, (List) null);
+      if (null != capabilities.get(CAP_COFACTOR)) {
+        int hmax = Integer.valueOf(capabilities.get(CAP_COFACTOR));
+        if (maxH > hmax) {
+          throw new WarpScriptException(getName() + " cofactor " + maxH + " is above the maximum " + hmax + " allowed by the '" + CAP_COFACTOR + "' capability.");
+        }
+      } else {
+        throw new WarpScriptException(getName() + " cofactor " + maxH + " is above allowed maximum " + MAX_COFACTOR + ", increase this limit using a token with the '" + CAP_COFACTOR + "' capability.");
+      }
+    }
```
ftence, line 184: Why don't you use `Capabilities.get(stack, CAP_COFACTOR);`?
Checking for a valid capability value is also nice and it can be merged with the null check like this:
```
    if (maxH > MAX_COFACTOR) {
      String hmaxCap = Capabilities.get(stack, CAP_COFACTOR);
      try {
        int hmax = Integer.valueOf(hmaxCap);
        if (maxH > hmax) {
          throw new WarpScriptException(getName() + " cofactor " + maxH + " is above the maximum " + hmax + " allowed by the '" + CAP_COFACTOR + "' capability.");
        }
      } catch (NumberFormatException nfe) {
        throw new WarpScriptException(getName() + " cofactor " + maxH + " is above allowed maximum " + MAX_COFACTOR + ", increase this limit using a token with the '" + CAP_COFACTOR + "' capability.");
      }
    }
```

ftence (changes requested):

ftence (approved):

PR #951, opened 2021-04-06
Add FUNCREF and improve snapshots of WarpScriptStackFunctions

ftence:
This is an attempt to improve the snapshot of `WarpScriptStackFunctions` and `NamedWarpScriptFunctions` for https://github.com/senx/warp10-platform/pull/945.

This also solve a bug with the snapshots of `WarpScriptStackFunctions` whose reference is on the stack, using `NPDF` for instance. `42 mapper.gt` does not trigger the bug because it is only a `NamedWarpScriptFunction`.

Tested using:
```
42 mapper.gt 'M' STORE
0 1 NPDF 'N' STORE
<%
  'Hello' 42 
  53 SQRT
  42 mapper.gt
  !$M
  0 0 1 NPDF EVAL
  0 !$N
%>
'macro' STORE

// Macro snapshot and EVAL
$macro
SNAPSHOT
EVAL
EVAL

// Snapshot of MACRO-> and EVAL
$macro
MACRO->
SNAPSHOT
EVAL
->MACRO
EVAL

// EVAL of the macro to check the results
@macro
```

### DIFF
--- a/warp10/src/main/java/io/warp10/script/NamedWarpScriptFunction.java
+++ b/warp10/src/main/java/io/warp10/script/NamedWarpScriptFunction.java
@@ -16,6 +16,11 @@
 
 package io.warp10.script;
 
+import io.warp10.WarpURLEncoder;
+
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
+
 public class NamedWarpScriptFunction {
   private String name = "##UNDEF##";
   
@@ -35,7 +40,17 @@ public class NamedWarpScriptFunction {
     return this.name;
   }
 
+  /**
+   * Gives the snapshot to get the reference to this function.
+   */
   public final String refSnapshot() {
-    return "'" + this.name + "' " + WarpScriptLib.FUNCREF;
+    if (null == name) {
+      return null;
+    }
+    try {
+      return "'" + WarpURLEncoder.encode(name, StandardCharsets.UTF_8) + "' " + WarpScriptLib.FUNCREF;
+    } catch (UnsupportedEncodingException e) {
+      return null;
+    }
   }
 }
--- a/warp10/src/main/java/io/warp10/script/NamedWarpScriptFunction.java
+++ b/warp10/src/main/java/io/warp10/script/NamedWarpScriptFunction.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018  SenX S.A.S.
+//   Copyright 2018-2021  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -22,9 +22,9 @@ public class NamedWarpScriptFunction {
   public NamedWarpScriptFunction(String name) {
     this.name = name;
   }
-  
+
   public String toString() {
-    return this.name;
+    return "'" + this.name + "' " + WarpScriptLib.FUNCREF;
   }
   
   public void setName(String name) {
--- a/warp10/src/main/java/io/warp10/script/NamedWarpScriptFunction.java
+++ b/warp10/src/main/java/io/warp10/script/NamedWarpScriptFunction.java
@@ -24,7 +24,7 @@ public class NamedWarpScriptFunction {
   }
 
   public String toString() {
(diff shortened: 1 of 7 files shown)


warp10/src/main/java/io/warp10/script/WarpScriptStack.java
```diff
@@ -428,6 +428,20 @@ public String snapshot(boolean hideSecure) {
             if (o instanceof Macro) {
               sb.append(((Macro) o).snapshot(hideSecure));
               sb.append(" ");
+            } else if (o instanceof WarpScriptStackFunction) {
+              String funcSnapshot = o.toString();
+              // In the case the snapshot of the function is 'MYFUNC' FUNCREF, instead of adding
+              // 'MYFUNC' FUNCREF EVAL to the snapshot, MYFUNC can simply be added.
+              if (o instanceof NamedWarpScriptFunction
+                  && null != funcSnapshot
+                  && funcSnapshot.equals("'" + ((NamedWarpScriptFunction) o).getName() + "' " + WarpScriptLib.FUNCREF)) {
```
hbs, line 437: This construct should be delegated to a static function of NamedWarpScriptFunction to avoid having to track where it is used throughout the code base should it change.

warp10/src/main/java/io/warp10/script/functions/FUNCREF.java
```diff
@@ -0,0 +1,47 @@
...
+package io.warp10.script.functions;
+
+import io.warp10.script.NamedWarpScriptFunction;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptLib;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.script.WarpScriptStackFunction;
+
+/**
+ * Push the reference of the giving function named based on functions known to WarpScriptLib.
```
hbs, line 26: given

hbs (commented):
This PR raises an issue with MACRO-> which is the case of the redefined functions (via DEF). How do you think we should handle that? Maybe the macro should be wrapped into a NamedWarpScripFunction instead of an anonymous WarpScriptStackFunction. And FUNCREF should probably apply the same logic as MemoryWarpScriptStack for determining the actual function to emit.

ftence:
Tested a wide range of cases with the following script but I may have missed some edge cases.

```
<%
  SAVE '.context' STORE

  'macro' STORE

  {
    'snapshot_macro'
    [
      $macro
      1 SNAPSHOTN
      EVAL
      <% EVAL %>
      <% ERROR -1 GET 'message' GET %>
      <% %>
      TRY
    ]

    'snapshot_macroto'
    [
      $macro
      MACRO2->
      1 SNAPSHOTN
      EVAL
      ->MACRO2
      <% EVAL %>
      <% ERROR -1 GET 'message' GET %>
      <% %>
      TRY
    ]

    'no_snapshot'
    [ 
      $macro 
      <% EVAL %>
      <% ERROR -1 GET 'message' GET %>
      <% %>
      TRY
    ]
  }

  $.context RESTORE
%>
'test_macro' STORE



<% PUSHR0 %>
{ 0 <% <% 'foobar' %> %> }
DEREF
@test_macro

<% NOW %> 'ALSONOW' DEF
<% ALSONOW %>
@test_macro

42 mapper.gt 'M' STORE
0 1 NPDF 'N' STORE
<%
  'Hello' 42 
  53 SQRT
  42 mapper.gt
  !$M
  0 0 1 NPDF EVAL
  0 !$N
%>
@test_macro

NULL 'NOW' DEF
<% NOW %>
@test_macro

0 1 NPDF 'N' STORE
<%
  'a' !$N
%>
@test_macro
```

warp10/src/main/java/io/warp10/script/NamedWarpScriptFunction.java
```diff
@@ -34,4 +34,8 @@ public void setName(String name) {
   public String getName() {
     return this.name;
   }
+
+  public final String refSnapshot() {
+    return "'" + this.name + "' " + WarpScriptLib.FUNCREF;
```
hbs, line 39: The name needs to be encoded as SNAPSHOT does for STRINGs as it may contain any characters.

hbs (changes requested):

PR #952, opened 2021-04-12
Merge TOBIN and ->BIN as well as TOHEX and ->HEX.

ftence:
Also make ->BIN/TOBIN work on BitSets.


warp10/src/main/java/io/warp10/script/functions/TOBIN.java
```diff
@@ -17,57 +17,69 @@
...
+      stack.push(sb.toString());
+    } else if (o instanceof Long) {
+      sb.append("0000000000000000000000000000000000000000000000000000000000000000");
+
+      sb.append(Long.toBinaryString(((Number) o).longValue()));
+      stack.push(sb.substring(sb.length() - 64));
+    } else if (o instanceof BitSet) {
+      BitSet bitset = (BitSet) o;
+
+      for (int i = 0; i < bitset.size(); i++) {
```
hbs, line 74: Shouldn't we use bitset.length() instead of .size()?

warp10/src/main/java/io/warp10/script/functions/TOBIN.java
```diff
@@ -17,57 +17,69 @@
...
+        sb.append(bitset.get(i) ? '1' : '0');
+      }
+
+      stack.push(sb);
```
hbs, line 78: sb.toString()

hbs (changes requested):

warp10/src/main/java/io/warp10/script/functions/TOBIN.java
ftence: You're right, `length()` is better than `size()` in our case.

PR #955, opened 2021-04-21
Add .noauth, .nometa and .nofetch attributes for read tokens

ftence:
This will allow read tokens that, among other things:
- can fetch data but cannot be used for authentication
- can be used for authentication but not to fetch data
- can used to find metadata but not actual data

This also offers an alternative solution for pure capabilities tokens as you can either:
- use a write token with .noauth, .nometa, .noupdate and .nodelete
- use a read token with .noauth, .nometa and .nofetch (instead of using a dummy app/owner/producer)


hbs:
I am not yet done with my review but I think you should introduce a .nofind attribute that you will use instead of .nometa to limit the read access to metadata. It is awkward to have .nometa restrict the find operation when those attributes are named .noOP (OP being update/delete/meta) for the other 3 restrictions.

hbs (changes requested):

PR #958, opened 2021-04-22
Added parameters i and even to limit the number of candidate keys of ECRECOVER

hbs:


warp10/src/main/java/io/warp10/script/functions/ECRECOVER.java
```diff
@@ -46,6 +46,8 @@
 
   private static final String KEY_R = "r";
   private static final String KEY_S = "s";
+  private static final String KEY_I = "i";
```
ftence, line 49: Why don't you call that parameter `h`?

warp10/src/main/java/io/warp10/script/functions/ECRECOVER.java
```diff
@@ -169,9 +171,15 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
     BigInteger rinv = spec.getCurve().fromBigInteger(r.modInverse(spec.getN())).toBigInteger();
 
+    int minH = 0;
     int maxH = H.intValue();
 
-    if (maxH > MAX_COFACTOR) {
+    if (params.get(KEY_I) instanceof Long) {
+      minH = ((Long) params.get(KEY_I)).intValue();
+      maxH = minH;
```
ftence, line 179: There must be a difference of 1 between those two for the `for` loop to execute. I'd do `minH = ((Long) params.get(KEY_I)).intValue() - 1;`.

ftence (changes requested):

warp10/src/main/java/io/warp10/script/functions/ECRECOVER.java
hbs: h is the actual cofactor of the curve, it is a fixed value, i is an index between 0 and the cofactor.

warp10/src/main/java/io/warp10/script/functions/ECRECOVER.java
hbs: It's better if maxH is set to minH + 1, because the loop starts at minH.

ftence (approved):

PR #960, opened 2021-04-28
Add names to threads or thread factories

ftence:
Naming threads will simplify debugging.

`new CustomThreadFactory("...")` creates a ThreadFactory that has the same behavior, except for the naming, as `Executors.defaultThreadFactory()` which is used by default by `ThreadPoolExecutor`s.


hbs:
Shouln't the CustomThreadFactory class be added to the WarpScript jar?

PR #962, opened 2021-04-28
STRICTMAPPER on macros

randomboolean:
Make STRICTMAPPER accept macros as input to improve readability of some warpscripts


ftence (changes requested):
Macros used in `MAP` are not, strictly speaking, mappers. Using `STRICTMAPPER` for those macros is thus debatable. However, as I can't find a better name, we can stick to `STRICTMAPPER`.

ftence (approved):

PR #968, opened 2021-05-07
fix bugs that raised cryptic error messages in STL

randomboolean:
First bug fixed: when a subcycle series was empty, NaN values were pushed into the transient seasonal GTS, and then were propagated with the subsequent aggregations.

Second bug fixed: the aggregation did not take into account correctly missing values if input GTS was not filled.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -10265,7 +10265,7 @@ public class GTSHelper {
 
         int idx_t = 0;
         for (int idx = 0 ; idx < nonnull; idx++){
-          idx_t = Arrays.binarySearch(trend.ticks, idx_t, nonnull, gts.ticks[idx]);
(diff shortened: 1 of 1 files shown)


randomboolean:
> It seems there is a bug when there is one or more missing values, here is the code I used:
> 
> ```
> 1337 PRNG
> 
> NEWGTS
> 1 1000
> <%
>   'i' STORE
>   $i s NaN NaN NaN $i SRAND + $i 30.0 / PI * SIN 100 * + ADDVALUE
> %>
> FOR
> 
> [
>   SWAP
>   bucketizer.last
>   0 2 s 0
> ]
> BUCKETIZE
> 0 GET
> 
> 100 REMOVE DROP
> 
> { 
>   'PERIOD' 30
> }
> STL
> ```

Good catch that's another bug. 71e5c17 fixes it.

randomboolean:
Do not merge yet, there are still some bugs with empty buckets when ROBUSTNESS > 1

randomboolean:
> Do not merge yet, there are still some bugs with empty buckets when ROBUSTNESS > 1

a9e9f762d2c48f1c98d5e34ee12a8f4d778a0ef3 fixes it

ftence (approved):

PR #973, opened 2021-06-14
METASORT Extends behavior to attributes

steven-gueguen:


warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
```diff
@@ -24,158 +24,213 @@
...
       // Advance labels
-      //
-      
-      idx1++;
-      idx2++;
+      idx++;
     }
 
-    if (idx1 == idx2) {
+    //
```
hbs, line 116: The original loop condition had an issue, but the original idea was that a GTS with fewer labels than another one would appear before. Your change does not adhere to this as the GTS are considered equal when they have the same labels for the smaller set of labels of the two.

warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
```diff
@@ -24,158 +24,213 @@
...
+
+      if (0 != comp) {
+        return comp;
+      }
+
+      // Advance attributes
+      idx++;
+    }
+
+    return 0;
```
hbs, line 164: Ditto here, if we want to sort on attributes then GTS with fewer attributes should appear first

hbs (changes requested):

warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
```diff
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018  SenX S.A.S.
+//   Copyright 2018 - 2021  SenX S.A.S.
```
ftence, line 2: Prefer `2018-2021`, without spaces, to keep the same format as other files.

ftence (changes requested):
```
    if (null == o1) {
      return -1;
    }

    if (null == o2) {
      return 1;
    }
```
If both are `null`, it will not return 0 could you fix that too?

Same with names, just below.

```
    if (0 == o1.getLabelsSize() && 0 == o2.getLabelsSize()) {
      return 0;
    }
```
In that case, attributes are not checked even if they could change the order.

warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
steven-gueguen: Done

warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
steven-gueguen: Done

warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
steven-gueguen: Done

warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
```diff
@@ -70,10 +74,6 @@ public int compare(Metadata o1, Metadata o2) {
     // Names are identical, compare labels
     //
 
-    if (0 == o1.getLabelsSize() && 0 == o2.getLabelsSize()) {
-      return 0;
-    }
-
```
ftence: If both size are zero, then it will return -1.

warp10/src/main/java/io/warp10/script/functions/METASORT.java
```diff
@@ -84,20 +89,21 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-    
+
     if (fields.isEmpty()) {
-      Collections.sort((List<GeoTimeSerie>) top, META_COMPARATOR);
+      ((List<GeoTimeSerie>) top).sort(META_COMPARATOR);
     } else {
-      Collections.sort((List<GeoTimeSerie>) top, new Comparator<GeoTimeSerie>() {
-        private final Comparator<Metadata> INNER_COMP = new MetadataTextComparator(fields);
-        
+      boolean finalConsiderAttributes = considerAttributes;
```
ftence, line 96: Explicitly mark it as final.

ftence (changes requested):

warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
```diff
@@ -29,9 +31,11 @@
 public class MetadataTextComparator implements Comparator<Metadata> {
 
   private final List<String> fields;
+  private final boolean considerAttributes;
 
-  public MetadataTextComparator(List<String> fields) {
+  public MetadataTextComparator(List<String> fields, boolean considerAttributes) {
```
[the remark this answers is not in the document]
steven-gueguen: done

warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
steven-gueguen: If both size are zero, then it will check the attributes

warp10/src/main/java/io/warp10/script/functions/METASORT.java
steven-gueguen: done

warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
ftence: Are you sure?
```
    //
    // Names are identical, compare labels
    //
    int size1 = o1.getLabelsSize();
    int size2 = o2.getLabelsSize();

    if (0 == size1) {
      return -1;
    }
```

steven-gueguen: Ok, got it. My bad…
Fixed

warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
```diff
@@ -19,163 +19,237 @@
...
 /**
- * Sort Metadata according to text order (on name then on labels)
- * 
+ * Sort Metadata according to field values (labels, attributes or classname if null provided)
  */
 public class MetadataTextComparator implements Comparator<Metadata> {
 
   private final List<String> fields;
-  
+  private final boolean considerAttributes;
```
ftence, line 34: `considerAttributes` is not taken into account when there `fields` is `null` or empty.
This is totally OK because I don't see why metadata ordering would leave order based on attributes unspecified.

However, for clarity I would add a comment for this field clearly stating that. Maybe renaming the variable `fieldsConsiderAttributes` would made it even clearer.

warp10/src/main/java/io/warp10/script/functions/LASTSORT.java
```diff
@@ -36,7 +36,7 @@
 
   private static final Comparator<GeoTimeSerie> LAST_COMPARATOR = new Comparator<GeoTimeSerie>() {
     
-    private final Comparator<Metadata> INNER_COMP = new MetadataTextComparator(null);
+    private final Comparator<Metadata> INNER_COMP = new MetadataTextComparator(null, false);
```
ftence, line 39: Remove `false` as it seems to indicate this sort does not take into account attributes, which is not the case (see previous comment).

warp10/src/main/java/io/warp10/script/functions/METASORT.java
```diff
@@ -35,23 +34,29 @@
 public class METASORT extends NamedWarpScriptFunction implements WarpScriptStackFunction {
 
   public static final Comparator<GeoTimeSerie> META_COMPARATOR = new Comparator<GeoTimeSerie>() {
-    
-    private final Comparator<Metadata> INNER_COMP = new MetadataTextComparator(null);
-    
+
+    private final Comparator<Metadata> INNER_COMP = new MetadataTextComparator(null, false);
```
ftence, line 38: Also remove `false`.

ftence (changes requested):

warp10/src/main/java/io/warp10/continuum/gts/MetadataTextComparator.java
steven-gueguen: Ok, renamed and commented

warp10/src/main/java/io/warp10/script/functions/METASORT.java
steven-gueguen: done

warp10/src/main/java/io/warp10/script/functions/LASTSORT.java
steven-gueguen: done

ftence (approved):

PR #974, opened 2021-06-15
Initial commit of POLYFIT / POLYFUNC

hbs:


randomboolean:
Does POLYFIT need to be constrained to bucket indices ?
Maybe we can add the option to fit coefficients wrt timestamps if it finds a boolean flag set at true with the syntax `true POLYFIT` ? (and accept that with high orders there would be overflow errors)

randomboolean:
A function that makes a weighted polynomial fit, using a second input numerical GTS could be useful too (for instance to give more weight to more recent values).

warp10/src/main/java/io/warp10/script/functions/POLYFIT.java
```diff
@@ -0,0 +1,119 @@
...
+
+    //
+    // Limit the number of iterations to twice the number of datapoints we attempt to fit
+    //
+
+    fitter.withMaxIterations(n * 2);
+
+    double[] coeffs = fitter.fit(points.toList());
+
+    stack.push(Arrays.asList(coeffs));
```
ftence, line 115: This will leave an Arrays.ArrayList with a single double[] element in it, prefer:

```
    ArrayList<Double> coeffsList = new ArrayList<>(coeffs.length);
    for(double coeff: coeffs) {
      coeffsList.add(coeff);
    }

    stack.push(coeffsList);
```

warp10/src/main/java/io/warp10/script/functions/POLYFUNC.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    return new Object[] { tick, locations[0], elevations[0], this.func.value(tick) };
+  }
+
+  @Override
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
+    try {
+      SNAPSHOT.addElement(sb, Arrays.asList(this.func.getCoefficients()));
```
ftence, line 163: Same problem with Arrays.asList, prefer:

```
      double[] coeffs = this.func.getCoefficients();
      ArrayList<Double> coeffsList = new ArrayList<>(coeffs.length);
      for(double coeff: coeffs) {
        coeffsList.add(coeff);
      }

      SNAPSHOT.addElement(sb, coeffsList);
```

ftence (changes requested):
Functions are not added to `WarpScriptLib`.

warp10/src/main/java/io/warp10/script/functions/POLYFIT.java
```diff
@@ -0,0 +1,119 @@
+//
+//   Copyright 2018-2021  SenX S.A.S.
```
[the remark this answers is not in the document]
hbs: nope, the code was indeed started in 2018

warp10/src/main/java/io/warp10/script/functions/POLYFUNC.java
hbs: What is the point of copying the list to create its SNAPSHOT?

warp10/src/main/java/io/warp10/script/functions/POLYFUNC.java
```diff
@@ -0,0 +1,171 @@
...
+    if (top instanceof List) {
+      List<Object> list = new ArrayList<Object>(((List) top).size());
+      for (Object elt: (List) top) {
+        if (!(elt instanceof Number)) {
+          throw new WarpScriptException(getName() + " can only be applied to numerical values.");
+        }
+        list.add(this.func.value(((Number) elt).doubleValue()));
+      }
+      stack.push(list);
+    } else if (top instanceof GeoTimeSerie) {
```
randomboolean, line 98: To be coherent with POLYFIT, this function may also accept a `useBucketId` argument and raise an error if the argument below is not a bucketized numerical GTS.

hbs: Nope, the function is created by POLYFIT based on some arguments, but there is no reason to constraint its use.

randomboolean: The function would just have an additional optional argument. It's use would not be constrained.

ftence (changes requested):

warp10/src/main/java/io/warp10/script/functions/POLYFUNC.java
randomboolean: yes that is why I don't understand why the function created by POLYFUNC should not also have that optional parameter when applied on bucketized GTS. It doesn't interfere with when this created function is applied on other input types.

warp10/src/main/java/io/warp10/script/functions/POLYFUNC.java
```diff
@@ -0,0 +1,171 @@
...
+      stack.push(out);
+    } else if (top instanceof GTSEncoder) {
+      GTSDecoder decoder = ((GTSEncoder) top).getDecoder();
+      GTSEncoder encoder = new GTSEncoder(decoder.getBaseTimestamp());
+      encoder.setMetadata(decoder.getMetadata());
+
+      while(decoder.next()) {
+        long ts = decoder.getTimestamp();
+        double value = this.func.value(ts);
+        encoder.addValue(ts, decoder.getLocation(), decoder.getElevation(), value);
```
ftence, line 126: As reported by Travis CI:
```
io/warp10/script/functions/POLYFUNC.java:126: error: unreported exception IOException; must be caught or declared to be thrown
        encoder.addValue(ts, decoder.getLocation(), decoder.getElevation(), value);
```

warp10/src/main/java/io/warp10/script/functions/POLYFUNC.java
hbs: Ok, got what you meant! Sorry for my slow brain today!

warp10/src/main/java/io/warp10/script/functions/POLYFUNC.java
```diff
@@ -0,0 +1,198 @@
...
+      super(name);
+    }
+
+    @Override
+    public Object apply(WarpScriptStack stack) throws WarpScriptException {
+      Object top = stack.pop();
+
+      boolean useBucketId = false;
+
+      if (top instanceof Boolean) {
```
randomboolean, line 62: Well what I had in mind was that this check would rather not done by the builder, but instead by the function that is created by POLYFUNC, so it would be more flexible (the function could then still be applied on other input type than bucketized GTS)

warp10/src/main/java/io/warp10/script/functions/POLYFUNC.java
```diff
@@ -0,0 +1,191 @@
...
+      stack.push(list);
+    } else if (top instanceof GeoTimeSerie) {
+      GeoTimeSerie gts = (GeoTimeSerie) top;
+      int n = GTSHelper.nticks(gts);
+      GeoTimeSerie out = new GeoTimeSerie(n);
+      if (GTSHelper.isBucketized(gts)) {
+        long lastbucket = GTSHelper.getLastBucket(gts);
+        long bucketspan = GTSHelper.getBucketSpan(gts);
+        for (int i = 0; i < n; i++) {
+          long ts = this.useBucketId ? n - 1 - i : lastbucket - i * bucketspan;
```
ftence, line 118: `useBucketId` instead of `this.useBucketId`.

warp10/src/main/java/io/warp10/script/functions/POLYFUNC.java
```diff
@@ -0,0 +1,185 @@
...
+        } catch (IOException ioe) {
+          throw new WarpScriptException(getName() + " error while generating ENCODER.");
+        }
+      }
+      stack.push(encoder);
+    } else if (top instanceof Number) {
+      double value = this.func.value(((Number) top).doubleValue());
+
+      stack.push(value);
+    } else if (!(top instanceof Number)) {
```
ftence, line 143: Can be simplified to `else`.

ftence (changes requested):

ftence (approved):

ftence (approved):

PR #976, opened 2021-06-18
Timebox

randomboolean:
Add capability to raise warpscript.timebox.maxtime configuration limit.
Add builder for timeboxing functions when registering them, with support for capabilities.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/Configuration.java
+++ b/warp10/src/main/java/io/warp10/continuum/Configuration.java
@@ -110,6 +110,11 @@ public class Configuration {
   public static final String CONFIG_WARPSCRIPT_TIMEBOX_MAXTIME = "warpscript.timebox.maxtime";
 
   /**
+   * Capability name used for extending timebox maxtime, if set.
+   */
+  public static final String MAXTIME_EXTENSION_ALLOWANCE_CAPNAME = "timebox.maxtime.extension.capname";
+
+  /**
    * Comma separated list of WarpScriptExtension classes to instantiate to modify
    * the defined WarpScript functions.
    */
--- a/warp10/src/main/java/io/warp10/continuum/Configuration.java
+++ b/warp10/src/main/java/io/warp10/continuum/Configuration.java
@@ -112,7 +112,7 @@ public class Configuration {
   /**
    * Capability name used for extending timebox maxtime, if set.
    */
-  public static final String MAXTIME_EXTENSION_ALLOWANCE_CAPNAME = "timebox.maxtime.extension.capname";
+  public static final String MAXTIME_EXTENSION_ALLOWANCE_CAPNAME = "warpscript.timebox.maxtime.capname";
 
   /**
    * Comma separated list of WarpScriptExtension classes to instantiate to modify
--- a/warp10/src/main/java/io/warp10/script/functions/TIMEBOX.java
+++ b/warp10/src/main/java/io/warp10/script/functions/TIMEBOX.java
@@ -34,7 +34,7 @@ import io.warp10.script.WarpScriptStack.Macro;
 import io.warp10.script.WarpScriptStackFunction;
 import io.warp10.warp.sdk.Capabilities;
 
-public class TIMEBOX<F extends NamedWarpScriptFunction & WarpScriptStackFunction > extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+public class TIMEBOX extends NamedWarpScriptFunction implements WarpScriptStackFunction {
   
   /**
    * Default timeboxing is 30s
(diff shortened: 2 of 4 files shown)


warp10/src/main/java/io/warp10/script/functions/TIMEBOX.java
```diff
@@ -44,30 +44,82 @@
...
+   */
+  private final WarpScriptStackFunction function;
+
+  public TIMEBOX(String name, WarpScriptStackFunction function) {
+    this(name, function, false);
   }
-  
-  public TIMEBOX(String name) {
+
+  public TIMEBOX(String name, WarpScriptStackFunction function, Boolean requireCapability) {
```
hbs: This is not needed, functions can simply be included in a macro.

randomboolean: but macros don't have a name and must share the same maxtime

hbs: There is no need to allow setting the maxtime per function. The wrapped functions would also not be snapshotable otherwise the snapshot would allow the setting of the various parameters which would expose to security risks.

randomboolean: And also this simplifies registration. For example, suppose an extension wants to timebox the function DBSCAN, here we could simply do:
`functions.add("DBSCAN", TIMEBOX(new DBSCAN("DBSCAN"), "extension.maxtime.capname")`
rather than building the macro with capability check for each function

hbs: As I've said this would require to allow the explicit setting of sensitive parameters, so this is not a feature that should be included in the core of Warp 10. It may be in an extension that users decide to explicitly allow, but not in the core.

randomboolean: I don't see why it can't be snapshoted, the snapshot would just be the name of the wrapped function. The fact that maxtime limits would be different on other instances, or that the capability that control it is different is not an issue.

warp10/src/main/java/io/warp10/script/functions/TIMEBOX.java
```diff
@@ -45,31 +45,46 @@
...
-  
   static {
     TIMEBOX_MAXTIME = Long.parseLong(WarpConfig.getProperty(Configuration.CONFIG_WARPSCRIPT_TIMEBOX_MAXTIME, Long.toString(DEFAULT_TIMEBOX_MAXTIME)));
   }
-  
+
+  /**
+   * Allowance capability to raise TIMEBOX_MAXTIME
+   */
+  private static final String SPECIAL_ALLOWANCE_CAPNAME = Configuration.CONFIG_WARPSCRIPT_TIMEBOX_MAXTIME;
```
hbs, line 55: Use a specific configuration key which will contain the actual capability name to use for extending timebox maxtime. If such key is not set then no capability will be checked.

hbs (changes requested):

warp10/src/main/java/io/warp10/script/functions/TIMEBOX.java
```diff
@@ -45,31 +45,46 @@
...
-  
+
+  /**
+   * Allowance capability to raise TIMEBOX_MAXTIME
+   */
+  private static final String SPECIAL_ALLOWANCE_CAPNAME = WarpConfig.getProperty(Configuration.MAXTIME_EXTENSION_ALLOWANCE_CAPNAME);
+
+  /**
+   * No wrapped function: TIMEBOX will expect a macro and a maxtime long on top of the stack.
+   */
```
ftence, line 59: Comment to be removed.

ftence (changes requested):

warp10/src/main/java/io/warp10/continuum/Configuration.java
```diff
@@ -109,6 +109,11 @@
    */
   public static final String CONFIG_WARPSCRIPT_TIMEBOX_MAXTIME = "warpscript.timebox.maxtime";
 
+  /**
+   * Capability name used for extending timebox maxtime, if set.
+   */
+  public static final String MAXTIME_EXTENSION_ALLOWANCE_CAPNAME = "timebox.maxtime.extension.capname";
```
ftence, line 115: Change configuration key name to make it coherent with `warpscript.timebox.maxtime`. Use `warpscript.timebox.maxtime.capname` for instance.

ftence: Also add it, commented, to the configuration templates to make it more visible.

ftence (changes requested):

ftence (approved):

PR #977, opened 2021-06-18
Support for inline macro as value in configurations

randomboolean:
- Configuration value can be a macro using syntax
`warp.configuration = <% 'this is a macro' %>`

- configuration shadow.macro.FUNC can parse an inline macro

- fix a bug related to configuration shadow.macro.FUNC

### DIFF
--- a/warp10/src/main/java/io/warp10/WarpConfig.java
+++ b/warp10/src/main/java/io/warp10/WarpConfig.java
@@ -317,7 +317,7 @@ public class WarpConfig {
         String key = tokens[0];
         String macro = line.substring(key.length() + 1);
         macro = macro.trim();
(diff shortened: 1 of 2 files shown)


randomboolean:
I'm not sure it's a good idea to allow the possibility to write inline macros that can be indefinitely long.
Maybe we should restrict to a certain number of statements ?
Also, instead of macros, it could be explicit warscript code (not enclosed in <% %> but detectable with another convention).

hbs:
The change cannot be made global like this, it would have many implications in different places throughout the code base and in terms of operations could lead to errors because WarpScript execution cannot happen before some of the configuration keys are actually set.

Closing the PR for this reason.

PR #978, opened 2021-06-22
Added stack attribute to indicate current execution is timeboxed

hbs:


warp10/src/main/java/io/warp10/script/functions/TIMEBOX.java
```diff
@@ -35,44 +35,47 @@
...
+
     if (!(top instanceof Macro)) {
       throw new WarpScriptException(getName() + " operates on a macro.");
     }
 
     final Macro macro = (Macro) top;
     final WarpScriptStack fstack = stack;
-    
+
+    Object timeboxed = Boolean.TRUE.equals(fstack.getAttribute(WarpScriptStack.ATTRIBUTE_TIMEBOXED));
```
ftence, line 76: `Boolean timeboxed` would be more readable.

ftence (changes requested):

warp10/src/main/java/io/warp10/script/functions/TIMEBOX.java
randomboolean: If that makes any sense to store this information, it might be preferable to store a Long with maxtime value.

ftence (approved):

PR #980, opened 2021-06-25
Modified scheduling so it uses System.nanoTime

hbs:


warp10/src/main/java/io/warp10/script/ScriptRunner.java
```diff
@@ -390,16 +390,16 @@ public int compare(String o1, String o2) {
...
             runnables.add(script);
           } else {
-            Long period = scriptAndPeriod.getValue();
-            long schedat = System.currentTimeMillis();
+            Long period = 1000000L * scriptAndPeriod.getValue();
+            long schedat = System.nanoTime();
 
             if (0 != schedat % period) {
               schedat = schedat - (schedat % period) + period;
             }
```
ftence, line 398: If I'm not mistaken, this is to align with epoch+N*period. As the origin of nanoTime() is unknown, this is not possible to do that like this.

ftence (changes requested):
Also, loosely related, in ScriptRunnerConsumerFactory, in case the clocks are out of sync between servers
```
              //
              // Check if running is overdue
              //
              
              long now = System.currentTimeMillis();
              
              if (request.getScheduledAt() + request.getPeriodicity() >= now) {
                continue;
              }
```
may skip a run even if it should not.

ftence (approved):

pi-r-p:
Tested successfully on an imx8.

PR #982, opened 2021-06-28
fix a npe in WarpScript exception message

### LINKED ISSUE #981: maxops error regression
Hi,

Between 2.7 and 2.8, there is a regression in the maxops reached error message.

For example:

```
'validtoken' AUTHENTICATE
6000 MAXOPS

[
  1 200 <% DROP NEWGTS %> FOR
] 'list' STORE

$list
<%
  'l' STORE
  $list
  <%
    'j' STORE
  %> FOREACH
%> FOREACH
```

leads to `Exception at '%> =>FOREACH<=' in section [TOP] (Exception at 'FOREACH' in section '[TOP]')`.

pi-r-p:
fix #981

### DIFF
--- a/warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
+++ b/warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
@@ -985,7 +985,10 @@ public class MemoryWarpScriptStack implements WarpScriptStack, Progressable {
         String name = macro.getName();
         String section = (String) this.getAttribute(WarpScriptStack.ATTRIBUTE_SECTION_NAME);
         Object statement = macro.get(i);
-        String statementString = statement.toString();
+        String statementString = "";
+        if (null != statement) {
+          statementString = statement.toString();
+        }
         // For NamedWarpScriptFunction, toString is used for snapshotting. Getting the name is better to generate
         // a clear error message.
         if(statement instanceof NamedWarpScriptFunction) {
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -985,7 +985,10 @@ public void exec(Macro macro) throws WarpScriptException {
         String name = macro.getName();
         String section = (String) this.getAttribute(WarpScriptStack.ATTRIBUTE_SECTION_NAME);
         Object statement = macro.get(i);
-        String statementString = statement.toString();
+        String statementString = "";
```
hbs, line 988: Do we want to have the empty String is statement is null or null?

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -984,9 +984,10 @@ public void exec(Macro macro) throws WarpScriptException {
       } else {
         String name = macro.getName();
         String section = (String) this.getAttribute(WarpScriptStack.ATTRIBUTE_SECTION_NAME);
-        Object statement = macro.get(i);
         String statementString = "";
-        if (null != statement) {
+        Object statement = null;
+        if (i >= macro.size()) {
+          statement = macro.get(macro.size() - 1);
```
hbs, line 990: The statementString = statement.toString() should be outside of the conditional and statement should be initialized to macro.get(i), not to null.

hbs: And statementString should be set to String.valueOf(statement)

ftence (approved):

PR #983, opened 2021-07-01
Make MVEXTRACT push ArrayList of lat/lon on the stack

ftence:
Currently an `Arrays.ArrayList` is pushed on the stack which is not consistent with all other functions.


pi-r-p:
take me a while to understand the point. arrays.aslist return a fixed size list. So the issue was :
```
NEWGTS 0 4.5 25.2 42  true ADDVALUE MVLOCATIONS
0 GET [ 45 ] +!
```
leads to an unexpected exception. `Exception at '[ 45 ] =>+!<=' in section [TOP]`

PR #984, opened 2021-07-05
Fix MOTIONSPLIT to handle points with no location

ftence:
When used on a GTS with some points with location and some without, MOTIONSPLIT may miss a split.

For instance:
```
// @preview gts

NEWGTS
1 3 // 3 points WITH location
<% 'i' STORE $i m $i $i $i $i ADDVALUE %>
FOR

4 6 // 3 points WITHOUT location
<% 'i' STORE $i m NaN NaN NaN $i ADDVALUE %>
FOR

7 9 // 3 points WITH location
<% 'i' STORE $i m $i $i $i $i ADDVALUE %>
FOR

{ 'distance.split' 1 }
MOTIONSPLIT
```
In that case, the points with timestamp 3 and 7 are in the same split even though the distance between them is well above 1 meter, see the image below.
image

This PR fix that, see the image below
image

These changes pass the MOTIONSPLIT unit tests.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/MOTIONSPLIT.java
+++ b/warp10/src/main/java/io/warp10/script/functions/MOTIONSPLIT.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2019  SenX S.A.S.
+//   Copyright 2019-2021  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


hbs (approved):

pi-r-p:
Did you add a unit test to cover this case ?

ftence:
Unit test for this case added in WarpScript doc.

pi-r-p (approved):

PR #986, opened 2021-07-06
Added support for setting capabilities via HTTP header

hbs:


warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
```diff
@@ -193,19 +194,32 @@ public void handle(String target, Request baseRequest, HttpServletRequest req, H
         }
         stack.store(expose, headers);
       }
-            
+
       //
-      // Execute the bootstrap code
+      // Execute the bootstrap code. This is done before adding capabilities.
```
ftence, line 199: Update copyright notice.

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
```diff
@@ -193,19 +194,32 @@ public void handle(String target, Request baseRequest, HttpServletRequest req, H
...
+      //
+      // Add capabilities specified in the capabilities header
+      //
+
+      String capabilities = req.getHeader(Constants.getHeader(Configuration.HTTP_HEADER_CAPABILITIES));
+
+      if (null != capabilities) {
+        String[] tokens = capabilities.split(",");
+        for (String token: tokens) {
+          Capabilities.add(stack, token.trim());
```
ftence, line 213: This makes mixing overlapping capabilities from different tokens impossible. For instance, if we have token1 with cap1=foo and cap2=foo and token 2 with cap1=bar and cap2=bar, it will not be possible to have cap1=foo and cap2=bar.

For now, I don't see a simple solution to this problem which should very rarely happen.

ftence (changes requested):

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
hbs: Via the header mechanism yes, inside of a WarpScript snippet you can be picky and mix capabilities from various tokens.

ftence (approved):

PR #989, opened 2021-07-09
Token.fetchattributes

hbs:


warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1485,4 +1537,154 @@ public static Long getTimestamp(Object timestampRepresentation, String timestamp
...
+        if (null == end) {
+          if (null != params && params.get(PARAM_END) instanceof Long) {
+            quietafter = ((Long) params.get(PARAM_END)).longValue() + offset;
+          } else {
+            quietafter = now + offset;
+          }
+        } else {
+          quietafter = end + offset;
+        }
+        quietafter = now + offset;
```
ftence, line 1608: If you want to keep the logic above this line, remove this line as it overrides the value set above in the conditional blocks.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1485,4 +1537,154 @@ public static Long getTimestamp(Object timestampRepresentation, String timestamp
...
+        long offset = DURATION.parseDuration(ref, value, true, true);
+        if (null == end) {
+          if (null != params && params.get(PARAM_END) instanceof Long) {
+            quietafter = ((Long) params.get(PARAM_END)).longValue() + offset;
+          } else {
+            quietafter = now + offset;
+          }
+        } else {
+          quietafter = end + offset;
+        }
```
ftence, line 1607: I don't see the point of using `end` as a reference for `quietafter` constrained using a duration. As I see it, one would want to allow access to GTSs not active in the last *hh* hours for instance, independently of `end`.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1485,4 +1537,154 @@ public static Long getTimestamp(Object timestampRepresentation, String timestamp
...
+        long offset = DURATION.parseDuration(ref, value, true, true);
+        if (null == end) {
+          if (null != params && params.get(PARAM_END) instanceof Long) {
+            activeafter = ((Long) params.get(PARAM_END)).longValue() + offset;
+          } else {
+            activeafter = now + offset;
+          }
+        } else {
+          activeafter = end + offset;
+        }
```
ftence, line 1630: Same as for `quietafter`, only `now` as a reference seems useful to me.

ftence (changes requested):

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1485,4 +1537,167 @@ public static Long getTimestamp(Object timestampRepresentation, String timestamp
...
+      } else if ("0".equals(value)) {
+        if (null == end) {
+          if (null != params && params.get(PARAM_END) instanceof Long) {
+            quietafter = ((Long) params.get(PARAM_END)).longValue();
+          } else {
+            quietafter = now;
+          }
+        } else {
+          quietafter = end;
+        }
```
ftence, line 1617: Won't it do the same as setting `PT0S`?

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -1485,4 +1537,167 @@ public static Long getTimestamp(Object timestampRepresentation, String timestamp
...
+    String value = tokenattr.get(Constants.TOKEN_ATTR_END);
+
+    Long end = null;
+
+    if (null != value) {
+      value = value.trim();
+
+      if (value.startsWith("P")) {
+        long offset = DURATION.parseDuration(ref, value, true, true);
+        end = now + offset;
```
ftence, line 1563: `DURATION.parseDuration(ref, value, true, true);` will return a positive duration for `DURATION.parseDuration(new Instant(), "PT1S", true, true);`, with or without https://github.com/hbs/warp10-platform/pull/10.

It seems to me all `+ offset` should be `- offset`.

ftence (changes requested):

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -235,6 +236,57 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+    }
+
+    if (null != tokenattr.get(PARAM_ACTIVE_AFTER)) {
+      long activity = (long) tokenattr.get(PARAM_ACTIVE_AFTER);
+      params.put(PARAM_ACTIVE_AFTER, activity);
+    }
+
+    if (null != tokenattr.get(PARAM_COUNT)) {
+      long count = (long) tokenattr.get(PARAM_COUNT);
+      if ((long) params.get(PARAM_COUNT) > count) {
```
ftence, line 271: `params.get(PARAM_COUNT)` can be null, which may cause a NPE.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -235,6 +236,57 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+        params.put(PARAM_COUNT, count);
+      }
+    }
+
+    if (null != tokenattr.get(PARAM_BOUNDARY_PRE)) {
+      long boundary = (long) tokenattr.get(PARAM_BOUNDARY_PRE);
+      if ((long) params.get(PARAM_BOUNDARY_PRE) > boundary) {
```
ftence, line 278: `params.get(PARAM_BOUNDARY_PRE)` can be null, which may cause a NPE.

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -235,6 +236,57 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+        params.put(PARAM_BOUNDARY_PRE, boundary);
+      }
+    }
+
+    if (null != tokenattr.get(PARAM_BOUNDARY_POST)) {
+      long boundary = (long) tokenattr.get(PARAM_BOUNDARY_POST);
+      if ((long) params.get(PARAM_BOUNDARY_POST) > boundary) {
```
ftence, line 285: `params.get(PARAM_BOUNDARY_POST)` can be null, which may cause a NPE.

ftence (changes requested):

ftence (approved):

PR #991, opened 2021-07-09
Add configuration to set LevelDB cache and checksums use during deletions

hbs:


ftence (changes requested):

etc/conf.templates/standalone/10-leveldb.conf.template
```diff
@@ -129,3 +129,13 @@ leveldb.compression.type = SNAPPY
 // Defaults to 0, meaning never disable the block cache
 //
 leveldb.blockcache.gts.threshold = 0
+
+//
+// Should the LevelDB cache be filled when scanning key space for deletes? Defaults to false.
+//
+#standalone.delete.leveldb.fillcache = false
```
ftence, line 136: `leveldb.delete.fillcache` would be more coherent with the other keys.

etc/conf.templates/standalone/10-leveldb.conf.template
```diff
@@ -129,3 +129,13 @@ leveldb.compression.type = SNAPPY
...
+
+//
+// Should LevelDB checksums be verified when scanning key space for deletes? Defaults to true.
+//
+#standalone.delete.leveldb.verifychecksums = true
```
ftence, line 141: `leveldb.delete.verifychecksums`

ftence (changes requested):

ftence (approved):

PR #994, opened 2021-07-12
Add compatibility with newer thrift version

steven-gueguen:


hbs:
This does not work when using Thrift compiler 0.11.0

steven-gueguen:
Close this PR as theThrift is constrained by HBase.
This should be reconsidered if we update the HBase version.

PR #1001, opened 2021-08-04
force a flush of underlying PrintStream object in LOGMSG, STDOUT and STDERR

randomboolean:
To flush stdout when using LOGMSG or STDOUT.

It might be better to add a function STDOUT.FLUSH, or to flush per default.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/ext/debug/LOGMSG.java
+++ b/warp10/src/main/java/io/warp10/script/ext/debug/LOGMSG.java
@@ -31,18 +31,9 @@ public class LOGMSG extends NamedWarpScriptFunction implements WarpScriptStackFu
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
     Object top = stack.pop();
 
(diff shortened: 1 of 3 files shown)


ftence:
It might even be better to add the SYSOUT.FLUSH function that simply does a `System.out.flush()`. It adds the functionality you want but it is more flexible if you need to flush less occasionally.

Your current solution changes the behavior of the functions if they are used to display the value of a Boolean and would most probably make the code fail in that situation. It is best to avoid that.

hbs:
We can simply add a call to flush, it won't have much effect on performance and is cleaner than exposing a function which can call flush.

ftence:
This is indeed a nice solution as these functions are meant for debug so a small performance deterioration is not a problem.

As well as `LOGMSG` and `STDOUT`, `STDERR` should also be modified.

ftence:
Please update the title of PR to reflect the change as it will be used by the changelog generator.

PR #1009, opened 2021-08-18
/delete to allow the deletion of all points while keeping metas

ftence:


warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
```diff
@@ -1525,36 +1525,36 @@ public void handleDelete(String target, Request baseRequest, HttpServletRequest
...
-        }
+
+      // Both or neither start and end must be specified
+      if (null == startstr ^ null == endstr) {
+        response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Both " + Constants.HTTP_PARAM_START + " and " + Constants.HTTP_PARAM_END + " should be defined.");
+        return;
+      }
+
+      // Parse start and end parameters
+      // The code below relies on the fact both startstr and endstr are non-null so the check both.
```
hbs, line 1536: Maybe rephrase, unclear as is.

PR #1010, opened 2021-08-19
Add LFILTER and MFILTER

ftence:
Introduces two new functions `LFILTER` and `MFILTER` with a lot of similarities with `LMAP` and `MMAP`. Instead of mapping the element, those functions filter the element according to a boolean-returning macro.

Like `LMAP` and `MMAP`, is is possible to push the index of the element or not. It is pushed by default.

While this logic is achievable through a macro, those functions execute in 2/3 of the time of the macro version.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/LFILTER.java
+++ b/warp10/src/main/java/io/warp10/script/functions/LFILTER.java
@@ -1,0 +1,81 @@
+//
+//   Copyright 2021  SenX S.A.S.
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+
+package io.warp10.script.functions;
+
+import io.warp10.script.NamedWarpScriptFunction;
+import io.warp10.script.WarpScriptException;
+import io.warp10.script.WarpScriptStack;
+import io.warp10.script.WarpScriptStack.Macro;
+import io.warp10.script.WarpScriptStackFunction;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class LFILTER extends NamedWarpScriptFunction implements WarpScriptStackFunction {
+
+  public LFILTER(String name) {
+    super(name);
+  }
+
+  @Override
+  public Object apply(WarpScriptStack stack) throws WarpScriptException {
+
+    Object top = stack.pop();
+
+    boolean pushIndex = true;
+    if (top instanceof Boolean) {
+      pushIndex = (Boolean) top;
+      top = stack.pop();
+    }
+
+    if (!(top instanceof Macro)) {
+      throw new WarpScriptException(getName() + " expects a macro on top of the stack.");
+    }
+
+    Object list = stack.pop();
+
+    if (!(list instanceof List)) {
+      throw new WarpScriptException(getName() + " expects a list below the macro on top of the stack.");
+    }
+
+    int n = ((List) list).size();
+    ArrayList<Object> result = new ArrayList<Object>(n);
+
+    for (int i = 0; i < n; i++) {
+      Object element = ((List) list).get(i);
+      stack.push(element);
+      if (pushIndex) {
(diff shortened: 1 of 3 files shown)


randomboolean:
How does it compare with the following line ?
`[ $list <% @some-filter-condition  <% DROP %> IFT %>   FOREACH ]`

ftence:
While this is a nice python-like list comprehension, this should be avoided because it can easily fail due to the stack depth limit.

The LFILTER function was compared to
```
[] SWAP
<% 
  DUP 2 % 0 ==
  <% +! %>
  <% DROP %>
  IFTE
%>
FOREACH
```

warp10/src/main/java/io/warp10/script/functions/LFILTER.java
```diff
@@ -0,0 +1,81 @@
...
+    Object top = stack.pop();
+
+    boolean pushIndex = true;
+    if (top instanceof Boolean) {
+      pushIndex = (Boolean) top;
+      top = stack.pop();
+    }
+
+    if (!(top instanceof Macro)) {
+      throw new WarpScriptException(getName() + " expects a macro on top of the stack.");
```
hbs, line 46: Avoid referring to the stack in error messages.

warp10/src/main/java/io/warp10/script/functions/LFILTER.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    Object list = stack.pop();
+
+    if (!(list instanceof List)) {
+      throw new WarpScriptException(getName() + " expects a list below the macro on top of the stack.");
```
hbs, line 52: Ditto ... "operates on a LIST"

warp10/src/main/java/io/warp10/script/functions/MFILTER.java
```diff
@@ -0,0 +1,92 @@
...
+    Object top = stack.pop();
+
+    boolean pushIndex = true;
+    if (top instanceof Boolean) {
+      pushIndex = (Boolean) top;
+      top = stack.pop();
+    }
+    
+    if (!(top instanceof Macro)) {
+      throw new WarpScriptException(getName() + " expects a macro on top of the stack.");
```
hbs, line 46: Ditto above

warp10/src/main/java/io/warp10/script/functions/MFILTER.java
```diff
@@ -0,0 +1,92 @@
...
+      if (pushIndex) {
+        stack.push((long) i);
+      }
+
+      stack.exec(macro);
+
+      Object o = stack.pop();
+
+      if(!(o instanceof Boolean)) {
+        throw new WarpScriptException(getName() + " expect the filter macro to return a BOOLEAN for each element.");
```
hbs, line 78: expects

warp10/src/main/java/io/warp10/script/functions/LFILTER.java
```diff
@@ -0,0 +1,81 @@
...
+      stack.push(element);
+      if (pushIndex) {
+        stack.push((long) i);
+      }
+      stack.exec((Macro) top);
+
+      Object o = stack.pop();
+
+      if (!(o instanceof Boolean)) {
+        throw new WarpScriptException(getName() + " expect the filter macro to return a BOOLEAN for each element.");
```
hbs, line 69: expects

warp10/src/main/java/io/warp10/script/functions/LFILTER.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    Object list = stack.pop();
+
+    if (!(list instanceof List)) {
+      throw new WarpScriptException(getName() + " expects a list below the macro on top of the stack.");
+    }
+
+    int n = ((List) list).size();
+    ArrayList<Object> result = new ArrayList<Object>(n);
```
hbs, line 56: Allocating n elements may be inefficient because we do not know ahead of time the proportion of elements which will be retained.

warp10/src/main/java/io/warp10/script/functions/MFILTER.java
```diff
@@ -0,0 +1,92 @@
...
+    
+    top = stack.pop();
+    
+    if (!(top instanceof Map)) {
+      throw new WarpScriptException(getName() + " expects a MAP below the macro on top of the stack.");
+    }
+
+    Map<?, ?> map = (Map) top;
+
+    LinkedHashMap<Object, Object> result = new LinkedHashMap<Object, Object>(map.size());
```
hbs, line 59: Allocating an identical number of elements may be inefficient because we do not know ahead of time the proportion of elements which will be retained.

hbs (changes requested):

hbs (approved):

PR #1014, opened 2021-08-25
Add time bounding to /exec requests

hbs:


warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
```diff
@@ -285,6 +303,66 @@ public void handle(String target, Request baseRequest, HttpServletRequest req, H
...
+            return;
+          }
+        }
+
+        //
+        // Check the configured maxtime as the upper limit unless a valid token was specified as timebox
+        //
+
+        if (!fromToken && 0 != MAXTIME && maxtime > 0) {
+          maxtime = Math.min(maxtime, MAXTIME);
```
ftence, line 331: If `maxtime < 0`, this will leave it negative thus removing the timebox, even if `MAXTIME > 0`.

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
```diff
@@ -285,6 +303,66 @@ public void handle(String target, Request baseRequest, HttpServletRequest req, H
...
+          try {
+            timeLimit = Long.valueOf(Capabilities.get(stack, TIMEBOX.TIMEBOX_MAXTIME_CAPNAME)) * Constants.TIME_UNITS_PER_MS;
+          } catch (NumberFormatException nfe) {
+          }
+        }
+      }
+
+      if (maxtime > timeLimit) {
+        maxtime = timeLimit;
+      }
```
ftence, line 354: I thought the token capability should increase the time limit, not reduce it.

I think the logic should be:
```
long maxtime = MAXTIME;
long maxtimeHeader = // parse from header
long maxtimeCapability = // parse from capabilties

if (maxtimeCapability > 0) {
  maxtime = Math.max(maxtime, maxtimeCapability);
} else {
  maxtime = 0;
}

if (maxtimeHeader > 0) {
  maxtime = Math.min(maxtime, maxtimeHeader);
}
```

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
```diff
@@ -321,12 +399,22 @@ public void handle(String target, Request baseRequest, HttpServletRequest req, H
...
       //
       // Make sure stack is balanced
       //
 
       stack.checkBalanced();
 
+      if (maxtime > 0) {
+        stack.push(maxtime);
+        TIMEBOX.apply(stack);
+        stack.handleSignal();
```
ftence, line 415: I don't see why you need to handleSignal here. Could you add a comment to explain?

ftence (changes requested):

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
```diff
@@ -321,12 +449,23 @@ public void handle(String target, Request baseRequest, HttpServletRequest req, H
...
       //
       // Make sure stack is balanced
       //
 
       stack.checkBalanced();
 
+      if (maxtime > 0) {
+        stack.push(maxtime);
+        TIMEBOX.apply(stack);
+        // The call to TIMEBOX above will signal the stack, so we need to handle the signal received.
```
ftence, line 465: When the stack is signaled, an exception is thrown, so the statement below will never be able to handle the signal.

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
```diff
@@ -321,12 +449,23 @@ public void handle(String target, Request baseRequest, HttpServletRequest req, H
...
+
       //
       // Make sure stack is balanced
       //
 
       stack.checkBalanced();
 
+      if (maxtime > 0) {
+        stack.push(maxtime);
+        TIMEBOX.apply(stack);
```
ftence, line 464: This will modify the error message. For instance a simple `FAIL` will result in:
```
TIMEBOX encountered an exception while executing macro (Exception at 'FAIL' in section '[TOP]' (FAIL failed.))
```

ftence (changes requested):
`UPDATE` and `DELETE` are not interrupted which, I think, is a good thing.

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
hbs: The exception is thrown in handleSignal, the call to signal simply sets flags.

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
hbs: Can you be more elaborate?

etc/conf.templates/standalone/10-egress.conf.template
```diff
@@ -43,3 +43,22 @@ egress.clients.expose = false
...
+#egress.maxtime = 1000
+
+//
+// Maximum validity of runner nonces (in ms). Runner scripts are submitted to the /exec endpoint
+// with the header X-Warp10-Runner-Nonce set to a nonce containing the time of execution in ns since
+// the epoch. If the nonce is correct and less than this many ms in the past, the time boxing is
+// waived for the execution.
+// The default value if 1000 ms
+//
+#egress.runner.nonce.validity = 1000
```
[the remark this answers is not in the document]
hbs: The /exec endpoint could be the target endpoint of another 'standalone' ScriptRunner.

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
ftence: For instance, if `egress.timebox` is not set and no special header is set, a script with `FAIL` will return
`Exception at '=>FAIL<=' in section [TOP] (FAIL failed.)`

In the case `egress.timebox` is set and no special header is set, the same script will return 
`TIMEBOX encountered an exception while executing macro (Exception at 'FAIL' in section '[TOP]' (FAIL failed.))`

It may not be obvious to the user the execution was TIMEBOXed and the message can be confusing.

PR #1015, opened 2021-08-27
Added GUARD function to execute a macro with no risks of leaks

hbs:


randomboolean (approved):

warp10/src/main/java/io/warp10/script/functions/GUARD.java
```diff
@@ -0,0 +1,108 @@
...
+            if (symbol instanceof String) {
+              stack.forget((String) symbol);
+            } else if (symbol instanceof Long) {
+              stack.store(((Long) symbol).intValue(), null);
+            }
+          }
+        }
+      }
+
+      throw new WarpScriptStopException("Exception in GUARDed macro.");
```
ftence, line 99: Why use a `WarpScriptStopException` instead of a `WarpScriptException`?

Using a `WarpScriptStopException` will make the execution look like a success:
`<% FAIL %> NULL GUARD` will return a 200 status  and a `[]` JSON.

warp10/src/main/java/io/warp10/script/functions/GUARD.java
```diff
@@ -0,0 +1,108 @@
...
+        } else {
+          hidden = stack.hide(((Long) hide).intValue());
+        }
+      }
+      stack.exec(macro);
+    } catch (Throwable t) {
+      //
+      // If any exception was raised during the execution of the macro,
+      // clear the stack and the specified symbols so no information leak
+      // occurs
```
ftence, line 75: Shouldn't that be done even if the execution is a success?
`<% 'this is a secret' 'secret' STORE %> [ 'secret' ] GUARD` will leak `$secret`.

warp10/src/main/java/io/warp10/script/functions/GUARD.java
```diff
@@ -0,0 +1,108 @@
...
+      hasHide = true;
+      top = stack.pop();
+    }
+
+    List<Object> symbols = null;
+
+    if (top instanceof List) {
+      symbols = (List<Object>) top;
+      top = stack.pop();
+    }
```
ftence, line 52: We also could give a context to do the following:
```
<%
  SAVE '.context' STORE
  <%
    // The code
  %>
  $.context GUARD
%>
```

This would be almost equivalent to a RESTORE in a finally clause the difference being we are sure that the context is restored. In the case of a finally clause, this execution could be stopped if the stack is signaled.

ftence (changes requested):

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -1471,6 +1465,21 @@ public void restore() throws WarpScriptException {
...
+    // Retrieve the object on top of the stack
+    //
+
+    Object top = this.pop();
+
+    if (!(top instanceof StackContext)) {
+      throw new WarpScriptException("Invalid stack context.");
+    }
+
+    StackContext context = (StackContext) top;
```
ftence, line 1480: You forgot to call to `restore(context)`.

ftence (changes requested):

ftence (approved):
Tested and working.

Here is a test:
```
<%
  SAVE '.context' STORE
  
  <%
    'should_fail' STORE
    'this is secret' 'secret' STORE
    'this is also secret' POPR0
    'this is another secret'
    $should_fail <% FAIL %> IFT
    DROP
    'result of the macro'
  %>
  $.context -1 GUARD

%>
'macro' STORE

// Object left on the stack, should be left untouched.
'should stay visible'

// Macro is failing, no result should be on the stack
<% T @macro %>
<% %>
<% %>
TRY

// Check internal symbols and regs are undefined
PUSHR0 ISNULL ASSERT
'secret' DEFINED F == ASSERT

// Macro is not failing, the result should be on the stack
F @macro
'result of the macro' == ASSERT

// Check internal symbols and regs are still undefined
PUSHR0 ISNULL ASSERT
'secret' DEFINED F == ASSERT

// Because of the -1, GUARD leaves what's on the stack alone.
'should stay visible' == ASSERT

// Check the stack is empty
DEPTH 0 == ASSERT
```

PR #1018, opened 2021-09-03
Added header to specify a timeshift at ingress time

hbs:


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -2473,6 +2477,8 @@ public static GTSEncoder parse(GTSEncoder encoder, String str, Map<String,String
       throw new ParseException("Invalid timestamp.", tsoffset);
     }
 
+    timestamp += timeshift;
```
ftence, line 2480: There could be an overflow. Not sure if we should check and if we do, how does it impact performance.

ftence (approved):

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
hbs: This is an edge case not worth checking at ingestion time. We do not do it for relative timestamps either (=T+..)

Issue #1022, opened 2021-09-14
tcp plugin with multiple connections


pi-r-p:
With tcp and debug plugin enabled:

Here is the warpscript in my tcp directory:
```
{
  'mode' 'server'
  'host' '0.0.0.0'
  'port' 43211
  'timeout' 5000    // Optional, ms, defaults to 0 (no timeout)
                    //  when timeout is reached, the macro will be called anyway, 
                    //  with NULL on top of the stack if nothing was received in the meantime
  'parallelism' 4   // Optional, defaults to 1
  'qsize' 2048      // Optional, defaults to 1024
  'partitioner'     // Optional, the macro must return a LONG which will be a queue number.
  <% 
    // on the stack, you will find the packet content : 
    //  from bottom to top: IP (STRING), Source port (LONG), and the text line (STRING) 
    // you may partition up to content received.
    SNAPSHOT 'textLine' STORE
    'TCP partitionner input stack snapshot=' $textLine  + STDOUT
    RAND 4 * TOLONG //return a random number.  
  %> 
  'maxClients' 10    // Optional, defaults to 1
  'maxMessages' 1024 // Optional, defaults to 1024
  'macro' <%  // the macro executed on new lines received 
    // on top of the stack, there is a list of packet.
    'packetLIST' STORE
    <% $packetLIST ISNULL %>
    <%
      //'timeout reached, nothing received' STDOUT
    %>
    <%
      $packetLIST
      <%
        'packet' STORE
        //each packet is a list of three item: IP (STRING), Source port (LONG), and the text line (STRING)
        'TCP: line from IP:' $packet 0 GET + ' port:' + $packet 1 GET TOSTRING + 
        ' content=' + $packet 2 GET + 
        STDOUT
      %> FOREACH
    %> IFTE
  %>
}

```

Looking at Warp 10 logs, open a socket:
`nc -v localhost 43211`
type `a` + enter : output `a` is correct in the logs.

in a new terminal, open a new socket:
`nc -v localhost 43211`
type `b` + enter : there in no output in the logs. It should.

in a new terminal, open a new socket:
`nc -v localhost 43211`
type `c` + enter : there is no output in the logs, and the socket is closed by Warp 10. it should not.

Close the 2nd terminal: `b` appears in the logs.

ftence:
By default the maximum number of connection is limited to 1. Would you need to increase the number of parallel connections, you should specify `maxConnections` in your Map. I tested with 10 and it works as you expect.

[...]

PR #1023, opened 2021-09-14
Fix TCPClient stuck in LinkedBlockingQueue see #1022

ftence:
Currently, to limit the number of parallel connections the following is done:
- All incoming connections are accepted
- After accepting a connection, it is immediately closed if there are already `maxConnections` `TCPClients` running.

While this is not ideal, it appears that are no other option to implement this.

There currently an issue with accepted and open connections which are not handled by a running `TCPClient`, see #1022.

This is because the `ThreadPoolExecutor` is created with a pool of core=max=maxConnections threads and a  `LinkedBlockingQueue` of capacity `maxConnections`. So `maxConnections` `TCPClient`s could be accepted and running (thread pool capacity) and `maxConnections` `TCPClient`s could be accepted and waiting to be run (in the Queue).

The fix is to change the `LinkedBlockingQueue` to a `SynchronousQueue` which as no capacity.


pi-r-p (approved):
tested ok !

PR #1027, opened 2021-09-30
Implements Welford's algorithm in variance aggregator for better numerical stabilty

ftence:
With the current variance implementation some problem can arise, see:
```
[ "T07oF2SaOaOaOaNoF2SaOaOaOaJL0.." UNWRAP F mapper.var MAXLONG MAXLONG 1 ] MAP
0 GET 0 ATINDEX 4 GET
```
which gives a negative variance of:
```
[-1.1368683772161603E-13]
```

This is because the variance is orders of magnitude lower than the values, see [this article](johndcook.com/blog/2008/09/26/comparing-three-methods-of-computing-standard-deviation/).

The solution is to implement Welford's algorithm which gives the positive variance:
```
[6.3108872417680944E-30]
```

This implementation is slower by 10% but is much more stable.

This passes the tests and give very similar result to the sum of square method when used on non-extreme cases.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/aggregator/Variance.java
+++ b/warp10/src/main/java/io/warp10/script/aggregator/Variance.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2020  SenX S.A.S.
+//   Copyright 2018-2021  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


randomboolean:
If it's true that the implementation is 10% slower at scale, then it's not worth it to replace the current implementation.
Since the problem arises only in rare typical extreme cases, it should be better the add a "stable" flag to the function, with a mention in the documentation that explains when using the "stable" option is relevant.

PR #1044, opened 2021-11-09
Make runners scan subdirectories

ftence:
Currently, runners scan all files like `${runner.root}/*/periodicity/*.mc2`

This PR makes runners also scan files like `${runner.root}/*/periodicity/**/*.mc2`

This allows more flexibility to organize files.


hbs:
How does variable access work with scripts from subdirectories? Have you tested those cases?

ftence:
It works as expected.
If you have files like these:
```
runners
- test
  - 1000
    - foo.mc2
    - dir
      - bar.mc2
```

You can use:
```
foobar@/test='this will be replaced in foo.mc2 and bar.mc2'
bar@/test/dir='this will be replaced only in bar.mc2'
```

PR #1045, opened 2021-11-12
Gradle perf

steven-gueguen:
Improve build time by enabling parallel execution and caching for Gradle.
Remove gradle.properties from .gitignore and revamp this file.


hbs:
The gradle.properties file was voluntarily excluded via .gitignore to mitigate the risk of committing sensitive credentials by mistake (like ossrhPassword or nexusPassword).

Is there another place where those configuration could be specified? Like in the build.gradle files themselves?

steven-gueguen:
This configuration has to be in set in a gradle.properties file or by argument for the CLI.

Each person can use ${HOME}/.gradle/gradle.properties file to enable this configuration.
Though, I suggest using the personal gradle.properties file for credentials and project gradle.properties file for performance setting.

ftence (approved):
I definitively see a speed up with this PR which comes mainly from the caching mechanism.

As for the gradle.properties, I also use the method described by @steven-gueguen by using a global one in my home folder for secrets.

PR #1051, opened 2021-12-07
rexec may not flush

pi-r-p:
Very hard to reproduce. 
Sometimes, SNAPSHOT and OPB64 conversion are not done server side...

### DIFF
--- a/warp10/src/main/java/io/warp10/script/ext/rexec/REXEC.java
+++ b/warp10/src/main/java/io/warp10/script/ext/rexec/REXEC.java
@@ -134,11 +134,8 @@ public class REXEC extends NamedWarpScriptFunction implements WarpScriptStackFun
       out.write(WarpScriptLib.TOOPB64.getBytes(StandardCharsets.UTF_8));
       out.write('\n');
       
-      if (this.compress) {
-        out.close();
-      }
-      
-      connout.flush();
+      out.flush();
+      out.close();
 
       if (HttpURLConnection.HTTP_OK != conn.getResponseCode()) {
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/ext/rexec/REXEC.java
```diff
@@ -134,11 +134,8 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       out.write(WarpScriptLib.TOOPB64.getBytes(StandardCharsets.UTF_8));
       out.write('\n');
       
-      if (this.compress) {
-        out.close();
-      }
-      
-      connout.flush();
+      out.flush();
```
hbs, line 137: The call to flush is not needed since the stream is closed right after.

warp10/src/main/java/io/warp10/script/ext/rexec/REXEC.java
```diff
@@ -134,11 +134,8 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       out.write(WarpScriptLib.TOOPB64.getBytes(StandardCharsets.UTF_8));
```
hbs, line 134: Change copyright notice to reflect we are in 2021.

hbs (changes requested):

warp10/src/main/java/io/warp10/script/ext/rexec/REXEC.java
pi-r-p: I hope abstract close() comments can be trusted, I didn't check all the close implementation...

PR #1052, opened 2021-12-08
Added support for gskip/gcount in /delete endpoint

hbs:


pi-r-p:
No gcount gskip into FETCH or FIND documentation... must add it too.

steven-gueguen (approved):

PR #1057, opened 2022-01-01
Add order selection to GOLDWRAP

hbs:
In some contexts wrappers are always stored in reverse chronological order, it is therefore useful to have an option to build "gold" wrappers with the choice of order.

The behavior of GOLDWRAP is unchanged, it is the equivalent of 'false GOLDWRAP'. Gold wrappers in reverse chronological order can be created using 'true GOLDWRAP'.


steven-gueguen:
It does not seem to work on GTS list.

hbs:
PTAL

steven-gueguen (approved):

PR #1066, opened 2022-01-25
Egress.fetch.cache

hbs:
Disabled the use of the GTS file cache in cases when there is no network connection which could timeout while data is fetched.


pi-r-p (approved):
Just tested a few api/v0/fetch requests on a standalone. 
Assume Warp.isStandaloneMode() is reliable, because used in a few other functions.

PR #1067, opened 2022-01-25
Togts.opt

hbs:
Optimized ->GTS by relying on an array rather than a Map

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/TOGTS.java
+++ b/warp10/src/main/java/io/warp10/script/functions/TOGTS.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2020  SenX S.A.S.
+//   Copyright 2020-2022  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -31,6 +31,7 @@ import io.warp10.script.WarpScriptStackFunction;
 import org.apache.thrift.TDeserializer;
 import org.apache.thrift.TException;
 import org.apache.thrift.protocol.TCompactProtocol;
+import org.bouncycastle.util.Arrays;
 
 import java.math.BigDecimal;
 import java.nio.charset.StandardCharsets;
@@ -49,6 +50,8 @@ public class TOGTS extends NamedWarpScriptFunction implements WarpScriptStackFun
 
   private static final String TYPE_LABEL_NAME_PARAMETER = "label.type";
   private static final String DEFAULT_TYPE_LABEL_NAME = ".type";
+
+  private static final String[] TYPES = { "LONG", "DOUBLE", "BOOLEAN", "STRING", "BINARY" };
 
   public TOGTS(String name) {
     super(name);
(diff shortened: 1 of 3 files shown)


warp10/src/main/java/io/warp10/script/functions/TOGTS.java
```diff
@@ -127,47 +131,68 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-        Map<String, GeoTimeSerie> series = new HashMap<String, GeoTimeSerie>();
+        Arrays.fill(series, null);
+        long remaining = decoder.getCount();
         GeoTimeSerie gts;
         while (decoder.next()) {
           Object value = decoder.getBinaryValue();
+          remaining--;
 
-          String type = "DOUBLE";
+          int typeidx = 0; // LONG
```
pi-r-p, line 144: don't need to be initialized.

warp10/src/main/java/io/warp10/script/functions/TOGTS.java
```diff
@@ -211,24 +236,22 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+            enforcedType = "STRING";
+          } else if (value instanceof Boolean) {
+            enforcedType = "BOOLEAN";
+          } else if (value instanceof Long) {
+            enforcedType = "LONG";
+          } else if (value instanceof Double || value instanceof BigDecimal) {
+            enforcedType = "DOUBLE";
+          } else if (value instanceof byte[]) {
+            enforcedType = "BINARY";
           }
```
pi-r-p, line 251: must keep the `if (mustGuessTypeFromFirstValue) {` (it is here for the extra label to be correct)

```
          if(mustGuessTypeFromFirstValue) {
            if (value instanceof String) {
              enforcedType = "STRING";
            } else if (value instanceof Boolean) {
              enforcedType = "BOOLEAN";
            } else if (value instanceof Long) {
              enforcedType = "LONG";
            } else if (value instanceof Double || value instanceof BigDecimal) {
              enforcedType = "DOUBLE";
            } else if (value instanceof byte[]) {
              enforcedType = "BINARY";
            }
          }
```

pi-r-p (changes requested):
Good news is that doc unit tests also catch the extra label bug.

warp10/src/main/java/io/warp10/script/functions/TOGTS.java
hbs: Always better to initialize explicitely, won't fix.

pi-r-p: OK, can close.

warp10/src/main/java/io/warp10/script/functions/TOGTS.java
pi-r-p: better explaination : 
the extra label needs to be correct when the filters did not enforce a type before. But if filter does enforce the type, the first value from encoder must not change the label.

pi-r-p (approved):

PR #1069, opened 2022-01-26
Subserie.opt

hbs:
Optimize subserie extraction when overwriting is not requested by relying on System.arraycopy to extract the GTS slice.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2021  SenX S.A.S.
+//   Copyright 2018-2022  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -1899,12 +1899,45 @@ public class GTSHelper {
 
     // We know how many data will the new GTS so we provision arrays to receive the data.
     int count = lastidx - firstidx + 1;
+    // Allocate locations/elevations arrays so they get expanded by the call to multiProvision
+    if (null != gts.locations) {
+      subgts.locations = new long[0];
+    }
+    if (null != gts.elevations) {
+      subgts.elevations = new long[0];
+    }
     GTSHelper.multiProvision(subgts, gts.type, count, count);
 
-    for (int i = firstidx; i <= lastidx; i++) {
-      setValue(subgts, gts.ticks[i], null != gts.locations ? gts.locations[i] : GeoTimeSerie.NO_LOCATION, null != gts.elevations ? gts.elevations[i] : GeoTimeSerie.NO_ELEVATION, valueAtIndex(gts, i), overwrite);
+    if (!overwrite) {
+      if (null != subgts.locations) {
+        System.arraycopy(gts.locations, firstidx, subgts.locations, 0, count);
+      }
+      if (null != subgts.locations) {
+        System.arraycopy(gts.locations, firstidx, subgts.locations, 0, count);
+      }
+      if (null != subgts.ticks) {
+        System.arraycopy(gts.ticks, firstidx, subgts.ticks, 0, count);
+      }
+      if (null != subgts.longValues) {
+        System.arraycopy(gts.longValues, firstidx, subgts.longValues, 0, count);
+      }
+      if (null != subgts.doubleValues) {
+        System.arraycopy(gts.doubleValues, firstidx, subgts.doubleValues, 0, count);
+      }
+      if (null != subgts.stringValues) {
+        System.arraycopy(gts.stringValues, firstidx, subgts.stringValues, 0, count);
+      }
+      if (null != gts.booleanValues) {
+        subgts.booleanValues = gts.booleanValues.get(firstidx, lastidx + 1);
+      }
+      subgts.values = count;
+      subgts.sorted = gts.sorted;
+      subgts.reversed = gts.reversed;
+    } else {
+      for (int i = firstidx; i <= lastidx; i++) {
+        setValue(subgts, gts.ticks[i], null != gts.locations ? gts.locations[i] : GeoTimeSerie.NO_LOCATION, null != gts.elevations ? gts.elevations[i] : GeoTimeSerie.NO_ELEVATION, valueAtIndex(gts, i), overwrite);
+      }
     }
-
     return subgts;
   }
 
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -1899,12 +1899,45 @@ public static final GeoTimeSerie subSerie(GeoTimeSerie gts, long starttimestamp,
 
     // We know how many data will the new GTS so we provision arrays to receive the data.
```
randomboolean, line 1900: Add missing word or rephrase comment

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -1899,12 +1899,45 @@ public static final GeoTimeSerie subSerie(GeoTimeSerie gts, long starttimestamp,
...
+    }
     GTSHelper.multiProvision(subgts, gts.type, count, count);
 
-    for (int i = firstidx; i <= lastidx; i++) {
-      setValue(subgts, gts.ticks[i], null != gts.locations ? gts.locations[i] : GeoTimeSerie.NO_LOCATION, null != gts.elevations ? gts.elevations[i] : GeoTimeSerie.NO_ELEVATION, valueAtIndex(gts, i), overwrite);
+    if (!overwrite) {
+      if (null != subgts.locations) {
+        System.arraycopy(gts.locations, firstidx, subgts.locations, 0, count);
+      }
+      if (null != subgts.locations) {
```
randomboolean, line 1915: locations -> elevations

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -1899,12 +1899,45 @@ public static final GeoTimeSerie subSerie(GeoTimeSerie gts, long starttimestamp,
...
+        System.arraycopy(gts.locations, firstidx, subgts.locations, 0, count);
```
randomboolean, line 1916: ditto

randomboolean (changes requested):

hbs:
PTAL

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -1897,14 +1897,49 @@ public static final GeoTimeSerie subSerie(GeoTimeSerie gts, long starttimestamp,
...
+      }
+      if (null != subgts.doubleValues) {
+        System.arraycopy(gts.doubleValues, firstidx, subgts.doubleValues, 0, count);
+      }
+      if (null != subgts.stringValues) {
+        System.arraycopy(gts.stringValues, firstidx, subgts.stringValues, 0, count);
+      }
+      if (null != gts.booleanValues) {
+        subgts.booleanValues = gts.booleanValues.get(firstidx, lastidx + 1);
+      }
```
randomboolean, line 1934: The type of subgts is still UNDEFINED. It needs either to be set with each array copy, or by multiProvision if it does not affect it other usages.

hbs:
PTYAL

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -1897,14 +1897,64 @@ public static final GeoTimeSerie subSerie(GeoTimeSerie gts, long starttimestamp,
...
+    }
+    if (null != gts.elevations) {
+      subgts.elevations = new long[0];
     }
 
+    GTSHelper.multiProvision(subgts, gts.type, count, count);
+
+    if (!overwrite) {
+      if (null != subgts.locations) {
+        System.arraycopy(gts.locations, firstidx, subgts.locations, 0, count);
```
randomboolean, line 1915: In fact, locations and elevations have not been expanded by multiProvision, so this gives an ArrayIndexOutOfBoundsException

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -1897,14 +1897,64 @@ public static final GeoTimeSerie subSerie(GeoTimeSerie gts, long starttimestamp,
...
+      }
+      if (null != subgts.elevations) {
+        System.arraycopy(gts.elevations, firstidx, subgts.elevations, 0, count);
+      }
+      if (null != subgts.ticks) {
```
randomboolean, line 1920: Checking if (count > 0) before multiProvision could save the null check for all non-geo arrays

randomboolean (changes requested):

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -1897,14 +1897,70 @@ public static final GeoTimeSerie subSerie(GeoTimeSerie gts, long starttimestamp,
...
+        } else {
+          subgts.elevations = null;
+        }
+        if (null != subgts.ticks) {
+          System.arraycopy(gts.ticks, firstidx, subgts.ticks, 0, count);
+        }
 
+        switch(gts.type) {
+          case LONG:
+            if (null != subgts.longValues) {
```
randomboolean, line 1931: We don't really need this check since multiProvision has created the array since count > 0.
However, we can do differently by doing the same thing than for geo arrays (and not use multiProvision in the case !overwrite). That would save writing twice in the same array when we expand and copy.

randomboolean (changes requested):

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -1897,14 +1897,65 @@ public static final GeoTimeSerie subSerie(GeoTimeSerie gts, long starttimestamp,
     // Extract values/locations/elevations that lie in the requested interval
     //
 
-    // We know how many data will the new GTS so we provision arrays to receive the data.
+    // We know how many data the new GTS will have so we provision arrays to receive the data.
     int count = lastidx - firstidx + 1;
+
     GTSHelper.multiProvision(subgts, gts.type, count, count);
```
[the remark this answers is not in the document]
hbs: In multiProvision if subgts.type is UNDEFINED then the value array will be allocated according to gts.type so allocation will be correctly done. Did you see another execution path which proves otherwise?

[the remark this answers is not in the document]
hbs: Ok got your execution path, the case when gts.ticks is not null but type is UNDEFINED.

PR #1070, opened 2022-01-27
Added support for macro placeholder in repo URLs

hbs:
Some URLs may require parameters which would need to appear after the macro name. This PR adds the possibility to specify a macro placeholder '{macro}' in the repo URLs. If present, the macro path will be inserted where the placeholder is thus allowing for URLs of the type:

http://host/path/to/root/{macro}?param=foo


pi-r-p (changes requested):
also need to update both 20-warpfleet.conf.template files with an example.

pi-r-p (approved):

PR #1073, opened 2022-02-08
Added support for multiline B64

hbs:


warp10/src/main/java/io/warp10/script/functions/B64TO.java
```diff
@@ -27,21 +27,21 @@
...
-    
+
     if (!(o instanceof String)) {
       throw new WarpScriptException(getName() + " operates on a String.");
     }
-    
-    stack.push(BaseEncoding.base64().decode(o.toString()));
-    
+
+    stack.push(BaseEncoding.base64().decode(o.toString().replace("\n","").replace(" ", "")));
```
pi-r-p, line 43: also remove \r , because of windows vscode.

pi-r-p (changes requested):

pi-r-p (approved):

PR #1074, opened 2022-02-09
allow double for limits related functions

pi-r-p:
```
10e6 MAXOPS
```
more easy to read scientific notation for big numbers. fixed for all limit related function.


hbs:
What is the point? Those limits are known to be LONGs, why would you pass a DOUBLE to the function which modify them?

hbs:
But don't you think it will incorrectly lead people to think that +Infinity MAXOPS would remove the limtis? What happens when doing NaN MAXOPS?

PR #1075, opened 2022-02-11
Added support for decoding directly from String instances

hbs:


pi-r-p (approved):
Exactly the same performances measured between:
- OrderPreservingBase64.decode(o.toString().getBytes(StandardCharsets.US_ASCII))
- OrderPreservingBase64.decode(o.toString())

(~170MB/s)

The `B64->` function is 6 times slower.

hbs:
The motivation was not performance but memory consumption. Decoding directly from the String saves the allocation of the byte array.

PR #1081, opened 2022-03-10
Fixed validator so credentials are not changed to lowercase

hbs:


warp10/src/main/java/io/warp10/script/WarpFleetMacroRepository.java
```diff
@@ -522,13 +536,18 @@ private static String validateRepo(String repo) {
...
-      repo = "https://" + host + repo.substring(host.length() + 8);
+    if (repo.startsWith("http://") || repo.startsWith("https://")) {
+      try {
+        URL url = new URL(repo);
+        // Force the host to be LC
+        URI uri = new URI(url.getProtocol(), url.getUserInfo(), url.getHost().toLowerCase(), url.getPort(), url.getPath(), url.getQuery(), url.getRef());
+        String canonical = uri.toString().replaceAll(WarpFleetMacroRepository.MACRO_PLACEHOLDER_ENCODED, WarpFleetMacroRepository.MACRO_PLACEHOLDER);
+        repo = canonical;
+      } catch (MalformedURLException|URISyntaxException e) {
+        return null;
```
pi-r-p, line 547: Why not raising a WarpScript exception, or a warning here ? For example, ipv6 host with capital letters are not supported by the underlying java lib, but it is not obvious:
```
"http://User29:5gq%25@fe80::f38:c469:B00B:807c/afad/{macro}?i=osef"  WF.ADDREPO
```

Same for guys who forget the square brackets around ipv6 before port 9994:
```
"http://User29:5gq%25@[fe80::f38:c469:d36d:807c]:9994/afad/{macro}?i=osef"  WF.ADDREPO
```
(tested, it works)

hbs: We return null because we do not want a single bad URL to clear the list, so incorrect URLs are simply silently ignored. We could add a log message though to make diagnostic easier.

pi-r-p (approved):

PR #1086, opened 2022-04-01
WarpRun improvements

hbs:


warp10/src/main/sh/warp10-standalone.sh
```diff
@@ -569,6 +569,11 @@ tokengen() {
   ${JAVACMD} -cp ${WARP10_JAR} -Dlog4j.configuration=file:${LOG4J_CONF} -Dfile.encoding=UTF-8 io.warp10.worf.TokenGen ${CONFIG_FILES} $2 -
 }
 
+run() {
+  getConfigFiles
+  ${JAVACMD} -cp ${WARP10_JAR} -Dlog4j.configuration=file:${LOG4J_CONF} -Dfile.encoding=UTF-8 -Dwarp10.config="${CONFIG_FILES}" io.warp10.WarpRun $1
```
steven-gueguen, line 574: Should be $2 -

steven-gueguen (commented):
Doc is missing for new function

steven-gueguen (approved):

hbs:
PTAL, the calling parameters were incorrect for WarpRun, the extraneous '-' needed to be removed.

warp10/src/main/java/io/warp10/WarpRun.java
```diff
@@ -25,61 +25,97 @@
...
+      // Register the WarpRun and Debug extensions
+      //
+
+      WarpScriptLib.register(new WarpRunWarpScriptExtension());
+      WarpScriptLib.register(new DebugWarpScriptExtension());
+
       MemoryWarpScriptStack stack = new MemoryWarpScriptStack(null, null, properties);
       stack.maxLimits();
-      
+
```
steven-gueguen, line 80: Break condition should be change at line 94

steven-gueguen (changes requested):

warp10/src/main/java/io/warp10/WarpRun.java
hbs: ?

steven-gueguen: When dealing with a "-" parameter, the reading process never end.
The `if (null == line) {` should be adapted.

hbs: When stdin reaches EOF, readline will return null, do you have a test case where that is not the case?

steven-gueguen: No, I misused it

steven-gueguen (approved):

PR #1091, opened 2022-05-09
Added buffered version of encodeTo{Stream,Writer}. Changed EgressFetchHandler to use it with 100000 bytes buffer.

hbs:


crypto/src/main/java/io/warp10/crypto/OrderPreservingBase64.java
```diff
@@ -156,6 +156,67 @@ public static void encodeToStream(OutputStream out, byte[] data, int offset, int
     }
   }
 
+  /**
+   * Encode to writer.
```
randomboolean, line 160: Not a writer here

randomboolean:
Bufferized encodeToStream can also be set for io.warp10.continuum.LogDumper and io.warp10.continuum.Pack.OutputChunk

crypto/src/main/java/io/warp10/crypto/OrderPreservingBase64.java
```diff
@@ -156,6 +156,75 @@ public static void encodeToStream(OutputStream out, byte[] data, int offset, int
...
+
+    int dataidx = 0;
+
+    int bufidx = 0;
+
+    int i = 0;
+
+    for (i = offset; i < offset + datalen; i++) {
+      // Ensure we have at least 2 slots free for case 2
+      if (buf.length  - bufidx < 2) {
```
randomboolean, line 193: why not use all slot at each batch ? use < 2 in case 2, and < 1 in case 0 and 1

crypto/src/main/java/io/warp10/crypto/OrderPreservingBase64.java
```diff
@@ -212,6 +281,74 @@ public static void encodeToWriter(Writer out, byte[] data, int offset, int datal
...
+
+    int len = 4 * (datalen / 3) + (datalen % 3 != 0 ? 1 + (datalen % 3) : 0);
+
+    int dataidx = 0;
+
+    int bufidx = 0;
+
+    for (i = offset; i < offset + datalen; i++) {
+      // Ensure we have at least 2 slots free for case 2
+      if (buf.length  - bufidx < 2) {
```
randomboolean, line 317: ditto

randomboolean (approved):

PR #1094, opened 2022-06-08
Added test for encodeTo{Stream,Writer}

hbs:


pi-r-p (changes requested):
It does not test the most complex ones (encodeToWriter(Writer out, byte[] data, int offset, int datalen, int buflen), and encodeToStream(OutputStream out, byte[] data, int offset, int datalen, int buflen)

hbs:
yes it does:

```
  public static void encodeToWriter(byte[] data, Writer out) throws IOException {
    encodeToWriter(out, data, 0, data.length);
  }
  public static void encodeToStream(byte[] data, OutputStream out) throws IOException {
    encodeToStream(out, data, 0, data.length);
  }
```

hbs:
indeed, I thought you meant the ones with offset + len. Tests will be added to the open PR.

PR #1095, opened 2022-06-09
Added tests for buffered encodeTo{Stream.Writer}

hbs:


pi-r-p (changes requested):
It could be wiser to test with an input greater than buffer size, to be sure there is no problem.

pi-r-p (approved):

PR #1097, opened 2022-06-17
Add METAMATCH to match metadata

hbs:


pi-r-p (changes requested):
It would be better to extend ElementOrListStackFunction

pi-r-p (approved):

Issue #1096, opened 2022-06-17
Mixed comments are incorrectly parsed


hbs:
Mixing C style comments and shell style comments lead to incorrect parsing:

/* ## Fail */

hbs:
Fixed by PR #1098

PR #1098, opened 2022-06-17
fix inline comments within block comments

pi-r-p:
Fix issue #1096

### DIFF
--- a/warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
+++ b/warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2020-2021  SenX S.A.S.
+//   Copyright 2020-2022  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -609,7 +609,7 @@ public void exec(String line) throws WarpScriptException {
           // End execution on encountering a comment
           //
 
-          if (!inMultiline.get() && stmt.length() > 0 && (stmt.charAt(0) == '#' || (stmt.charAt(0) == '/' && stmt.length() >= 2 && stmt.charAt(1) == '/'))) {
+          if (!inMultiline.get() && !inComment.get() && stmt.length() > 0 && (stmt.charAt(0) == '#' || (stmt.charAt(0) == '/' && stmt.length() >= 2 && stmt.charAt(1) == '/'))) {
```
hbs, line 612: intuitively I would have thought that inComment.get() was the costliest of all those conditions, if that is the case then it would be better to move it at the end of the conditional to degrade performance less.

The same could be thought of inMultiline.get()

hbs (changes requested):

PR #1108, opened 2022-07-13
feat(directory): add a series clean while reading HBase

### LINKED ISSUE #1107: Directory cleaner
# Directory cleaner

Hello Warp10 community, we are a big user of the Warp10 platform, and might be the first one to reach a billion series. However, this is way too much (at least for our use case). This is not something due to the technology in itself, but mainly how we use it. 

## Ephemeral series

When we write our series data model, we choose to create ephemeral series. At first, it seems to be the best solution as it would avoid having our metrics overlapping.  However soon, the drawback of this choice was that we generated way too much series, which quickly became obsolete. This means that our directories took more and more ram, they also were slower to load their metadata from HBase. But what is now worse is: that a lot of the series have the same class slowing all directories requests. This means that we cannot easily delete inactive series by executing a WarpScript.

## Solution

We needed an efficient way to delete our series without generating too much traffic on our platform. We decided to boot a new server which will host a custom directory a little bit modified. We thought about something similar to the HDFS balancer that we can control with some `systemd timers` or `cron`. 

This new directory will read HBase, and based on a given parameter clean the inactive series. We compare the current date and the last activity of each series we read from HBase. If the inactivity window is above the parameter set in the `directory cleaner` configuration, we send a delete message in Kafka. This means that he has its own producer, and that all other directories up will be notified of the deletion. The `directory cleaner` can then be booted on a new server and is fully independent. In our case, we wanted that when it finishes its jobs to stop. This enables us to boot this new `directory cleaner` with a lot less RAM than our directories. 

## Customisable

In our code, we add a lot of new parameters to set this new feature as customisable as generic as possible. 

We added a boolean parameter whether the directory should stop at the end of the process. 
We added a global time window (ex: 2 years) to delete series that are inactive (ex: series inactive for 2 years).
We added a pair that can contain a map of applications and a time window. This parameter overwrites the previous check to enable series deletion series with a custom-time window for selected applications.
We added the same parameters as an ingress node to customise the directory Kafka producer.

## In conclusion

We have just started to use it on our cluster and started to clear our directories. Now that we start to use it, we want to share it with the community as it can help other people that have the same experience with ephemeral series (or only some Warp10 applications that would have ephemeral series).

We will appreciate any feedback, and/or how could we improve such a component?

aurrelhebert:
closes #1107

### DIFF
--- a/etc/conf.templates/distributed/10-directory.conf.template
+++ b/etc/conf.templates/distributed/10-directory.conf.template
@@ -97,6 +97,24 @@ directory.kafka.metadata.groupid = directory.metadata-localhost
 directory.kafka.metadata.commitperiod = 1000
 
 //
+// Maximum message size for the 'metadata' topic 
+// Used by directory when used as a producer to clean Warp10 metadata
+//
+# directory.kafka.metadata.maxsize
+
+//
+// Kafka broker list for the 'meta' topic 
+// Used by directory when used as a producer to clean Warp10 metadata
+//
+# directory.kafka.metadata.brokerlist
+
+//
+// Size of Kafka Producer pool for the 'metadata' topic
+// Used by directory when used as a producer to clean Warp10 metadata
+//
+# directory.kafka.metadata.poolsize
+
+//
 // Maximum byte size we allow the pending Puts list to grow to
 //
 directory.hbase.metadata.pendingputs.size = 1000000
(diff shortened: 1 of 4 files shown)


steven-gueguen (approved):

pi-r-p:
Hello,
The 2.11.0 is now released with [TrustedDirectoryPlugin](https://github.com/senx/warp10-platform/blob/master/warp10/src/main/java/io/warp10/warp/sdk/TrustedDirectoryPlugin.java).

As your feature is very specific, it won't be merged into the core, but it could now be a plugin. 
Please contact us if TrustedDirectoryPlugin do not fit your needs, or if you want us to help or add features in the core to build your plugin.

hbs:
Given HBase is due for removal in 3.x and this won't be fixed in 2.x, it is probably better you close this PR.

aurrelhebert:
How will you handle series lifecycle in 3.x?

pi-r-p:
Last activity remains, and we have to make sure that people with lots of "temporary" gts use it...

aurrelhebert:
As you prefer, we do use this PR on production and it's working well since July.

PR #1109, opened 2022-07-20
Opb64 12bits

pi-r-p:
Process bytes 3 by 3 instead of one by one to create OPB64 4 by 4 characters.

Twice faster on tested platforms.


crypto/src/main/java/io/warp10/crypto/OrderPreservingBase64.java
```diff
@@ -43,8 +44,20 @@ public class OrderPreservingBase64 {
     for (int i = 0; i < ALPHABET.length; i++) {
       TEBAHPLA[ALPHABET[i]] = (byte) i;
     }
-  }
 
+    // also build a 12bit look up table. Encoding per 12bits double speed. (now reaching 30MB/s)
+    // output is the list of two bytes representing the 12 bits input:
```
randomboolean, line 49: This does not help understanding why this is useful.
Instead, please explain why this lookup table is useful.
For example recall that the encoding will wrap 3 bytes into 4 characters, each representing 6 bits.
So that when we process these 3 bytes at once, we can split them into two separates sets of 12 bits, mapping to the same couple of characters as if we were looking up twice 6 bits.

pi-r-p:
Detailed comments added.

randomboolean (approved):

PR #1117, opened 2022-09-02
uniform support of empty buckets among available mappers

pi-r-p:
Some mappers did support empty buckets, some failed with `can only be applied to a single value` error message. This was confusing, as the error message leads to DEDUP, not to UNBUCKETIZE.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/mapper/MapperAbs.java
+++ b/warp10/src/main/java/io/warp10/script/mapper/MapperAbs.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2020  SenX S.A.S.
+//   Copyright 2018-2022  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -39,6 +39,10 @@ public class MapperAbs extends NamedWarpScriptFunction implements WarpScriptMapp
     long[] elevations = (long[]) args[5];
     Object[] values = (Object[]) args[6];
 
+    if (0 == values.length) {
+      return new Object[] {0L, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, null};
+    }
+    
     if (values.length > 1) {
       throw new WarpScriptException(getName() + " can only be applied to a single value.");
     }
--- a/warp10/src/main/java/io/warp10/script/mapper/MapperCeil.java
+++ b/warp10/src/main/java/io/warp10/script/mapper/MapperCeil.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2020  SenX S.A.S.
+//   Copyright 2018-2022  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package io.warp10.script.mapper;
 
+import io.warp10.continuum.gts.GeoTimeSerie;
 import io.warp10.script.NamedWarpScriptFunction;
 import io.warp10.script.WarpScriptMapperFunction;
 import io.warp10.script.WarpScriptException;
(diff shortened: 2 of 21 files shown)


warp10/src/main/java/io/warp10/script/mapper/MapperAbs.java
```diff
@@ -39,6 +39,10 @@ public Object apply(Object[] args) throws WarpScriptException {
     long[] elevations = (long[]) args[5];
     Object[] values = (Object[]) args[6];
 
+    if (0 == values.length) {
+      return new Object[] {0L, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, null};
```
randomboolean, line 43: This case was already handled correctly by this mapper. Also this change may make the MAP delete a value at epoch 0.

warp10/src/main/java/io/warp10/script/mapper/MapperCeil.java
```diff
@@ -38,6 +39,10 @@ public Object apply(Object[] args) throws WarpScriptException {
     long[] elevations = (long[]) args[5];
     Object[] values = (Object[]) args[6];
 
+    if (0 == values.length) {
+      return new Object[] {0L, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, null};
```
randomboolean, line 43: Change 0L to tick. Or else value at epoch 0 could get deleted unintentionally if overwrite flag is true.

warp10/src/main/java/io/warp10/script/mapper/MapperDateTime.java
```diff
@@ -44,6 +45,15 @@ public Object apply(Object[] args) throws WarpScriptException {
     long tick = (long) args[0];
     long[] locations = (long[]) args[4];
     long[] elevations = (long[]) args[5];
+    Object[] values = (Object[]) args[6];
+
+    if (0 == values.length) {
+      return new Object[] {0L, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, null};
```
randomboolean, line 51: ditto

warp10/src/main/java/io/warp10/script/mapper/MapperFinite.java
```diff
@@ -39,6 +40,10 @@ public Object apply(Object[] args) throws WarpScriptException {
     long[] elevations = (long[]) args[5];
     Object[] values = (Object[]) args[6];
 
+    if (0 == values.length) {
+      return new Object[] {0L, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, null};
```
randomboolean, line 44: ditto

randomboolean: this comment is applicable to all other mappers

warp10/src/main/java/io/warp10/script/mapper/MapperTick.java
```diff
@@ -51,6 +52,10 @@ public Object apply(Object[] args) throws WarpScriptException {
     long[] elevations = (long[]) args[5];
     Object[] values = (Object[]) args[6];
 
+    if (0 == values.length) {
+      return new Object[] {0L, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, null};
```
randomboolean, line 56: The bucket is empty regarding its value, but it does have a tick. Are we certain we don't want to return that ?

randomboolean (changes requested):

warp10/src/main/java/io/warp10/script/mapper/MapperAbs.java
pi-r-p: It is consistent with all other existing mapper implementation. Look at MapperAdd.
(I mark as resolved the related conversations)

image

warp10/src/main/java/io/warp10/script/mapper/MapperTick.java
pi-r-p: Yes. A use case of mapper.tick is to compute data loss  (mapper.tick followed by mapper.delta).

pi-r-p:
PTAL

randomboolean (approved):
Not clean because we consider implicitly that if the value returned is null then the tick is not to be interpreted. But functionally that works wherever these methods are used.

PR #1119, opened 2022-09-09
simplify single value MAP

randomboolean:
`[ $gts $mapper 0 0 0 ] MAP` is simplified into `[ $gts $mapper ] MAP`

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/MAP.java
+++ b/warp10/src/main/java/io/warp10/script/functions/MAP.java
@@ -69,8 +69,18 @@ public class MAP extends NamedWarpScriptFunction implements WarpScriptStackFunct
 
     List<Object> params = (List<Object>) top;
 
+    if ( 2 > params.size()) {
+      throw new WarpScriptException(getName() + " needs a list of at least 2 parameters as input.");
+    }
+
+    if (!(params.get(params.size() - 1) instanceof Number)) {
+      params.add(0L);
+      params.add(0L);
+      params.add(0L);
+    }
+
     if (5 > params.size()) {
-      throw new WarpScriptException(getName() + " needs a list of at least 5 parameters as input.");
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -69,8 +69,18 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
     List<Object> params = (List<Object>) top;
 
+    if ( 2 > params.size()) {
```
hbs, line 72: remove extraneous space before '2'

warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -69,8 +69,18 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      throw new WarpScriptException(getName() + " needs a list of at least 2 parameters as input.");
+    }
+
+    if (!(params.get(params.size() - 1) instanceof Number)) {
```
hbs, line 76: You are assuming that params.size() is 2, which might not be the case. How about adding 0s until params.size() is 5 if params.size() is initially less than 5?

hbs (changes requested):

warp10/src/main/java/io/warp10/script/functions/MAP.java
randomboolean: It would not work with the original signature of MAP that is still supported (there can be any number of GTS or lists before the mapper), unless we compare with (nseries + 3) instead of 5.

hbs: In that case check the last 1-3 parameters and complete with 0s for the missing ones.

hbs (approved):

PR #1120, opened 2022-09-09
Variables

randomboolean:
Add LSTORE and MSTORE to more easily manipulate variables.

1/
The usual pattern:
`some-list LIST-> DROP [ "varname1" "varname2" "varname3" ] STORE`
is simplified into:
`some-list [ "varname1" "varname2" "varname3" ] LSTORE`
which makes it a bit more readable and does not spill the list on the stack (bad practice for big lists).

2/
`some-map MSTORE` is a convenient way to store objects of a map into its keys. This makes some warpscripts more readable since multiple `$map "varnameN" GET` are simplified into `$varnameN` (which is also better for auto-completion).

3/ A commit had the side effect to sort the imports of WarpScriptLib.java and unfold import io.warp10.script.functions.* (due to my IDE automatic import settings)


warp10/src/main/java/io/warp10/script/functions/LSTORE.java
```diff
@@ -0,0 +1,75 @@
...
+      }
+    }
+
+    o = stack.pop();
+    if (!(o instanceof List)) {
+      throw new WarpScriptException(getName() + " expects a list as first argument.");
+    }
+
+    List objects = (List) o;
+    if (variables.size() != objects.size()) {
```
hbs, line 55: We should support those size discrepancies. If the variable list is shorter simply store those variables, if the first list is shorter, assume NULL as the value for the other elements.

warp10/src/main/java/io/warp10/script/functions/LSTORE.java
```diff
@@ -0,0 +1,75 @@
...
+      Object symbol = variables.get(i);
+
+      if (null == symbol) {
+        continue;
+      }
+
+      if (symbol instanceof Long) {
+        stack.store(((Long) symbol).intValue(), objects.get(i));
+      } else {
+        stack.store(symbol.toString(), objects.get(i));
```
hbs, line 69: Use a cast to String, since you've checked that you only have Longs and Strings.

warp10/src/main/java/io/warp10/script/functions/MSTORE.java
```diff
@@ -0,0 +1,65 @@
...
+      Object symbol = entry.getKey();
+
+      if (null == symbol) {
+        continue;
+      }
+
+      if (symbol instanceof Long) {
+        stack.store(((Long) symbol).intValue(), entry.getValue());
+      } else {
+        stack.store(symbol.toString(), entry.getValue());
```
hbs, line 59: Use cast to String

hbs (changes requested):

hbs:
You should add LCSTORE and MCSTORE variants too.

warp10/src/main/java/io/warp10/script/functions/LCSTORE.java
```diff
@@ -0,0 +1,80 @@
+//
```
hbs, line 1: Merce LCSTORE with LSTORE by adding a flag to its constructor, the code being almost identical.

warp10/src/main/java/io/warp10/script/functions/MCSTORE.java
```diff
@@ -0,0 +1,69 @@
+//
```
hbs, line 1: Merge MCSTORE with MSTORE with an added flag in constructor.

hbs (changes requested):

hbs (approved):

pi-r-p:
```
some-map MSTORE is a convenient way to store objects of a map into its keys. This makes some warpscripts more readable since multiple $map "varnameN" GET are simplified into $varnameN (which is also better for auto-completion).
```

Well, no, those created by MSTORE won't be available for auto completion at all.

(LSTORE can be added to VSCode parsing, no problem)

randomboolean:
Got it. Auto-completion would have just been a consequence of this feature anyway.

hbs:
Actually one point I overlooked in my review is the fact that ASREGS should be modified to fail if LSTORE/LCSTORE/MSTORE/MCSTORE are used.

pi-r-p:
`some-map MSTORE`  would lead to no autocompletion at all. But `some-map [ 'key1' 'key3' ] MSTORE` could be a good idea, to store only a few keys of the input, and to provide autocompletion.

PR #1122, opened 2022-09-14
Fixed cases when replacement is impossible

hbs:


randomboolean:
If a list of variable names is passed as second argument of STORE, ASREGS replaces them if the list was created explicitly or if it was imported from early binding.
Id I'm not mistaken, the same thing can be done for LSTORE and MSTORE.
This could also be implemented of *CSTORE conditional variants.

pi-r-p:
Is it possible to use a String instead of the abort flag, to get a nice error message ? For example:
- was unable to convert variables to registers. (encounter LSTORE or MSTORE)
- was unable to convert variables to registers. (encounter STORE with a non explicit variable list name)
- was unable to convert variables to registers. (encounter STORE with a non explicit variable name)
- was unable to convert variables to registers. (external macro call may modify variables)

pi-r-p (changes requested):

pi-r-p:
@randomboolean : btw, there is no doc for LSTORE

Opinion: VARS was designed to be used with ASREGS, but it is tricky to use as soon as there are global variables (true VARS was added later on). 
Then ASREGS was fixed in 2.6.0 to avoid the use of VARS.  No need to add more complexity to VARS...

randomboolean (approved):

PR #1123, opened 2022-09-15
add optionnal list of key to MSTORE args

pi-r-p:
add optionnal list of key to MSTORE args

Allows autocompletion, key selection, and missing keys in the input will be defined as variable (with null value).

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/MSTORE.java
+++ b/warp10/src/main/java/io/warp10/script/functions/MSTORE.java
@@ -21,6 +21,7 @@ import io.warp10.script.WarpScriptException;
 import io.warp10.script.WarpScriptStack;
 import io.warp10.script.WarpScriptStackFunction;
 
+import java.util.List;
 import java.util.Map;
 
 public class MSTORE extends NamedWarpScriptFunction implements WarpScriptStackFunction {
@@ -35,11 +36,16 @@ public class MSTORE extends NamedWarpScriptFunction implements WarpScriptStackFu
   public MSTORE(String name) {
     this(name, false);
   }
-  
+
   @Override
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
+    List optionnalKeyList = null;
     Object o = stack.pop();
 
+    if (o instanceof List) {
+      optionnalKeyList = (List) o;
+      o = stack.pop();
+    }
     if (!(o instanceof Map)) {
       throw new WarpScriptException(getName() + " expects a Map as argument.");
     }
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/functions/MSTORE.java
```diff
@@ -35,42 +36,66 @@ public MSTORE(String name, boolean conditional) {
   public MSTORE(String name) {
     this(name, false);
   }
-  
+
   @Override
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
+    List optionnalKeyList = null;
```
randomboolean, line 42: remove a 'n'

warp10/src/main/java/io/warp10/script/functions/MSTORE.java
```diff
@@ -35,42 +36,66 @@ public MSTORE(String name, boolean conditional) {
...
     Object o = stack.pop();
 
+    if (o instanceof List) {
+      optionnalKeyList = (List) o;
+      o = stack.pop();
+    }
```
randomboolean, line 48: add line break after }

warp10/src/main/java/io/warp10/script/functions/MSTORE.java
```diff
@@ -35,42 +36,66 @@ public MSTORE(String name, boolean conditional) {
...
+            stack.store(((Long) symbol).intValue(), v);
+          }
+        } else {
+          if (!conditional || !stack.getSymbolTable().containsKey((String) symbol)) {
+            stack.store((String) symbol, v);
+          }
+        }
+      }
+    } else {
+      for (Object elt: variables.keySet()) {
```
randomboolean, line 75: Simply check here that elt is in optionalKeyList or not, so as to reuse code for better readability.

randomboolean (changes requested):

warp10/src/main/java/io/warp10/script/functions/MSTORE.java
pi-r-p: If the input map has lots of keys, iteration on the optional list is more efficient.  (I don't say it is a good use of this function... but it may happen)

randomboolean: I don't think we would see any difference by optimizing looping through maps in comparison than loop done through GTS data for example.
Anyway, it is still possible to refactor the code using one iterator while loop rather than two for loops:
```
Iterator itr = null != optionalKeyList ? optionalKeyList.iterator() : variables.keySet().iterator();
    while(itr.hasNext()) {  ...  }
```

pi-r-p:
Refactored. Can you check with your unit tests of MSTORE MCSTORE ?

randomboolean (approved):

PR #1134, opened 2022-10-14
Removed legacy in-memory implementation

hbs:


etc/conf.templates/standalone/30-in-memory.conf.template
```diff
@@ -97,16 +97,6 @@ accelerator.ephemeral = false
 //
 in.memory = false
 
-//
```
steven-gueguen, line 100: Please update the copyright date

steven-gueguen (changes requested):

steven-gueguen (approved):

PR #1137, opened 2022-10-19
Manager.capability

hbs:


warp10/src/main/java/io/warp10/script/functions/MANAGERONOFF.java
```diff
@@ -19,35 +19,36 @@
...
   public MANAGERONOFF(String name, String attr, boolean on) {
     super(name);
     this.on = on;
     this.attr = attr;
   }
-  
+
   @Override
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
     Object top = stack.pop();
```
steven-gueguen, line 39: As the manager secret is not needed anymore, this line can be removed.
The documentation of the implemented WarpScript functions should be updated accordingly.

steven-gueguen (approved):

PR #1141, opened 2022-10-19
Replaced secret with a capability

hbs:


steven-gueguen (approved):

steven-gueguen (changes requested):
Need an update of the copyright date for the configuration files before approval

steven-gueguen (approved):

PR #1145, opened 2022-11-04
Added support for attributes

hbs:


warp10/src/main/java/io/warp10/hadoop/Warp10InputFormat.java
```diff
@@ -161,6 +161,7 @@ public class Warp10InputFormat extends InputFormat<Text, BytesWritable> {
   public static final String PROPERTY_WARP10_FETCH_TIMESTEP = "warp10.fetch.timestep";
   public static final String PROPERTY_WARP10_FETCH_PREBOUNDARY = "warp10.fetch.preboundary";
   public static final String PROPERTY_WARP10_FETCH_POSTBOUNDARY = "warp10.fetch.postboundary";
+  public static final String PROPERTY_WARP10_FETCH_ATTR_PREFIX = "warp10.fetch.attr.";
```
pi-r-p, line 164: Wouldn't it be better to compose the name with warp10.fetch. + HTTP_PARAM_ATTR_PREFIX ?

warp10/src/main/java/io/warp10/hadoop/Warp10InputFormat.java
```diff
@@ -191,6 +192,7 @@ public class Warp10InputFormat extends InputFormat<Text, BytesWritable> {
   public static final String HTTP_HEADER_POSTBOUNDARY = "X-Warp10-Postboundary";
   public static final String HTTP_HEADER_TIMESTEP = "X-Warp10-Timestep";
   public static final String HTTP_HEADER_SHOW_ERRORS = "X-Warp10-ShowErrors";
+  public static final String HTTP_HEADER_ATTR_PREFIX = "X-Warp10-Attr.";
```
pi-r-p, line 195: boundary.post  matching header is X-Warp10-Postboundary, and this is a legal http header.

But attr.this.that  matching header will be X-Warp10-Attr.attr.this.that

Lots of software will filter out the header because of the dot (nginx for example).
This may cause troubles.

pi-r-p (changes requested):

warp10/src/main/java/io/warp10/hadoop/Warp10RecordReader.java
```diff
@@ -171,6 +189,14 @@ public void initialize(InputSplit split, TaskAttemptContext context)
           conn.setRequestProperty(Warp10InputFormat.HTTP_HEADER_DEDUP, dedup);
         }
 
+        //
+        // Position attributes
+        //
+
+        for (Entry<String,String> attr: attributes.entrySet()) {
+          conn.setRequestProperty(Warp10InputFormat.HTTP_HEADER_ATTR_PREFIX + attr.getKey(), attr.getValue());
```
pi-r-p, line 197: Could it be a good idea here to raise a runtime exception if the attribute contains a dot ? It can avoid painful debugging later.

warp10/src/main/java/io/warp10/hadoop/Warp10InputFormat.java
hbs: not supposed to change, and in case of change better not to propagate it.

warp10/src/main/java/io/warp10/hadoop/Warp10RecordReader.java
```diff
@@ -198,6 +226,7 @@ public void initialize(InputSplit split, TaskAttemptContext context)
         break;
       } catch (Exception e) {
         LOG.error(e.getMessage(), e);
+        throw e;
```
pi-r-p, line 229: Why here ? Is it safe not to disconnect as initially planned in finally ?

pi-r-p (approved):

PR #1146, opened 2022-11-04
Added support for attributes

hbs:


pi-r-p:
Same remarks as #1145

PR #1149, opened 2022-11-10
Refactored WarpFleet functions to rely on capability for using WF.* functions

hbs:


pi-r-p (changes requested):
Does not compile.
```
WarpScriptLib.register(new WarpFleetWarpScriptExtension());
```
is still present in `io/warp10/script/WarpFleetMacroRepository.java`

warp10/src/main/java/io/warp10/script/functions/WFGETREPOS.java
```diff
@@ -14,40 +14,32 @@
...
   public WFGETREPOS(String name) {
     super(name);
   }
 
   @Override
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
-    
-    if (disabled) {
+
+    if (null == Capabilities.get(stack, WarpScriptStack.CAPABILITY_WFSET)) {
```
pi-r-p, line 37: must be WFGET here.

pi-r-p (changes requested):

pi-r-p (approved):

PR #1151, opened 2022-11-14
Fixed capabilities handling in GUARD calls

hbs:


warp10/src/main/java/io/warp10/script/functions/GUARD.java
```diff
@@ -114,8 +137,52 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+              }
+              capabilities.remove(cap);
+              capabilities.putIfAbsent(cap, newcaps.get(cap));
+            }
+          }
+        }
+      }
+
+      if (null != capabilities) {
+        Capabilities.set(stack, capabilities);
```
steven-gueguen, line 170: If there is no capability set with before the call to GUARD, all capabilities in macro will leak.

warp10/src/main/java/io/warp10/script/functions/GUARD.java
```diff
@@ -114,8 +137,52 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      if (null != caps) {
+        // Export the specified capabilities if no error occurred
+        if (null == error) {
+          Capabilities newcaps = caps;
+          for (String cap: getExportedCapabilities(stack)) {
+            if (null != cap) {
+              if (copyExported) {
+                ((AtomicReference<Set>) exportedCapabilities).get().add(cap);
+              }
+              capabilities.remove(cap);
```
steven-gueguen, line 162: If capabilities is null, the export will fail

hbs:
PTAL

steven-gueguen (approved):

pi-r-p (changes requested):
There is one weird edge case: CAPEXPORT also export cap deletion when the guarded macro does not fail. 

```
<%
  SAVE '.context' STORE
  // token with 'runner.reschedule.min.period' 2000 and "http.size": "100000000"  
  CAPADD // will  leak outside the macro, with 'runner.reschedule.min.period' 2000
  <%
    'runner.reschedule.min.period' CAPDEL // erase the previous one (2000)
    // token with 'runner.reschedule.min.period' 10
    
    CAPADD 
    'runner.reschedule.min.period' CAPGET  "10" == 'not 10 here ?' ASSERTMSG
    'runner.reschedule.min.period' CAPDEL // erase it
    //'runner.reschedule.min.period' CAPGET ISNULL 'is not erased here ?' ASSERTMSG

    // should not export "runner.reschedule.min.period": null, but does so...
    [ 'runner.reschedule.min.period' '' 'toto' ] CAPEXPORT

    //FAIL 
    DROP
    'result of the macro'
  %> 
  $.context -1 GUARD

%>
'macro' STORE

[ 'runner.reschedule.min.period' 'http.size' 'toto' ] CAPGET 
// "runner.reschedule.min.period": "2000"  if the macro fails
// "runner.reschedule.min.period": null if the macro does not fail
```

hbs:
The GUARD call restores the capabilities to their value pre-call, except for those which were altered during the call and which were explicitely exported.

Isn't that excactly what you observed?

PR #1153, opened 2022-11-16
Added support for default token attributes

hbs:


warp10/src/main/java/io/warp10/continuum/Tokens.java
```diff
@@ -299,55 +353,63 @@ public static ReadToken extractReadToken(String token) throws WarpScriptExceptio
...
+      }
     }
-    
-    for (String attr: blockedAttributes) {
-      if (attributes.containsKey(attr)) {
-        throw new RuntimeException("Invalid token attribute.");
+
+    if (null != defaults) {
+      if (null == attributes) {
+        attributes = new LinkedHashMap<String,String>(defaults);
```
steven-gueguen, line 376: Defaults attributes wont' be propagated

steven-gueguen (changes requested):

hbs:
PTAL

etc/conf.templates/distributed/00-warp.conf.template
```diff
@@ -103,6 +103,18 @@ warp.components =
...
+// be overriden if they exist
+//
+warp.token.read.attributes.default = { '.cap:limits' '' }
+
+//
+// WarpScript map defining the default attributes to allocate to write tokens. Existing attributes will not
+// be overriden if they exist
+//
+#warp.token.write.attributes.default = { '.nodelete' '' }
+
```
pi-r-p, line 117: Not the best example, .nodelete leads to a non delete instance.

pi-r-p (changes requested):

hbs:
PTAL

steven-gueguen (approved):

PR #1161, opened 2022-11-21
Added capability support. Integrated FUNCTIONS in core lib

hbs:


warp10/src/main/java/io/warp10/script/functions/FUNCTIONS.java
```diff
@@ -31,6 +32,10 @@ public FUNCTIONS(String name) {
 
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
+    if (null == Capabilities.get(stack, WarpScriptStack.CAPABILITY_INVENTORY)) {
+      throw new WarpScriptException(getName() + " missing capability.");
```
pi-r-p, line 36: missing the capability name in the message.

pi-r-p (changes requested):

warp10/src/main/java/io/warp10/script/functions/FUNCTIONS.java
hbs: The capability name is not missing, it is intentionally omitted. If you don't have a token with the capability set then it is highly probable that it is because you are not supposed to have it, so don't bother asking your Warp 10 admin for it. We want to protect the admin from demanding users :-)

pi-r-p: Doing so, you add a workload to the admin, that must look for the solution, and may do silly things, such as copy paste all the possible authorizations that he will find on a blog post to get rid of this demanding user.

hbs: BOFH rulez ;-)

hbs:
PTAL

pi-r-p (approved):

PR #1163, opened 2022-11-23
Added config for default capabilities

hbs:


warp10/src/main/java/io/warp10/warp/sdk/Capabilities.java
```diff
@@ -37,6 +40,31 @@ public class Capabilities {
...
+        }
+      } catch (WarpScriptException wse) {
+        throw new RuntimeException("Error initializing default capabilities.");
+      }
+    }
+  }
+
   public static String get(WarpScriptStack stack, String name) {
     Capabilities capabilities = null;
     if (stack.getAttribute(WarpScriptStack.CAPABILITIES_ATTR) instanceof Capabilities) {
```
steven-gueguen, line 70: Default capabilities should be added here also

steven-gueguen (changes requested):

warp10/src/main/java/io/warp10/warp/sdk/Capabilities.java
```diff
@@ -56,14 +84,15 @@ public static String get(WarpScriptStack stack, String name) {
     if (null != capabilities) {
       return capabilities.capabilities.get(name);
     } else {
-      return null;
+      return DEFAULT_CAPABILITIES.get(name);
```
steven-gueguen, line 87: If they are some capabilities defined, the default one cannot be accessed.

steven-gueguen (changes requested):

hbs:
PTAL

steven-gueguen (approved):

PR #1164, opened 2022-11-24
Added loading of extensions for TokenGen. Corrected classpath for tok…

hbs:
…engen/run commands


pi-r-p (approved):
tested with an internal extension (STDOUT) and an external one (SLEEP)

PR #1171, opened 2022-12-22
check if MAP ticks input are sorted instead of sorting them

pi-r-p:
In case of a concurrent execution of MAP, sorting the outputTicks will lead to a concurrent modification exception.

So, I just check if the input is correctly sorted instead of always calling Collections.sort()

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -5691,8 +5691,12 @@ public class GTSHelper {
         if (reversed) {
           boolean descending = true;
           int i = 1;
+          long lastElt = outputTicks.get(0);
+          long elt;
           while (i < outputTicks.size() && descending) {
-            descending = outputTicks.get(i) <= outputTicks.get(i - 1);
+            elt = outputTicks.get(i);
+            descending = elt <= lastElt;
+            lastElt = elt;
             i++;
           }
           if (!descending) {
@@ -5701,8 +5705,12 @@ public class GTSHelper {
         } else {
           boolean ascending = true;
           int i = 1;
+          long lastElt = outputTicks.get(0);
+          long elt;
           while (i < outputTicks.size() && ascending) {
-            ascending = outputTicks.get(i) >= outputTicks.get(i - 1);
+            elt = outputTicks.get(i);
+            ascending = elt >= lastElt;
+            lastElt = elt;
             i++;
           }
           if (!ascending) {
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -5684,12 +5684,31 @@ public class GTSHelper {
     // Retrieve ticks if GTS is not bucketized.
     long[] ticks = isBucketized(gts) ? null : Arrays.copyOf(mapped.ticks, gts.values);
 
-    // Sort outputTicks
+    // Check if outputTicks is correctly sorted
+    // In case of a concurrent execution, sorting outputTicks here will lead to a ConcurrentModificationException
     if (null != outputTicks) {
-      if (reversed) {
-        Collections.sort(outputTicks, Collections.<Long>reverseOrder());
-      } else {
-        Collections.sort(outputTicks);
+      if (outputTicks.size() > 1) {
+        if (reversed) {
+          boolean descending = true;
+          int i = 1;
+          while (i < outputTicks.size() && descending) {
+            descending = outputTicks.get(i) <= outputTicks.get(i - 1);
+            i++;
+          }
+          if (!descending) {
+            throw new WarpScriptException("The tick list must be reverse sorted.");
+          }
+        } else {
+          boolean ascending = true;
+          int i = 1;
+          while (i < outputTicks.size() && ascending) {
+            ascending = outputTicks.get(i) >= outputTicks.get(i - 1);
+            i++;
+          }
+          if (!ascending) {
+            throw new WarpScriptException("The tick list must be sorted.");
+          }
+        }
       }
     }
 
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2022  SenX S.A.S.
+//   Copyright 2018-2023  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


hbs:
Did you actually check performance improvement for various list sizes? Around what size does the difference become notable?

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -5684,12 +5684,31 @@ public static List<GeoTimeSerie> map(GeoTimeSerie gts, Object mapper, long prewi
...
-      if (reversed) {
-        Collections.sort(outputTicks, Collections.<Long>reverseOrder());
-      } else {
-        Collections.sort(outputTicks);
+      if (outputTicks.size() > 1) {
+        if (reversed) {
+          boolean descending = true;
+          int i = 1;
+          while (i < outputTicks.size() && descending) {
+            descending = outputTicks.get(i) <= outputTicks.get(i - 1);
```
hbs, line 5695: get does a rangecheck so it is better if you keep track of the previous element to avoid calling get twice with the same index.

pi-r-p:
The fix was not designed for performances, but is improves them.

Warp 10 2.11:
- 1 millions points gts, 20 times MAP : 10.6s
- 10 000 points gts, 20 times MAP : 120ms

With current approach:
- 1 millions points gts, 20 times MAP : 8.5s
- 10 000 points gts, 20 times MAP : 80ms

Removing one get is a bit faster:
- 1 millions points gts, 20 times MAP : 7.9s
- 10 000 points gts, 20 times MAP : 77ms

Using for(long l: outputTicks) syntax is worst.

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -5684,12 +5684,39 @@ public static List<GeoTimeSerie> map(GeoTimeSerie gts, Object mapper, long prewi
...
-        Collections.sort(outputTicks, Collections.<Long>reverseOrder());
-      } else {
-        Collections.sort(outputTicks);
+      if (outputTicks.size() > 1) {
+        if (reversed) {
+          boolean descending = true;
+          int i = 1;
+          long lastElt = outputTicks.get(0);
+          long elt;
+          while (i < outputTicks.size() && descending) {
```
hbs, line 5696: keeping track of size in a variable instead of calling size() is probably faster too.
checking from the end and checking if index is > 0 is probably even a tad faster.

pi-r-p: and not zero is faster, it should be an nz flag in the cpu register :+1:

pi-r-p: I tried, but cannot go under 7.9s . Obviously, JVM already optimize this.

hbs: Dorn this Java thing is slow...

pi-r-p: I tested some variations of this code in a small function, there is no visible improvement. jvm, compiler and branch prediction do a good job.

randomboolean:
I think it is better to make this check in io.warp10.script.function.MAP to save entering map() method if the check fails on the argument.

randomboolean:
I think it is better to make this check in io.warp10.script.function.MAP to save entering map() method if the check fails on the argument.

pi-r-p:
> I think it is better to make this check in io.warp10.script.function.MAP to save entering map() method if the check fails on the argument.

Indeed, check once for multiple gts input.

pi-r-p:
PTAL

warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -205,7 +206,7 @@ private Object applyWithParamsFromMap(WarpScriptStack stack, Map params) throws
     }
 
     Object outputTicks = params.get(PARAM_OUTPUTTICKS);
-    // Make sure outputTicks is a List<Long>
+    // Make sure outputTicks is a List<Long>, and that it the list is sorted accordingly.
```
randomboolean, line 209: fix typo

warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -215,6 +216,40 @@ private Object applyWithParamsFromMap(WarpScriptStack stack, Map params) throws
           throw new WarpScriptException(getName() + " expects '" + PARAM_OUTPUTTICKS + "' to be list of LONG values.");
         }
       }
+
+      // Check if outputTicks is correctly sorted
+      // In case of a concurrent execution, sorting outputTicks here will lead to a ConcurrentModificationException
```
randomboolean, line 221: replace "will lead" by "would lead"

warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -215,6 +216,40 @@ private Object applyWithParamsFromMap(WarpScriptStack stack, Map params) throws
...
+      if (((List<Long>) outputTicks).size() > 1) {
+        if (occurrences < 0) { // reversed
```
randomboolean, line 223: create a final variable called "reversed" to make the code a bit more readable,
then reuse this variable when calling map

warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -215,6 +216,40 @@ private Object applyWithParamsFromMap(WarpScriptStack stack, Map params) throws
...
+          long lastElt = ((List<Long>) outputTicks).get(0);
+          long elt;
+          while (i < ((List<Long>) outputTicks).size() && descending) {
+            elt = ((List<Long>) outputTicks).get(i);
+            descending = elt <= lastElt;
+            lastElt = elt;
+            i++;
+          }
+          if (!descending) {
+            throw new WarpScriptException("The tick list must be reverse sorted.");
```
randomboolean, line 235: use getName() method

warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -215,6 +216,40 @@ private Object applyWithParamsFromMap(WarpScriptStack stack, Map params) throws
...
+          long lastElt = ((List<Long>) outputTicks).get(0);
+          long elt;
+          while (i < ((List<Long>) outputTicks).size() && ascending) {
+            elt = ((List<Long>) outputTicks).get(i);
+            ascending = elt >= lastElt;
+            lastElt = elt;
+            i++;
+          }
+          if (!ascending) {
+            throw new WarpScriptException("The tick list must be sorted.");
```
randomboolean, line 249: ditto

warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -26,6 +26,7 @@
 import io.warp10.script.WarpScriptException;
 import io.warp10.script.WarpScriptStack;
 
+import java.awt.peer.LabelPeer;
```
randomboolean, line 29: fix typo

pi-r-p:
PTAL

randomboolean (approved):

PR #1172, opened 2022-12-22
add SLEEP function and capability

pi-r-p:
When pushing data to external services (using HTTP or WEBCALL), we might need to respect a minimum period between requests. Current solution to do so is an infinite loop bounded by TIMEBOX, quite CPU intensive.

Adds `sleep.maxtime` capability.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/SLEEP.java
+++ b/warp10/src/main/java/io/warp10/script/functions/SLEEP.java
@@ -23,6 +23,8 @@ import io.warp10.script.WarpScriptStack;
 import io.warp10.script.WarpScriptStackFunction;
 import io.warp10.warp.sdk.Capabilities;
 
+import java.util.concurrent.locks.LockSupport;
+
 public class SLEEP extends NamedWarpScriptFunction implements WarpScriptStackFunction {
 
   public SLEEP(String name) {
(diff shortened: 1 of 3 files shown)


warp10/src/main/java/io/warp10/script/functions/SLEEP.java
```diff
@@ -0,0 +1,70 @@
...
+    if (maxSleepMs <= 0) {
+      throw new WarpScriptException(getName() + " requires capability " + WarpScriptStack.CAPABILITY_SLEEP_MAXTIME + " to be set to a value strictly greater than 0 ms.");
+    }
+
+    Object o = stack.pop();
+    if (!(o instanceof Long)) {
+      throw new WarpScriptException(getName() + " expects a LONG period as parameter.");
+    }
+    // convert to milliseconds    
+    long millis = ((Long) o).longValue() / Constants.TIME_UNITS_PER_MS;
```
hbs, line 56: The value of the capability should be expressed in milliseconds since the token in itself does not bear the notion of time units.

warp10/src/main/java/io/warp10/script/functions/SLEEP.java
```diff
@@ -0,0 +1,70 @@
...
+    if (millis > maxSleepMs) {
+      throw new WarpScriptException(getName() + " cannot sleep during more greater than " + maxSleepMs + " ms, as defined in " + WarpScriptStack.CAPABILITY_SLEEP_MAXTIME + " capability.");
```
hbs, line 58: Please correct error message.

warp10/src/main/java/io/warp10/script/functions/SLEEP.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    try {
+      Thread.sleep(millis);
```
hbs, line 62: Use LockSupport.parkNanos instead

hbs:
Change copyright notices to -2023

pi-r-p:
Happy new year commit.

steven-gueguen (approved):

PR #1176, opened 2023-01-02
Fixed TokenGen so there is no risk of overwriting the input file when…

hbs:
… no output was specified.


warp10/src/main/java/io/warp10/TokenGen.java
```diff
@@ -92,18 +91,14 @@ public void parse(String[] args) throws Exception {
 
   public void usage(String[] args) {
     if (args.length < 3) {
-      System.err.println("Usage: TokenGen config ... in out");
+      System.err.println("Usage: TokenGen config ... in");
```
steven-gueguen, line 94: Should be `Arrays.copyOf(args, args.length - 1))` line 54

steven-gueguen (changes requested):

hbs:
PTAL

steven-gueguen (approved):

PR #1177, opened 2023-01-06
Output ticks in MAP

### LINKED ISSUE #953: Behavior of 'ticks' in MAP framework is disconcerting
Although the documentation of `ticks` in [`MAP`](https://www.warp10.io/doc/MAP) it somewhat clear on how MAP behave with this parameter, it is rather intuitive to use. In the following example, we could expect all the values to be replaced by 0, but is is not the case:

```
[] 0 10 <% +! %> FOR
'0to10' STORE

$0to10
[] [] []
$0to10
MAKEGTS

{ 
  'mapper' 0 mapper.replace 
  'ticks' $0to10
} 
MAP
```

Instead the resulting GTS is empty.

I wonder why the `ticks` parameter behave this way. Is there some logic I am missing or is it just because it is simpler to code this behavior?

I propose we drop the pre/post counting logic and make `MAP` anchor its windows like if `ticks` exists in the GTS. In the following, with have 3 ticks in `ticks` which anchors 3 windows. Blue ones are time based and green ones are 1-pre, 1-post.
map_ticks

On the last blue window, the mapper would be given the current tick value from `ticks` but empty Lists for ticks, lats, lons, elevs and values.

Changing this behavior may be ill-advised to avoid breaking code, but it could be done in the next major version.

randomboolean:
- Enable passing output ticks parameter when using the input list syntax of MAP instead of occurrences parameters
- Fix #953 : the current output tick is no longer counted in parameter pre (or post) if it matches a tick of the input series


steven-gueguen (changes requested):
Please also update the corresponding tests.

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -5803,39 +5800,26 @@ public static List<GeoTimeSerie> map(GeoTimeSerie gts, Object mapper, long prewi
...
-            // Note that if output ticks are provided,
-            // then if the current output tick matches an input tick,
-            // then if (reversed)
-            // then prewindow counts the current tick
-            // else it is postwindow that counts the current tick
-            //
-
             if (reversed) {
-              start = leftIdx - 1 + prewindow < ticks.length ? (ticks[leftIdx - 1 + (int) prewindow]) : Long.MIN_VALUE;
+              // if the current output tick matches an input tick then there will be one tick more in the sliding window
```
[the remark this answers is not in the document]
randomboolean: Maybe we can allow output tick to be also reverse sorted if the input syntax is a list or if the the input syntax is a map and occurrences is 0.

warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -205,14 +203,48 @@ private Object applyWithParamsFromMap(WarpScriptStack stack, Map params) throws
...
+      // In case of a concurrent execution, sorting outputTicks here would lead to a ConcurrentModificationException
+      if (((List<Long>) outputTicks).size() > 1) {
+        if (reversed) { // reversed
+          boolean descending = true;
+          int i = 1;
+          long lastElt = ((List<Long>) outputTicks).get(0);
+          long elt;
+          while (i < ((List<Long>) outputTicks).size() && descending) {
+            elt = ((List<Long>) outputTicks).get(i);
+            descending = elt <= lastElt;
```
steven-gueguen, line 227: Should it be `descending = elt < lastElt;` instead?

warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -205,14 +203,48 @@ private Object applyWithParamsFromMap(WarpScriptStack stack, Map params) throws
...
+            throw new WarpScriptException(getName() + " expects a reverse sorted list for " + PARAM_OUTPUTTICKS + " parameter.");
+          }
+        } else {
+          boolean ascending = true;
+          int i = 1;
+          long lastElt = ((List<Long>) outputTicks).get(0);
+          long elt;
+          while (i < ((List<Long>) outputTicks).size() && ascending) {
+            elt = ((List<Long>) outputTicks).get(i);
+            ascending = elt >= lastElt;
```
steven-gueguen, line 241: Should it be `ascending = elt > lastElt;` instead?

steven-gueguen (changes requested):

steven-gueguen (changes requested):
Please correct the unit test accordingly.

pi-r-p (changes requested):
To discuss: 
- We must fail instead of doing unexpected things
Ready for your review:
- tests on 3.x branch in documentation (compile with revision 3.0.0).

warp10/src/main/java/io/warp10/script/functions/MAP.java
```diff
@@ -138,6 +141,14 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
   }
 
   private Object applyWithParamsFromMap(WarpScriptStack stack, Map params) throws WarpScriptException {
+    if (params.containsKey(DEPRECATED_PARAM_OCCURENCES)) {
+      throw new WarpScriptException(getName() + " was given a deprecated parameter: " + DEPRECATED_PARAM_OCCURENCES + ". Use instead parameter " + PARAM_OCCURRENCES);
+    }
+
+    if (params.containsKey(DEPRECATED_PARAM_TICKS)) {
+      throw new WarpScriptException(getName() + " was given a deprecated parameter: " + DEPRECATED_PARAM_TICKS + ". Use instead parameter " + PARAM_OCCURRENCES);
+    }
```
pi-r-p, line 150: Can you add something like "parameter name and behavior are changed, read the doc"
It is not a simple "replace".

pi-r-p (changes requested):

pi-r-p (approved):

steven-gueguen (approved):

PR #1180, opened 2023-01-11
add HTTP timeout

pi-r-p:
Added 'timeout' parameter for `HTTP` function.
- Default timeout is 120 s
- `warpscript.http.maxtimeout` configuration override the default timeout
- `http.maxtimeout` override the configuration

Values 0 and MAXLONG (or anything greater than maximum integer) are allowed for "no timeout at all".


hbs:
Timeout should have a maximum value set in configuration and a capability to extend this maximum value for specific users.

pi-r-p:
ready for review

steven-gueguen (commented):
The 'http.maxtimeout' capability is only used if the user specifies a timeout parameter in the HTTP request.
If no timeout is specified in the parameter, then even if you set the 'http.maxtimeout' capability, the timeout defined in the Warp 10 configuration that will be used.

Is this the intended behavior ?

pi-r-p:
It is. 
Capability is here to allow more than the configuration. (same as maxops for example... Even if you have it, you must call MAXOPS explicitly)

PR #1181, opened 2023-01-16
Added maximum waiting time for MUTEX. Made MUTEX reentrant if by the …

hbs:
…same thread with the same mutex.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/ext/shm/MUTEX.java
+++ b/warp10/src/main/java/io/warp10/script/ext/shm/MUTEX.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2022  SenX S.A.S.
+//   Copyright 2018-2023  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package io.warp10.script.ext.shm;
 
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.regex.Pattern;
 
@@ -23,8 +24,8 @@ import io.warp10.script.NamedWarpScriptFunction;
 import io.warp10.script.WarpScriptException;
 import io.warp10.script.WarpScriptStack;
 import io.warp10.script.WarpScriptStack.Macro;
-import io.warp10.warp.sdk.Capabilities;
 import io.warp10.script.WarpScriptStackFunction;
+import io.warp10.warp.sdk.Capabilities;
 
 public class MUTEX extends NamedWarpScriptFunction implements WarpScriptStackFunction {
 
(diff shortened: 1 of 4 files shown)


warp10/src/main/java/io/warp10/script/ext/shm/MUTEX.java
```diff
@@ -54,6 +52,10 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
     String mutex = String.valueOf(top);
 
+    if(null != stack.getAttribute(MUTEX_ATTRIBUTE + stack.getUUID()) && mutex != stack.getAttribute(MUTEX_ATTRIBUTE + stack.getUUID())) {
```
steven-gueguen, line 55: Should be `==`

steven-gueguen (changes requested):

warp10/src/main/java/io/warp10/script/ext/shm/MUTEX.java
hbs: The condition has been changed to use equals() but it is correct in the sense that we want to detect nested calls to MUTEX with different mutex names.

pi-r-p (changes requested):
The configuration and capability introduce a "maximum". So the user should be able to pass a parameter for timeout, that will be allowed or not by the max.
- could you change the signature to accept a LONG in platform time unit on top ? `<% %> 3 s MUTEX` .
- could you forbid negative values for maxwait and the new argument ? It makes no sense and always lead to a failure to acquire the mutex.
- should 0 be kept as a special "no limit" value, or should the function accept `<% %> MAXLONG MUTEX` when mutex.maxwait is 0 ? better to support both... If the cap or conf is defined as 0, then user can put whatever positive value as maxtime argument.

warp10/src/main/java/io/warp10/script/ext/shm/MUTEX.java
```diff
@@ -68,17 +70,41 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
     ReentrantLock lock = SharedMemoryWarpScriptExtension.getLock(mutex);
 
+    long maxwait = SharedMemoryWarpScriptExtension.MUTEX_DEFAULT_MAXWAIT;
+
+    if (null != Capabilities.get(stack, SharedMemoryWarpScriptExtension.CAPABILITY_MUTEX_MAXWAIT)) {
+      try {
+        maxwait = Long.parseLong(Capabilities.get(stack, SharedMemoryWarpScriptExtension.CAPABILITY_MUTEX_MAXWAIT));
+      } catch(NumberFormatException nfe) {
+        throw new WarpScriptException(getName() + " invalid '" + SharedMemoryWarpScriptExtension.CAPABILITY_MUTEX_MAXWAIT + "' capability value.");
```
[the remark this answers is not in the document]
hbs: Omitting the mention of unit to avoid having values such as "1000 ms" in the capability.

pi-r-p (changes requested):

pi-r-p (approved):

warp10/src/main/java/io/warp10/script/ext/shm/MUTEX.java
```diff
@@ -38,22 +40,56 @@ public MUTEX(String name) {
   public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
     String cap = Capabilities.get(stack, SharedMemoryWarpScriptExtension.CAPABILITY_MUTEX);
+
     if (null == cap) {
-      throw new WarpScriptException(getName() + " expected capability '" + SharedMemoryWarpScriptExtension.CAPABILITY_MUTEX + "' to be set.");
+      throw new WarpScriptException(getName() + " expected capability '" + SharedMemoryWarpScriptExtension.CAPABILITY_MUTEX + "' to be set to a value >= 0.");
     }
```
pi-r-p, line 47: Wrong error message. Should refer to a pattern, not a number.

pi-r-p (changes requested):

hbs:
PTAL

pi-r-p (approved):

PR #1185, opened 2023-01-20
Fix BUCKETIZE.CALENDAR

randomboolean:
In BUCKETIZE.CALENDAR, the computation of the underlying bucketoffset wasn't correct in some cases.

In the preliminary calculations of bucketoffset and bucket index of lastbucket, we were sliding Epoch up to its congruent timestamp contained in the lastbucket using provided timezone parameter. Instead, we now use UTC timezone so this calculation does not depend on the timezone.

Also, for each addition or subtraction of multiple periods from an origin timestamp, we now always add all periods at once instead of adding a single period from a previous intermediate timestamp. With this change we avoid cases where less duration would have been added than expected if an intermediate timestamp would have landed in the margin of a period component (for example 28th of February instead of 29th).


steven-gueguen (approved):

hbs:
Can you re-submit this PR against 3.x and 2.x if you want it fixed theretoo

PR #1186, opened 2023-01-27
fix variance and sd snapshot bug

randomboolean:


steven-gueguen (changes requested):
Please update copyright date

hbs:
Need to provide another PR against 3.x

steven-gueguen (approved):

PR #1188, opened 2023-01-30
Initial commit of XEVAL

hbs:


warp10/src/main/java/io/warp10/standalone/StandaloneStoreClient.java
```diff
@@ -250,6 +250,9 @@ public void remove() {
 
       @Override
       public GTSDecoder next() {
+        if (idx >= metadatas.size()) {
```
steven-gueguen, line 253: Update copyright date, please

steven-gueguen (changes requested):

steven-gueguen (approved):

PR #1197, opened 2023-02-15
Removal of WEBCALL related code

hbs:


pi-r-p (changes requested):
There is still code generated for Webcall, from thrift scripts:

warp10/src/main/thrift/io_warp10_script_thrift_data.thrift

hbs:
PTAL

pi-r-p (approved):

PR #1203, opened 2023-03-01
Standalone+ implementation

hbs:


warp10/src/main/java/io/warp10/standalone/Warp.java
```diff
@@ -135,16 +136,48 @@ public static void main(String[] args) throws Exception {
...
+        case Constants.BACKEND_PLASMA:
+          plasmabackend = true;
+          break;
+        case Constants.BACKEND_MEMORY:
+          inmemory = true;
+          break;
+        default:
+          throw new RuntimeException("Missing '" + Configuration.BACKEND + "' specification.");
+      }
+    }
```
pi-r-p, line 174: switch cannot accept null.
So when backend is undefined (the default, for people who will migrate to 3.x from 2.x standalone), it must be managed before, and should default to leveldb.

pi-r-p (changes requested):

warp10/src/main/java/io/warp10/standalone/Warp.java
```diff
@@ -138,16 +139,51 @@ public static void main(String[] args) throws Exception {
...
-    boolean plasmabackend = "true".equals(properties.getProperty(Configuration.PURE_PLASMA));
+    if (analyticsEngineOnly) {
+      nullbackend = true;
+      if (null != backend && !Constants.BACKEND_NULL.equals(backend)) {
+        LOG.warn("Backend specification '" + backend + "' ignored since '" + Configuration.ANALYTICS_ENGINE_ONLY + " is true.");
+      }
+    } else {
+      if (null == backend) {
+        throw new RuntimeException("Backend specification '" + backend + "' MUST be set since Warp 10 3.0");
+      }
```
pi-r-p, line 160: "Backend specification '" + **Configuration.BACKEND** + "' MUST be set since Warp 10 3.0"

Message may also include the thee main possible choices Constants.BACKEND_LEVELDB , Constants.BACKEND_FDB , Constants.BACKEND_MEMORY

hbs: > "Backend specification '" + **Configuration.BACKEND** + "' MUST be set since Warp 10 3.0"

 done

> Message may also include the thee main possible choices Constants.BACKEND_LEVELDB , Constants.BACKEND_FDB , Constants.BACKEND_MEMORY

Not fan of that type of message, the user needs to know what s/he's doing.

etc/conf.templates/standalone/10-fdb.conf.template
```diff
@@ -0,0 +1,62 @@
...
+//
+// F o u n d a t i o n D B
+//
+/////////////////////////////////////////////////////////////////////////////////////////
+
+
+##
+## This the API version to use for FoundationDB
+##
+fdb.api.version = 710
```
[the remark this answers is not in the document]
hbs: The requirement is 7.2 since we need tenants. Results are undefined if using 6.x or any prior version. The actual api version will depend on the FDB version deployed so it will need to be changed in some cases.

We consider the users sufficiently competent if they consider using FDB, so no need to treat them as children and hold their hand all the time.

PR #1204, opened 2023-03-01
DatalogNG

hbs:


warp10/src/main/java/io/warp10/standalone/datalog/DatalogWorkers.java
```diff
@@ -0,0 +1,94 @@
...
+
+  public static void offer(DatalogConsumer consumer, String ref, DatalogRecord record) throws IOException {
+    //
+    // Compute partitioning key and partition
+    //
+
+    long classid = record.getMetadata().getClassId();
+    long labelsid = record.getMetadata().getLabelsId();
+
+    long partkey = (int) (((classid << 16) & 0xFFFF0000L) | ((labelsid >>> 48) & 0xFFFFL));
```
steven-gueguen, line 77: Cast should be remove

steven-gueguen (changes requested):

warp10/src/main/java/io/warp10/standalone/datalog/DatalogWorker.java
```diff
@@ -0,0 +1,123 @@
...
+              break;
+          }
+        }
+      } catch (Throwable t) {
+        err = t;
+        if (null != job) {
+          job.consumer.failure(job.ref);
+        }
+      } finally {
+        labels.put(SensisionConstants.SENSISION_LABEL_TYPE, job.record.getType().name());
```
steven-gueguen, line 114: This could lead to NullPointerException.
There are some cases where job is null or job.record.getType() is null.

steven-gueguen (changes requested):

warp10/src/main/java/io/warp10/standalone/datalog/DatalogWorker.java
hbs: When can job be null? There is an explicit conditional that continues the loop if job is null. When can job ne non null and type be null?

hbs: Solved, the job is null because the call to continue will first go through the finally block. For the type it was a concurrency issue which has been fixed in commit 677fd1e.

steven-gueguen (changes requested):

steven-gueguen (approved):

PR #1205, opened 2023-03-02
Added finer grain capabilities

hbs:


pi-r-p (changes requested):
These functions are added to WarpScript in a static block of FDBStoreClient.

Why isn't it in WarpScriptLib, to be consistent with leveldb functions ? 

Usage: 
In a standalone (leveldb), I may want to REXEC something to a fdb one, and this will fail:
```
<% 
  $x FDBGET   //will fail here, not able to build the function
%> { 'x' $tenant } DEREF  $url REXEC
```

hbs:
The functions are only added when using FDB. If you want to do what you showed, use 'FDBGET' EVAL

pi-r-p:
I know the workaround, but it is not as convenient. 

Functions are defined as a standard or by an extension, not by the backend type. And you already check the backend in all these functions, so it won't be a problem to treat them as leveldb ones.

In the current documentation process, it will more easy not to add this kind of exceptions.

hbs:
Those functions were added in PR#1192 that you approved. We can modify the way they are defined but those types of concerns should be raised within the convo of the appropriate PR.

pi-r-p (changes requested):
- `FDBSIZE` should have the same optional regexp whitelist as `FDBTENANT`.

Outside of this PR scope, but not spotted before :
- `FDBTENANT` returns an empty map when tenant does not exists. `FDBSIZE` raises an exception. It is not consistent.
- Is there a way to list all existing tenants ? (it could be very handy, and restricted to fdb.admin cap)

warp10/src/main/java/io/warp10/fdb/FDBSIZE.java
```diff
@@ -34,8 +36,18 @@ public FDBSIZE(String name) {
...
+    if (null != Capabilities.get(stack, FDBUtils.CAPABILITY_ADMIN)) {
+      regexp = null;
+    } else if (null == Capabilities.get(stack, FDBUtils.CAPABILITY_SIZE)) {
+      regexp = Capabilities.get(stack, FDBUtils.CAPABILITY_SIZE);
+      // Empty string means match all tenants
+      if ("".equals(regexp)) {
+        regexp = null;
+      }
+    } else {
+      throw new WarpScriptException(getName() + " missing '" + FDBUtils.CAPABILITY_SIZE + "' or '" + FDBUtils.CAPABILITY_ADMIN + "' capability.");
```
pi-r-p, line 50: if you do not have CAPABILITY_ADMIN nor CAPABILITY_SIZE, you can read everything !

warp10/src/main/java/io/warp10/fdb/FDBTENANT.java
```diff
@@ -32,8 +35,18 @@ public FDBTENANT(String name) {
...
+
+    if (null != Capabilities.get(stack, FDBUtils.CAPABILITY_ADMIN)) {
+      regexp = null;
+    } else if (null == Capabilities.get(stack, FDBUtils.CAPABILITY_TENANT)) {
+      regexp = Capabilities.get(stack, FDBUtils.CAPABILITY_TENANT);
+      // Empty string means match all tenants
+      if ("".equals(regexp)) {
+        regexp = null;
+      }
+    } else {
```
pi-r-p, line 48: Same wrong condition path, everything is allowed with no caps at all.

pi-r-p (changes requested):

hbs:
PTAL

pi-r-p (approved):

PR #1206, opened 2023-03-03
fix IDENT function

pi-r-p:
`warp.ident` is defined into 00-warp.conf, but `IDENT` only look at System variables.


hbs:
This is more a change of paradigm than a fix

PR #1209, opened 2023-03-06
Modified error message when URL is rejected

hbs:


warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
```diff
@@ -335,7 +335,7 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
     }
 
     if (!webAccessController.checkURL(url)) {
-      throw new WarpScriptException(getName() + " invalid host or scheme in URL.");
+      throw new WarpScriptException(getName() + " URL forbidden by configuration.");
     }
```
pi-r-p, line 339: Suggestions more than change request:
- Could be great to change WebAccessController to expose patternConf, and add it to the error message ?
- Change the message to encourage or discover the use of capability ?

pi-r-p (changes requested):

warp10/src/main/java/io/warp10/script/ext/http/HTTP.java
hbs: No, this would leak some potentially sensitive information.

pi-r-p (approved):

PR #1211, opened 2023-03-06
Cleaned up key definitions

hbs:
This is a preliminary work to the refactoring of bootstrap in warp10.sh

### DIFF
--- a/etc/conf.templates/distributed/00-secrets.conf.template
+++ b/etc/conf.templates/distributed/00-secrets.conf.template
@@ -1,5 +1,5 @@
 //
-//   Copyright 2019-2022  SenX S.A.S.
+//   Copyright 2019-2023  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


pi-r-p:
To discuss: why removing keys from 00-secrets.conf ?  The file name is pretty clear, and this file must remain portable between all the Warp10 configurations to share tokens.

hbs:
Keys are still in 00-secrets, their default value is simply a reference to another generic name which can be generated easily without modifying existing files. Should the user wish to propagate the actual values in 00-secrets they can still do so.

hbs:
> To discuss: why removing keys from 00-secrets.conf ? The file name is pretty clear, and this file must remain portable between all the Warp10 configurations to share tokens.

They are not removed from this file.

hbs:
> So, you plan to update io.warp10.GenerateCryptoKey#fillConfigFile to detect keys and not `= hex:hhhh` in a later PR ?

My intent is to modify the way the keys are generated, we should not attempt to modify a file, rather create a new one and let users decide if they want to keep the definitions in that new file or move them to 00-secrets (or rename the file 00-secrets if that makes sense).

pi-r-p (approved):

PR #1219, opened 2023-03-10
Aggregator nameset cleaning

randomboolean:
This is a work in progress.

To do: remove duplicate reducers and fix naming incoherences.

To help define the correct changes to implement, in the following the reducers are grouped by their null handling behavior.

Remove null values:
- reducer.and.exclude-nulls
- reducer.min
- reducer.max
- reducer.mean.exclude-nulls
- reducer.median
- reducer.mad
- reducer.or.exclude-nulls
- reducer.sum
- reducer.join.forbid-nulls
- reducer.join.nonnull
- reducer.join.urlencoded
- reducer.var
- reducer.var.welford
- reducer.sd
- reducer.sd.welford
- reducer.count.exclude-nulls
- reducer.count.nonnull
- reducer.mean.circular.exclude-nulls
- reducer.rms.exclude-nulls

Keep null values:
- reducer.count
- reducer.count.include-nulls

Skip timestamp if there is at least one null value:
- reducer.and
- reducer.min.forbid-nulls
- reducer.min.nonnull
- reducer.max.forbid-nulls
- reducer.max.nonnull
- reducer.mean
- reducer.or
- reducer.sum.forbid-nulls
- reducer.sum.nonnull
- reducer.join
- reducer.var.forbid-nulls
- reducer.var.welford.forbid-nulls
- reducer.sd.forbid-nulls
- reducer.sd.welford.forbid-nulls
- reducer.mean.circular
- reducer.rms

Fail if there is at least one null value:
- reducer.median.forbid-nulls
- reducer.percentile.forbid-nulls

### DIFF
--- a/warp10/src/main/java/io/warp10/script/WarpScriptLib.java
+++ b/warp10/src/main/java/io/warp10/script/WarpScriptLib.java
@@ -2630,7 +2630,6 @@ public class WarpScriptLib {
     addNamedWarpScriptFunction(new MapperKernelUniform("mapper.kernel.uniform"));
 
     addNamedWarpScriptFunction(new Percentile.Builder("mapper.percentile", false));
-    addNamedWarpScriptFunction(new Percentile.Builder("mapper.percentile.forbid-nulls", true));
 
     //functions.put("mapper.abscissa", new MapperSAX.Builder());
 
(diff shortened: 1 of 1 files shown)


Giwi (approved):

pi-r-p (changes requested):
Do not approve for alpha release.
It removes functions without functional replacement. People cannot migrate their codebase in this case.

randomboolean:
Only functions that have no use have been removed by the current commits. Duplicate functions are not removed yet.

hbs:
> > > Only functions that have no use have been removed by the current commits. Duplicate functions are not removed yet. The functions that are removed are mapper and bucketizer that operate on windows with null values. That cannot happen. I bleieve that the reason they exist is because mappers and bucketizers could be used with REDUCE, and these functionalities could have been used by some old scripts. However the functional replacements still exist, it is just the equivalent reducers which are not removed.
> > 
> > 
> > Can you elaborate the "that have no use" part?
> 
> I meant that we can use the same mapper or bucketizer without the suffix, it will have the same result.

Can you provide the match between removed function and replacement one, so the audit tool can be updated accordingly.

randomboolean:
```
{
    "bucketizer.count.exclude-nulls": "bucketizer.count",
    "bucketizer.count.include-nulls": "bucketizer.count",
    "bucketizer.count.nonnull": "bucketizer.count",
    "bucketizer.join.forbid-nulls": "bucketizer.join",
    "bucketizer.max.forbid-nulls": "bucketizer.max",
    "bucketizer.mean.circular.exclude-nulls": "bucketizer.mean.circular",
    "bucketizer.mean.exclude-nulls": "bucketizer.mean",
    "bucketizer.median.forbid-nulls": "bucketizer.median",
    "bucketizer.min.forbid-nulls": "bucketizer.min",
    "bucketizer.percentile.forbid-nulls": "bucketizer.percentile",
    "bucketizer.sd.forbid-nulls": "bucketizer.sd",
    "bucketizer.sum.forbid-nulls": "bucketizer.sum",
    "mapper.count.exclude-nulls": "mapper.count",
    "mapper.count.include-nulls": "mapper.count",
    "mapper.count.nonnull": "mapper.count",
    "mapper.join.forbid-nulls": "mapper.join",
    "mapper.max.forbid-nulls": "mapper.max",
    "mapper.mean.circular.exclude-nulls": "mapper.mean.circular",
    "mapper.mean.exclude-nulls": "mapper.mean",
    "mapper.median.forbid-nulls": "mapper.median",
    "mapper.min.forbid-nulls": "mapper.min",
    "mapper.percentile.forbid-nulls": "mapper.percentile",
    "mapper.sd.forbid-nulls": "mapper.sd",
    "mapper.sum.forbid-nulls": "mapper.sum",
    "mapper.var.forbid-nulls": "mapper.var"

```}

randomboolean:
However, this replacement works if the mapper / bucketizer is used in a MAP / BUCKETIZE.
If they are used with REDUCE, the right replacement is to replace the prefixes (mapper and bucketizer by reducer)

pi-r-p:
I don't think people use mapper.xx inside REDUCE (only those who read the code knows that it is the same under the hood).

Issue #1222, opened 2023-03-21
max execution time versus stop interruption


pi-r-p:
- Example 1
```
<% 
 5 STOP 
%> 10000 TIMEBOX
```
- Example 2
Enable `egress.maxtime` , and try to run STOP. It leads to an unexpected 500 error with no message.

hbs:
What is the stack trace?

pi-r-p:
The problem seems that WarpScriptATCException such as stop are casted into WarpScriptException by Timebox.

```json
[...]

PR #1225, opened 2023-03-21
fix TIMEBOX STOP handling

pi-r-p:
fix issue #1222

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/TIMEBOX.java
+++ b/warp10/src/main/java/io/warp10/script/functions/TIMEBOX.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2022  SenX S.A.S.
+//   Copyright 2018-2023  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/functions/TIMEBOX.java
```diff
@@ -155,7 +156,9 @@ public Object call() throws Exception {
       }
       throw new WarpScriptException(getName() + " reached the execution time limit (" + maxtime + " " + Constants.timeunit.name() + ").");
     } catch (ExecutionException ee) {
-      if (this.quiet && ee.getCause() instanceof WarpScriptException) {
+      if (ee.getCause() instanceof WarpScriptStopException) {
```
hbs, line 159: Instead of WarpScriptStopException, shouldn't WarpScriptATCException be handled?

pi-r-p:
```
<% 
 BREAK
%> 10000 TIMEBOX
```
must still fail.

pi-r-p: WarpScriptKillException still remains a good question, but it is not used outside the UDP plugin.

PR #1228, opened 2023-03-27
Support for JDK 8 and above

hbs:


warp10/src/main/java/io/warp10/script/functions/REVERSE.java
```diff
@@ -58,10 +58,8 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
       Collections.reverse((List) top);
       stack.push(top);
     } else if (top instanceof String) {
-      if (!this.stable) {
-        top = new String(UnsafeString.getChars(top.toString()));
-      }
-      char[] chars = UnsafeString.getChars(top.toString());
+      char[] chars = ((String) top).toCharArray();
```
pi-r-p, line 61: won't work with char > 0xFFFF , but anyway. no one will reverse emoji. 
- previous behavior on 2.x was to return  "??" for emoji and it transforms the input. (if you try '💩' DUP REVERSE , you see the input is transformed)
- this behavior output \uD83D\uDCA9  and \uDCA9\uD83D reversed. It is better not to transform the input anymore.

So this is a behavior difference, but acceptable.

pi-r-p (changes requested):

warp10/src/main/java/io/warp10/script/WarpScriptLib.java
```diff
@@ -2852,10 +2856,21 @@ public class WarpScriptLib {
...
+    // We need to force Java Version so it looks like x.y.z-aaa while the PApplet class is loaded
     //
 
+    String jversion = WarpConfig.getOriginalFormatJavaVersion();
+
+    if (!jversion.equals(System.getProperty("java.version"))) {
+      String tmp = System.getProperty("java.version");
+      System.setProperty("java.version", jversion);
+      Preconditions.checkArgument(jversion.equals(PApplet.javaVersionName), "Processing was not correctly initialized, consider setting configuraiton " + Configuration.WARP_JAVA_VERSION + ".");
+      System.setProperty("java.version", tmp);
```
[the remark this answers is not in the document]
hbs: Why? We want to actually modify the `java.version` property of the JVM since it is the one checked by Processing.

etc/conf.templates/distributed/00-warp.conf.template
```diff
@@ -36,6 +36,13 @@
 warp.timeunits = us
 
 
+// 
+// Some libraries (like Processing), expect the Java version to be 1.y.z, as this has changed
+// with JDK9+, those libraries may fail to parse the java version. This configuration key
+// can be used to force the "java.version" property to the given value.
+//
+#warp.java.version = 1.11.0
+
```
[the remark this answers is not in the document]
hbs: It may be needed for some extensions or plugins which have dependencies which incorrectly check the java.version

warp10/src/main/java/io/warp10/script/functions/REVERSE.java
hbs: The behavior is indeed undefined if there are composite UNICODE characters coded on more than two UTF-16 characters. But behavior should be identical to that of 2.x.

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -3867,11 +3910,10 @@ public static byte[] unpackGTSId(String s) {
       return bytes;
     }
 
-    char[] c = UnsafeString.getChars(s);
-
     for (int i = 0; i < 8; i++) {
```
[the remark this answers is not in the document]
hbs: Those functions were not added by this PR, so leaving them for now.

warp10/src/main/java/io/warp10/script/WarpScriptLib.java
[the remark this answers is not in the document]
pi-r-p: "consider setting configur**ait**on"

pi-r-p (changes requested):
unused imports `import io.warp10.continuum.gts.UnsafeString`  remains in REVERSE , SNAPSHOT,  HEXTOBIN, WarpURLEncoder

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -3759,9 +3798,13 @@ public static final long labelsId(long sipkey0, long sipkey1, Map<String,String>
       buf[idx++] = (byte) ((hash >> 16) & 0xffL);
       buf[idx++] = (byte) ((hash >> 8) & 0xffL);
       buf[idx++] = (byte) (hash & 0xffL);
+      //bb.putLong(hash);
```
Giwi, line 3801: Commented code could be deleted

warp10/src/main/java/io/warp10/script/StackUtils.java
```diff
@@ -16,6 +16,17 @@
 
 package io.warp10.script;
 
+//import io.warp10.json.JsonUtils;
```
Giwi, line 19: Commented code should be deleted

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -3912,14 +3950,16 @@ public static String gtsIdToString(long classId, long labelsId, boolean intern)
       y >>>= 16;
     }
 
+    String s = new String(c);
```
Giwi, line 3952: Should be replaced by String.valueOf(c)

hbs: Why?

Giwi (approved):

pi-r-p (approved):

PR #1229, opened 2023-03-29
Fixed sanitizeString for cases when quote is followed by a wsp inside a string

hbs:


hbs:
hasSep is actually useless since it always has the same value as instring, will update the PR accordingly

hbs:
PTAL

pi-r-p:
The failure is also be very weird, because it fails at symbol = stmt.substring(1, stmt.length() - 1);
```
" "" "
```
leads to:
`Exception at '"%20"" =>"<=' in section [TOP] (begin 1, end 0, length 1)`

hbs:
> The failure is also be very weird, because it fails at symbol = stmt.substring(1, stmt.length() - 1);
> 
> ```
> " "" "
> ```
> 
> leads to: `Exception at '"%20"" =>"<=' in section [TOP] (begin 1, end 0, length 1)`

Added a check to ensure strings are enclosed in quotes and not just a quote

pi-r-p (approved):
leads to unknown function `"` in audit mode. consistent.

PR #1236, opened 2023-04-14
Sensision runner

steven-gueguen:


warp10/src/main/sh/warp10.sh
```diff
@@ -212,6 +212,18 @@ postInit() {
...
 
+
+  echo "
+//
+// Uncomment the following lines to enable metrics collections.
+// Report here the result of the token generation with the envelope ${WARP10_HOME}/tokens/sensision-tokengen.mc2
+//
+// SENSISIONEVENT, SENSISIONGET, SENSISIONSET, SENSISIONUPDATE
+#warpscript.extension.sensision = io.warp10.script.ext.sensision.SensisionWarpScriptExtension
+#sensisionReadToken@/sensision=
```
hbs, line 223: be more specific in the path, only allowing the script we added to access the tokens

etc/sensision-tokengen.mc2
```diff
@@ -0,0 +1,63 @@
...
+//
+
+//
+// This is an envelope to generate sensision tokens.
+//
+
+JSONPRETTY
+
+'sensision' 'applicationName' STORE
+NOW 100 ADDYEARS 'expiryDate' STORE
```
hbs, line 24: Set the issuance and expiry to fixed values so token generation is deterministic for a given value of IDENT.

Explain that the values are chosen so multiple calls to tokengen sensision-tokengen.mc2 lead to the same token.

hbs (changes requested):

steven-gueguen:
PTAL

hbs:
LGTM

hbs (approved):

PR #1237, opened 2023-04-18
fix leveldb multiline ingestion

pi-r-p:
When trying to ingest compact gts input format on a levelDB standalone, it fails with error 500 and no explaination.
The root cause is that FDBUtils static method `hasCriticalTransactionSize` is not available, because FDBUtils class static block will fail (FDB.selectAPIVersion(version) will fail).

I can fix FDBUtils, but I think is is better to skip the test when not using FDB as a store, for performances.

To test, try to ingest `https://warp10.io/assets/files/ibtracs-2010-2015.gts` on a leveldb 3.x standalone...

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/ingress/IngressStreamUpdateHandler.java
+++ b/warp10/src/main/java/io/warp10/continuum/ingress/IngressStreamUpdateHandler.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2022  SenX S.A.S.
+//   Copyright 2018-2023  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
--- a/warp10/src/main/java/io/warp10/continuum/ingress/IngressStreamUpdateHandler.java
+++ b/warp10/src/main/java/io/warp10/continuum/ingress/IngressStreamUpdateHandler.java
@@ -29,6 +29,7 @@ import io.warp10.continuum.sensision.SensisionConstants;
 import io.warp10.continuum.store.Constants;
 import io.warp10.continuum.store.thrift.data.Metadata;
 import io.warp10.crypto.OrderPreservingBase64;
+import io.warp10.fdb.FDBUtils;
 import io.warp10.quasar.token.thrift.data.WriteToken;
 import io.warp10.sensision.Sensision;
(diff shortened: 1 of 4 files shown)


warp10/src/main/java/io/warp10/standalone/StandaloneIngressHandler.java
```diff
@@ -193,6 +196,8 @@ public StandaloneIngressHandler(KeyStore keystore, StandaloneDirectoryClient dir
     }
 
     this.maxValueSize = Long.parseLong(WarpConfig.getProperty(Configuration.STANDALONE_VALUE_MAXSIZE, DEFAULT_VALUE_MAXSIZE));
+    
+    this.isFDBStore = storeClient instanceof FDBStoreClient;
```
hbs, line 200: This won't work for a sharded instance. You need to determine the backend from the configurationm checking the value of backend

hbs: `this.isFDBStore = Constants.BACKEND_FDB.equals(WarpConfig.getProperty(Configuration.BACKEND))`

hbs (changes requested):

warp10/src/main/java/io/warp10/standalone/StandaloneIngressHandler.java
hbs: a similar check should be added on line 285 of IngressStreamUpdateHandler (without the check for useFDB) and on line 322 of StandaloneStreamUpdateHandler (with the check that we use FDB).

pi-r-p:
I also have a doubt on FDBUtils usage in StandaloneDirectoryClient, for the same reason.

warp10/src/main/java/io/warp10/standalone/StandaloneIngressHandler.java
pi-r-p: check added in both files

pi-r-p:
PTAL

warp10/src/main/java/io/warp10/continuum/ingress/IngressStreamUpdateHandler.java
hbs, line 1: Please update Copyright notice

hbs (changes requested):

pi-r-p:
PTAL

hbs (approved):

PR #1239, opened 2023-04-19
Bumped dependencies to their latest versions

hbs:


warp10/build.gradle
```diff
@@ -56,15 +56,16 @@ dependencies {
...
-    api 'org.eclipse.jetty.websocket:websocket-client:9.4.48.v20220622'
+    api 'org.eclipse.jetty:jetty-server:9.4.51.v20230217'
+    api 'org.eclipse.jetty.websocket:websocket-server:9.4.51.v20230217'
+    api 'org.eclipse.jetty.websocket:websocket-client:9.4.51.v20230217'
     // https://mvnrepository.com/artifact/org.apache.curator/curator-x-discovery
-    api 'org.apache.curator:curator-x-discovery:5.3.0'
-    api 'com.fasterxml.jackson.core:jackson-databind:2.10.2'
-    api 'org.slf4j:slf4j-api:1.7.5'
+    api 'org.apache.curator:curator-x-discovery:5.4.0'
+    //api 'com.fasterxml.jackson.core:jackson-databind:2.10.2'
```
pi-r-p, line 66: why does this remain ?

warp10/build.gradle
```diff
@@ -80,7 +81,8 @@ dependencies {
     //
     // Pyrolite - need to stick with 4.x
     //
-    api 'net.razorvine:pyrolite:4.10'
+    //api 'net.razorvine:pyrolite:4.10'
```
pi-r-p, line 84: ditto

pi-r-p (changes requested):

warp10/build.gradle
pi-r-p: (the comment)

warp10/build.gradle
```diff
@@ -40,10 +40,10 @@ dependencies {
     api 'org.bouncycastle:bcpg-jdk15on:1.70'
 
     // https://mvnrepository.com/artifact/org.foundationdb/fdb-java
-    api 'org.foundationdb:fdb-java:7.1.25'
+    api 'org.foundationdb:fdb-java:7.2.5'
```
pi-r-p, line 43: It fails with fdb 7.1.26 installed. 
```
Caught exception when initializing FoundationDB API version, please ensure the FoundationDB clients package is installed.
```

Need to fix FDBUtils error message to add the exeption detail in logs and runtime exception, such as:
```
FDBException: This version of the FoundationDB Java binding is not supported by the installed FoundationDB C library. The binding requires a library that supports API version 720, but the installed library supports a maximum version of 710.
```
With latest release installed (7.1.30), it also fails with DEFAULT_FDB_API_VERSION 710.

With latest release installed (7.1.30), it also fails with DEFAULT_FDB_API_VERSION 720.
It also fails with fdb-java:7.2.0

Better to update to fdb-java:7.1.29 , and keep protocol version 710, it seems to early for 7.2.x and 720.

pi-r-p (changes requested):

warp10/build.gradle
pi-r-p: My mistake, there is a 7.2.4 release (without avx). This one works with fdb-java:7.2.5, and protocol 710 and 720.

Following the previous comment, there is no point to keep api version to 710, because fdb-java:7.2.5 force the installation of fdb 7.2.4

So DEFAULT_FDB_API_VERSION can be updated, and 10-fdb.conf.template should be updated and line should be commented, as this is the default.

pi-r-p (changes requested):

pi-r-p (changes requested):
For clarity, you can separate errors in the FDBUtils static block:

```
  static {
    if (!Warp.isStandaloneMode() || Constants.BACKEND_FDB.equals(WarpConfig.getProperty(Configuration.BACKEND))) {
      int version = Integer.parseInt(WarpConfig.getProperty(Configuration.FDB_API_VERSION, DEFAULT_FDB_API_VERSION));
      try {
        FDB.selectAPIVersion(version);
      } catch (UnsatisfiedLinkError t) {
        LOG.error("Unable to initialize FoundationDB API version, please ensure the FoundationDB clients package is installed.");
        throw new RuntimeException("Caught exception when initializing FoundationDB API version, please ensure the FoundationDB clients package is installed.");
      } catch (Throwable t) {
        LOG.error("Unable to initialize FoundationDB API version, " + t.getMessage());
        throw new RuntimeException("Caught exception when initializing FoundationDB API version, " + t.getMessage());
      }
    }
  }
```

In case of mismatch, the message is clearer:

*Caught exception when initializing FoundationDB API version, This version of the FoundationDB Java binding is not supported by the installed FoundationDB C library. The binding requires a library that supports API version 720, but the installed library supports a maximum version of 710.*

PR #1240, opened 2023-04-19
Removed use of finalize which is deprecated in recent JDKs

hbs:


warp10/src/main/java/io/warp10/script/fastdtw/DTW.java
Giwi, line 1: Is it really related to the deprecation of finalize?

hbs: yes

Giwi (approved):

PR #1244, opened 2023-04-25
3x.thrift 0.17.0

hbs:
Bumping Thrift dependency to version 0.17.0. This is not the latest version but it is the last one compiled with a target of JDK8 which is the minimum JDK level we want to support. Any version above 0.17.0 requires JDK 9+.


warp10/src/main/java/io/warp10/script/functions/UNSECURE.java
```diff
@@ -16,6 +16,7 @@
 
 package io.warp10.script.functions;
 
+import io.warp10.ThriftUtils;
```
steven-gueguen, line 19: Please update copyright notice

steven-gueguen (changes requested):

PR #1246, opened 2023-04-25
Added forced sorting of GTS when gskip/gcount is specified.

hbs:
Ensure that GTS results from Directory requests are sorted when gskip/gcount are specified. This is of importance if there are several Directory instances.


steven-gueguen (changes requested):
The StandaloneSplitsHandler class should also be modified

hbs:
> The StandaloneSplitsHandler class should also be modified

The standalone version of the Directory is accessed by a single thread and the results are always sorted, so this is not necessary per se. Will add it anyway in case we make the inner workings evolve so we won't have then bad surprises.

hbs:
PTAL

steven-gueguen (approved):

PR #1248, opened 2023-05-02
update SHRINK behavior

pi-r-p:
Fix `SHRINK` function.
Previously:
- when size was less than requested size, exception
- when size was equal, do nothing (no sort)
- when size was bigger than requested size, sort (or reverse sort), then shrink.

Now:
- when size greater than requested size : sort (or reverse sort) then shrink
- when size is less than requested size : let gts untouched
- can be applied on list of gts.

Documentation updated.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/SHRINK.java
+++ b/warp10/src/main/java/io/warp10/script/functions/SHRINK.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018  SenX S.A.S.
+//   Copyright 2018-2023  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/functions/SHRINK.java
```diff
@@ -18,56 +18,59 @@
...
-      } else {
-        GTSHelper.sort(gts, false);
+
+    return new ElementStackFunction() {
+      @Override
+      public Object applyOnElement(Object element) {
+        if (element instanceof GeoTimeSerie) {
+          GeoTimeSerie gts = (GeoTimeSerie) element;
+
+          if (shrinkto < 0) {
```
Giwi, line 52: do not sort when sizes are the same

pi-r-p: Done, documentation updated.

pi-r-p:
PTAL

PR #1250, opened 2023-05-15
Update of WarpScript parser

pi-r-p:
- when not in multiline, `multiline.toString(); ` previously lead to a npe
- `/* "x": true, */ 42` was previously failing
- tab characters were previously tolerated at the beginning and end of each line (side effect of trim()). Now all control characters are valid separator
- statement number is replaced by position in line, easier for any debugging extension or code audit.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
+++ b/warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
@@ -598,6 +598,8 @@ public class MemoryWarpScriptStack implements WarpScriptStack, Progressable {
           }
           multiline.append(rawline);
         }
+        handleSignal();
+        progress();
         return;
       } else if (inComment.get()) {
         int end = line.indexOf(COMMENT_END);
@@ -624,7 +626,7 @@ public class MemoryWarpScriptStack implements WarpScriptStack, Progressable {
           //
           // Start of comment block /*
           //
-          if (line.length() - pos > 1 && line.charAt(pos) == '/' && line.charAt(pos + 1) == '*') {
+          if (pos < line.length() - 1 && line.charAt(pos) == '/' && line.charAt(pos + 1) == '*') {
             inComment.set(true);
             end = line.indexOf(COMMENT_END, pos + 2); // Look at the end of the comment block on the same line
             if (-1 != end) {
@@ -637,7 +639,7 @@ public class MemoryWarpScriptStack implements WarpScriptStack, Progressable {
           //
           // End of comment block */
           //
-          if (line.length() - pos > 1 && line.charAt(pos) == '*' && line.charAt(pos + 1) == '/') {
+          if (pos < line.length() - 1 && line.charAt(pos) == '*' && line.charAt(pos + 1) == '/') {
             if (!inComment.get()) {
               if (auditMode && !(macros.isEmpty() || macros.size() == forcedMacro)) {
                 WarpScriptAuditStatement err = new WarpScriptAuditStatement(WarpScriptAuditStatement.STATEMENT_TYPE.WS_EXCEPTION, null, "Not inside a comment.", lineNumber, pos, pos + 1);
@@ -694,6 +696,9 @@ public class MemoryWarpScriptStack implements WarpScriptStack, Progressable {
             trimmedLength--;
           }
           trimmedLength++;
+          if (0 == trimmedLength) {
+            break; // empty line
+          }
           if (line.charAt(pos) == '"' || line.charAt(pos) == '\'') {
             char sep = line.charAt(pos);
             boolean warnSepInclusion = false;
--- a/warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
+++ b/warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
@@ -636,7 +636,7 @@ public class MemoryWarpScriptStack implements WarpScriptStack, Progressable {
           if (line.length() - pos > 1 && line.charAt(pos) == '*' && line.charAt(pos + 1) == '/') {
             if (!inComment.get()) {
               if (auditMode && !(macros.isEmpty() || macros.size() == forcedMacro)) {
-                WarpScriptAuditStatement err = new WarpScriptAuditStatement(WarpScriptAuditStatement.STATEMENT_TYPE.WS_EXCEPTION, null, "Not inside a comment.", lineNumber, pos);
+                WarpScriptAuditStatement err = new WarpScriptAuditStatement(WarpScriptAuditStatement.STATEMENT_TYPE.WS_EXCEPTION, null, "Not inside a comment.", lineNumber, pos, pos + 1);
                 macros.get(0).add(err);
                 addAuditError(err);
               } else {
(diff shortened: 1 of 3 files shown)


hbs:
The approach should be modified to simply remove commented code, as it is currently implemented the parsing changes compared with the existing one, which is not acceptable.

hbs:
We should stick with the supported /* ... */ comments of the C standard, see 6.4.9 of https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -564,165 +563,200 @@ public void exec(String line, long lineNumber) throws WarpScriptException {
...
-          // Trim statement
+          // Start of comment block /*
           //
-
-          if (!inMultiline.get()) {
-            stmt = stmt.trim();
+          if (line.length() - pos > 1 && line.charAt(pos) == '/' && line.charAt(pos + 1) == '*') {
+            end = line.indexOf(COMMENT_END, pos + 2); // Look at the end of the comment block on the same line
+            if (-1 != end) {
+              pos = end + 2; // Skip the comment block
```
hbs, line 629: This violates the concatenative language rule which separates statements (which comment end is) by a space. This would allow for 42 /* dadsdad */43 to be valid WarpScript.

So you need to check that the end of comment is either at the end of the line or followed by a whitespace.

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -564,165 +563,200 @@ public void exec(String line, long lineNumber) throws WarpScriptException {
...
           }
 
           //
-          // Trim statement
+          // Start of comment block /*
           //
-
-          if (!inMultiline.get()) {
-            stmt = stmt.trim();
+          if (line.length() - pos > 1 && line.charAt(pos) == '/' && line.charAt(pos + 1) == '*') {
```
hbs, line 626: Comment start MUST either be at the beginning of the line or after a whitespace as the following is not valid WarpScript:

42/* .... */

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -564,165 +563,200 @@ public void exec(String line, long lineNumber) throws WarpScriptException {
...
-      progress();
-
+      
       //
-      // Loop over the statements
+      // Process line character by character, looking at block comments, comments, strings, then process statements.
       //
-
-      for (int st = 0; st < statements.length; st++) {
-        handleSignal();
```
hbs, line 607: You removed the call to handleSignal, it must be re-introduced.

hbs (changes requested):

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
pi-r-p: The way it is done, no problem. in your example, at pos = 0, we see a statement `42/*`, so it will fail as the previous parser will.

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
pi-r-p: I moved it, but it is there. 
I check it at the beginning of the for loop (so, between each statement or comment block).

pi-r-p: my mistake. reintroduced.

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
pi-r-p: done. Also updated syntax highlight in monaco rules.

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -564,165 +563,180 @@ public void exec(String line, long lineNumber) throws WarpScriptException {
...
-          // End execution on encountering a comment
-          //
-
-          if (stmt.length() > 0 && (stmt.charAt(0) == '#' || (stmt.charAt(0) == '/' && stmt.length() >= 2 && stmt.charAt(1) == '/')) && !inMultiline.get() && !inComment.get()) {
-            // Skip comments and blank lines
-            return;
+          // start of comment block /* 
+          if (line.length() - pos > 1 && line.charAt(pos) == '/' && line.charAt(pos + 1) == '*') {
+            // look at the end of the comment block on the same line
+            end = line.indexOf(COMMENT_END, pos + 2);
```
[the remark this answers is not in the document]
pi-r-p: before 2019: */ not inside a comment
after : Unknown function '*/'
now : */ not inside a comment

pi-r-p:
PTAL

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -564,181 +564,251 @@ public void exec(String line, long lineNumber) throws WarpScriptException {
...
           }
 
           //
-          // Trim statement
+          // Start of comment block /*
           //
-
-          if (!inMultiline.get()) {
-            stmt = stmt.trim();
+          if (line.length() - pos > 1 && line.charAt(pos) == '/' && line.charAt(pos + 1) == '*') {
```
hbs, line 627: pos < line.length() - 1 would be clearer than line.length() - pos > 1

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -564,181 +564,251 @@ public void exec(String line, long lineNumber) throws WarpScriptException {
...
           }
+          multiline.setLength(0);
+        } else {
+          // Append current line to existing multiline 
+          if (multiline.length() > 0) {
+            multiline.append("\n");
+          }
+          multiline.append(rawline);
+        }
+        return;
```
hbs, line 601: progress / handleSignal should maybe be called before return statements, otherwise this will delay signal handling when large comment sections and / or multiline are encountered.

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -564,181 +564,251 @@ public void exec(String line, long lineNumber) throws WarpScriptException {
...
           //
-
-          if (stmt.length() > 0 && (stmt.charAt(0) == '#' || (stmt.charAt(0) == '/' && stmt.length() >= 2 && stmt.charAt(1) == '/')) && !inMultiline.get() && !inComment.get()) {
-            // Skip comments and blank lines
-            return;
-          }
-
-          if (WarpScriptStack.MULTILINE_END.equals(stmt)) {
-            if (!inMultiline.get()) {
+          if (line.length() - pos > 1 && line.charAt(pos) == '*' && line.charAt(pos + 1) == '/') {
```
hbs, line 640: pos < line.length() - 1

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -564,181 +564,251 @@ public void exec(String line, long lineNumber) throws WarpScriptException {
...
+          //
+          // Then detect strings, "xx" or 'xx' if the separator is at the end of the line or
+          // followed by a whitespace then we consider we exited the string, otherwise
+          // it is just part of the string, but we store a warning flag.
+          //
+          int trimmedLength = line.length() - 1;
+          while (trimmedLength > 0 && line.charAt(trimmedLength) <= ' ') {
+            trimmedLength--;
+          }
+          trimmedLength++;
```
hbs, line 696: Could early exit if trimmedLength is 0 after the ++

hbs (changes requested):

pi-r-p:
Done

steven-gueguen (approved):

warp10/src/main/java/io/warp10/script/MemoryWarpScriptStack.java
```diff
@@ -564,181 +564,256 @@ public void exec(String line, long lineNumber) throws WarpScriptException {
...
+          }
+          multiline.append(rawline);
+        }
+        handleSignal();
+        progress();
+        return;
+      } else if (inComment.get()) {
+        int end = line.indexOf(COMMENT_END);
+        if (-1 == end) {
+          return; // No end of comment in this line, skip it
```
hbs, line 607: handleSignal / progress

hbs (approved):

PR #1255, opened 2023-05-30
Removed forced version of Guava since HBase is no longer in use

hbs:


pi-r-p (approved):
up to https://github.com/google/guava/wiki/Compatibility , there are no big risks.

steven-gueguen (approved):

PR #1259, opened 2023-06-01
atomic counter for recursion

pi-r-p:
Solve recursion problem with CEVAL


hbs:
This was considered a long time ago, but performance suffers too much from this so it was discarded. Will not approve this PR for that reason.

hbs:
You can modify the substacks so their implementations of recurseIn/recurseOut use an atomic instance. Since those implementations will be used when a CEVAL is called it won't impact the performance of normal executions.

PR #1274, opened 2023-07-05
Removed recursion which could lead to a stack overflow when a lot of GTS have no datapoints in the queried range

hbs:


pi-r-p (approved):

warp10/src/main/java/io/warp10/continuum/store/MultiScanGTSDecoderIterator.java
```diff
@@ -142,187 +142,189 @@ public MultiScanGTSDecoderIterator(boolean fdbUseTenantPrefix, FetchRequest req,
...
-    //
 
-    if (this.scanner.hasNext()) {
-      return true;
-    } else {
-      return hasNext();
+      if (this.scanner.hasNext()) {
+        return true;
+      } else {
+        continue;
```
steven-gueguen, line 326: 'continue' is unnecessary

steven-gueguen (changes requested):
Remove unused import: `import org.bouncycastle.util.encoders.Hex;`

Issue #1276, opened 2023-07-12
Bucketize not working perfectly in latest version


nidgup89:
100  NaN NaN NaN 10 ADDVALUE
100  NaN NaN NaN 10 ADDVALUE
200  NaN NaN NaN  9 ADDVALUE
200  NaN NaN NaN  9 ADDVALUE
300  NaN NaN NaN  8 ADDVALUE
300  NaN NaN NaN  8 ADDVALUE
400  NaN NaN NaN  7 ADDVALUE
400  NaN NaN NaN  7 ADDVALUE
500  NaN NaN NaN  6 ADDVALUE
500  NaN NaN NaN  6 ADDVALUE
600  NaN NaN NaN  5 ADDVALUE
700  NaN NaN NaN  4 ADDVALUE
**800  NaN NaN NaN 42 ADDVALUE
800  NaN NaN NaN 42 ADDVALUE
800  NaN NaN NaN 42 ADDVALUE**
[ SWAP bucketizer.count 0 100 0 ] BUCKETIZE trying to use this getting result -
[{"c":"","l":{},"a":{},"la":0,"v":[[**800,2**],[700,1],[600,1],[500,2],[400,2],[300,2],[200,2],[100,2]]}] but see we have 3 values for 800.

hbs:
Fixed by #1277

PR #1277, opened 2023-07-12
fix issue #1276

pi-r-p:
Fix last bucket timestamp index, in case of duplicate timestamps.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -2338,6 +2338,11 @@ public class GTSHelper {
       } else if (i < 0) {
         // just before the insertion point
         i = -i - 1 - 1;
+      } else {
+        // binary search is not deterministic in case of multiple same timestamps (non dedup gts). Must find the last one.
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -2338,6 +2338,11 @@ public static final GeoTimeSerie bucketize(GeoTimeSerie gts, long bucketspan, in
       } else if (i < 0) {
```
randomboolean, line 2338: also, on line 2231, in the most usual case you can save the call to binary search by using >= instead of >

pi-r-p: indeed for line 2331, but what is the problem with line 2338 ?

randomboolean: The editor didn't allowed commenting on line 2331 so I add the comment line 2338

pi-r-p:
PTAL

randomboolean (approved):

PR #1281, opened 2023-07-31
Added support for returning NULL. Fixed issue with write token decoding.

hbs:


warp10/src/main/java/io/warp10/continuum/MacroFilterAuthenticationPlugin.java
```diff
@@ -103,7 +103,7 @@ public WriteToken extractWriteToken(String token) throws WarpScriptException {
         return null;
       }
```
steven-gueguen, line 105: Apply same check as for read token at line 102
`if (null == MACRO || null == token || "".equals(token)) {`

hbs:
PTAL

PR #1285, opened 2023-08-04
Emptybuckets

randomboolean:
EMPTYBUCKETS function that returns the list of empty buckets of a GTS.

For example, can be used as 'occurences' parameters of MAP.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -4784,6 +4784,69 @@ public class GTSHelper {
   }
 
   /**
+   * Find empty buckets in a bucketized gts
+   * @param gts
+   * @return list of empty buckets
+   */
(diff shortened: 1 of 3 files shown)


randomboolean:
> Why not rather introduce a function BUCKETS which would return a map keyed by bucket end timestamp or "pre" and "post" and whose values would be the number of values in each bucket or boundary? EMPTYBUCKETS could then simply be a post-processing of the output of BUCKETS.

That would not work. EMPTYBUCKETS aim is to outputs bucketend timestamps that are not associated with a value, whereas BUCKETS outputs the number of values in a full bucket

randomboolean:
> If number of value is 0 it means the bucket is empty right?

Not in the sense we want to use this function. We want to obtain bucketend timestamp that are not associated with a value, whether they are empty or not.

hbs:
In that case the name of the function is highly misleading and should not be used for that purpose. It also seems this function is very specific, better implement it as a macro IMHO.

PR #1286, opened 2023-08-09
Added tenant.prefix configs. Added config to ignore tenant prefix issues in data messages.

hbs:


steven-gueguen (commented):
Is `fdb.ignore.tenant.prefix.issues` also needed for StandaloneFDBStoreClient ?

hbs:
> Is `fdb.ignore.tenant.prefix.issues` also needed for StandaloneFDBStoreClient ?

Nope, the issue is simply when using Kafka.

PR #1299, opened 2023-08-25
recursive macro clone, CLONENODUP variation

pi-r-p:
-  Introduce recursive macro cloning, to easily DEREF imbricated macros.
```
0 2 
<%
  'a' STORE 
  <%
    DUP $a > <% $a * %> <% $a + %> IFT 
  %> CLONE SWAP DROP 
  {
    'a' $a
  } DEREF 
%> FOR 
```

-  Introduce `CLONENODUP` (name to discuss) to avoid CLONE SWAP DROP (new name to avoid signature variation up to type like `CONTAINS`)

### DIFF
--- a/warp10/src/main/java/io/warp10/script/WarpScriptLib.java
+++ b/warp10/src/main/java/io/warp10/script/WarpScriptLib.java
@@ -1478,6 +1478,7 @@ public class WarpScriptLib {
   public static final String ATTICK = "ATTICK";
   public static final String ATBUCKET = "ATBUCKET";
   public static final String CLONE = "CLONE";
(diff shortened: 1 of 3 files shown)


randomboolean:
like there is a CLONEREVERSE, maybe it would be better to introduce CLONEDEREF instead of CLONENODUP

pi-r-p:
> like there is a CLONEREVERSE, maybe it would be better to introduce CLONEDEREF instead of CLONENODUP

I'll be happy to have such a function for all other types... `CLONED` is also a possible name (let a cloned object on the stack)

hbs:
Closed because `1 SNAPSHOTN XEVAL` does what was intended by this PR.

PR #1305, opened 2023-09-12
extend ROUND to lists and GTSs

randomboolean:


warp10/src/main/java/io/warp10/script/WarpScriptLib.java
```diff
@@ -2799,7 +2798,7 @@ public class WarpScriptLib {
     addNamedWarpScriptFunction(new NumericalUnaryFunction(SIGNUM, null, Math::signum));
     addNamedWarpScriptFunction(new NumericalUnaryFunction(FLOOR, null, Math::floor));
     addNamedWarpScriptFunction(new NumericalUnaryFunction(CEIL, null, Math::ceil));
-    addNamedWarpScriptFunction(new ROUND(ROUND));
+    addNamedWarpScriptFunction(new NumericalUnaryFunction(ROUND, null, Math::round));
```
steven-gueguen, line 2801: The `ROUND` function will now return a `DOUBLE` instead of a `LONG`

steven-gueguen (changes requested):

steven-gueguen (approved):

pi-r-p (changes requested):
Is it possible to add a shortcut when the gts is already a LONG one ? (juste clone it)

pi-r-p:
(I didn't look at #1306 first, but it still applies)

randomboolean:
> Is it possible to add a shortcut when the gts is already a LONG one ? (juste clone it)

PTAL

pi-r-p (approved):

PR #1306, opened 2023-09-13
Optimize numerical functions

randomboolean:
depends on PR #1305

Refactor instanciation of GTS operators when related warpscript functions are registered.
Previously, they were instantiated for each GTS at execution time.


pi-r-p (changes requested):
add a shortcut when double to long and already long, for public NumericalUnaryFunction(String name, DoubleToLongFunction opDL)  (this is only used for ROUND)

randomboolean:
> add a shortcut when double to long and already long, for public NumericalUnaryFunction(String name, DoubleToLongFunction opDL) (this is only used for ROUND)

PTAL

pi-r-p (changes requested):
You cannot clone() at this level. For example ABS takes long and outputs long... It won't work with this approach. ROUND is the only long->long example where clone() can be used safely.

randomboolean:
> You cannot clone() at this level. For example ABS takes long and outputs long... It won't work with this approach. ROUND is the only long->long example where clone() can be used safely.

I cannot reproduce this bug.
This test `NEWGTS 1 10 <% NaN NaN NaN -1 ADDVALUE %> FOR ABS` outputs the correct new GTS object that is not a clone.

warp10/src/main/java/io/warp10/script/functions/NumericalUnaryFunction.java
```diff
@@ -61,32 +94,34 @@ public Object applyOnElement(Object element) throws WarpScriptException {
...
+          // Choose the operator depending on which ones are defined and the GTS type.
           if (null != opD && (null == opL || gts.getType() == GeoTimeSerie.TYPE.DOUBLE)) {
             // Consider all values as doubles because only the double operator is defined or the GTS is of type DOUBLE.
-            op = new GTSOpsHelper.GTSUnaryOp() {
-              @Override
-              public Object op(GeoTimeSerie gts, int idx) {
-                return opD.applyAsDouble(((Number) GTSHelper.valueAtIndex(gts, idx)).doubleValue());
-              }
-            };
+            if (LongUnaryOperator.identity() == opD) { // only case where GTSopD is null and opD is not
```
pi-r-p, line 102: you mean DoubleUnaryOperator.identity() for opD ?

warp10/src/main/java/io/warp10/script/functions/NumericalUnaryFunction.java
```diff
@@ -38,16 +39,48 @@
...
+      throw new RuntimeException("Incoherent instantiation parameters for numerical function: " + name);
+    }
+
+    GTSopL = null == opL || LongUnaryOperator.identity() == opL ? null : new GTSOpsHelper.GTSUnaryOp() {
+      @Override
+      public Object op(GeoTimeSerie gts, int idx) {
+        return opL.applyAsLong(((Number) GTSHelper.valueAtIndex(gts, idx)).longValue());
+      }
+    };
+    GTSopD = null == opD || LongUnaryOperator.identity() == opD ? null : new GTSOpsHelper.GTSUnaryOp() {
```
pi-r-p, line 63: you mean DoubleUnaryOperator.identity() for opD ?

pi-r-p (changes requested):

pi-r-p:
PTAL

pi-r-p (approved):

PR #1316, opened 2023-10-02
Refactored instatiation of TCompactProtocol.Factory so we can set the  max message size which was limited to 100Mb since 0.14

hbs:
This fixes an issue which arises when attempting to deserialize wrappers over 100Mb


warp10/src/main/java/io/warp10/script/ScriptRunner.java
Giwi, line 1: You have to update the copyright

Giwi (approved):

steven-gueguen (approved):

PR #1319, opened 2023-10-09
update CALL example to python3

pi-r-p:


etc/calls/testcall.py
```diff
@@ -1,19 +1,20 @@
-#!/usr/bin/python -u 
+#!/usr/bin/env -S python3 -u 
 
-import cPickle
+import pickle
```
hbs, line 3: The version of the pyrolite library we use only supports the Pyro4 protocol, is that ok with that import change?

hbs: Why import pickle and not _pickle which is the C based version, way faster than pickle?

etc/calls/testcall.py
```diff
@@ -1,19 +1,20 @@
...
 # The absolute maximum is set in the configuration file via 'warpscript.call.maxcapacity'
 #
 
-print 10
+print(1)
 
 #
-# An example of Warpscript to trigger this script: NOW ISO8601 ->PICKLE ->B64URL 'testcall.py' CALL
+# An example of Warpscript to test this script: 
+# [ NOW ISO8601 ] ->PICKLE ->B64URL 'testcall.py' CALL B64-> PICKLE->
```
hbs, line 17: Why change the example?

etc/calls/testcall.py
```diff
@@ -29,24 +30,26 @@
     #
     line = sys.stdin.readline()
     line = line.strip()
-    line = urllib.unquote(line.decode('utf-8'))
     # Remove Base64 encoding
-    mystr = base64.b64decode(line)
-    args = cPickle.loads(mystr)
+    pickledB64Arguments = base64.b64decode(urllib.parse.unquote(line))
```
hbs, line 34: Change the name, it is confusing to use pickledB64Arguments when the value has nothing to do with B64

hbs (changes requested):

etc/calls/testcall.py
pi-r-p: The previous example did not show a fully pickle unpickle that manipulates LIST and MAP.

PR #1320, opened 2023-10-10
Initial commit of BigDecimal support

hbs:


randomboolean:
Do you need to implement specific functions for big decimals rather than implementing support for them in existing ones?

hbs:
Yes this is intentional as otherwise having BigDecimal which are not explicit may lead to issues across Warp 10 or its extensions and plugins.

steven-gueguen (changes requested):

hbs:
PTAL

steven-gueguen (approved):

PR #1322, opened 2023-10-11
Backend.archive

hbs:


warp10/src/main/sh/warp10.sh
```diff
@@ -269,6 +269,20 @@ backend = leveldb" >>"${WARP10_CONFIG_DIR}/99-init.conf"
   postInit
 }
 
+leveldbConf() {
```
steven-gueguen, line 272: Replace with `archiveConf`

steven-gueguen (changes requested):

warp10/src/main/sh/warp10.sh
hbs: PTAL

steven-gueguen (approved):

PR #1323, opened 2023-10-17
Introduce support for Argon2

hbs:


hbs:
PTAL

crypto/src/main/java/io/warp10/crypto/OSSKeyStore.java
```diff
@@ -91,17 +87,14 @@ public byte[] decodeKey(String encoded) {
...
+
+    if (encoded.startsWith("wrapped:")) {
       if (null == this.masterKey) {
         throw new RuntimeException("Master Key not retrieved from OSS, aborting.");
       }
-      return CryptoHelper.unwrapBlob(this.masterKey, decodeKey(encoded.substring(8)));
+      return CryptoHelper.unwrapBlob(this.masterKey, CryptoUtils.decodeKey(this, encoded.substring(8)));
     } else {
-      return null;
+      return CryptoUtils.decodeSimpleKey(encoded);
```
steven-gueguen, line 97: This function does not handle argon2. Is this intentional?

hbs: It was with the original signatures. Updated to support argon2.

PR #1329, opened 2023-11-10
Release artifacts

steven-gueguen:
Add Warp 10 archive signature and checksum
Ask to publish only if head is a tag

### DIFF
--- a/warp10/src/main/sh/package.sh
+++ b/warp10/src/main/sh/package.sh
@@ -113,7 +113,14 @@ find "${WARP10_HOME}" -type f \( -name \*.sh -o -name \*.py -o -name \*.init \) 
 chmod 700 "${WARP10_HOME}/etc"
 
 # Build tar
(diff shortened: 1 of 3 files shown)


warp10/src/main/sh/package.sh
```diff
@@ -113,7 +113,14 @@ find "${WARP10_HOME}" -type f \( -name \*.sh -o -name \*.py -o -name \*.init \)
 chmod 700 "${WARP10_HOME}/etc"
 
 # Build tar
-tar czpf "../build/libs/warp10-${VERSION}.tar.gz" "${WARP10_HOME}"
+FILE=../build/libs/warp10-${VERSION}.tar.gz
+tar czpf "${FILE}" "${WARP10_HOME}"
+
+# Generate signature
+gpg --local-user BD49DA0A --output "${FILE}".asc --detach-sign "${FILE}"
```
hbs, line 120: The key id should be a parameter

steven-gueguen: PTAL

PR #1332, opened 2023-11-16
Bigdecimal easier

pi-r-p:
Few fixes for easier use of BigDecimal functions.

Documentation is ready, but depends on this PR


steven-gueguen (changes requested):
You should also invert the parameters for BDCOMP

steven-gueguen (approved):

PR #1338, opened 2023-12-22
Use LinkedHashMap when applicable

steven-gueguen:
Avoid error when using MSORT:
MSORT operates on a map created by {} or ->MAP.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/ext/stackps/WSINFO.java
+++ b/warp10/src/main/java/io/warp10/script/ext/stackps/WSINFO.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2020-2023  SenX S.A.S.
+//   Copyright 2020-2024  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -17,7 +17,6 @@
 package io.warp10.script.ext.stackps;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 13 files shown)


warp10/src/main/java/io/warp10/script/ext/stackps/WSINFO.java
```diff
@@ -28,17 +29,9 @@
...
-      WarpScriptStack.ATTRIBUTE_FETCH_COUNT,
-      WarpScriptStack.ATTRIBUTE_GTS_COUNT,
-      WarpScriptStack.ATTRIBUTE_MACRO_NAME,
-      WarpScriptStack.ATTRIBUTE_NAME,
-      WarpScriptStack.ATTRIBUTE_SECTION_NAME,
-      StackPSWarpScriptExtension.ATTRIBUTE_SESSION,
-  };
-  
+
+  public static final String[] EXPOSED_ATTRIBUTES = new String[]{WarpScriptStack.ATTRIBUTE_CREATION_TIME, WarpScriptStack.ATTRIBUTE_FETCH_COUNT, WarpScriptStack.ATTRIBUTE_GTS_COUNT, WarpScriptStack.ATTRIBUTE_MACRO_NAME, WarpScriptStack.ATTRIBUTE_NAME, WarpScriptStack.ATTRIBUTE_SECTION_NAME, StackPSWarpScriptExtension.ATTRIBUTE_SESSION,};
```
hbs, line 33: This is far harder to read than the original version, is that an IDE proposed change?

hbs (changes requested):

warp10/src/main/java/io/warp10/script/ext/stackps/WSINFO.java
steven-gueguen: Indeed, wrong auto formatting parameters. reverted

hbs (approved):

PR #1341, opened 2024-02-05
fix SHRINK with MINLONG MAXLONG value

pi-r-p:
fixes MINLONG SHRINK.

### DIFF
--- a/warp10/src/main/java/io/warp10/script/functions/SHRINK.java
+++ b/warp10/src/main/java/io/warp10/script/functions/SHRINK.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2023  SenX S.A.S.
+//   Copyright 2018-2024  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


warp10/src/main/java/io/warp10/script/functions/SHRINK.java
```diff
@@ -49,13 +49,15 @@ public Object applyOnElement(Object element) {
         if (element instanceof GeoTimeSerie) {
           GeoTimeSerie gts = (GeoTimeSerie) element;
 
-          if (GTSHelper.nvalues(gts) > Math.abs(shrinkto)) {
-            if (shrinkto < 0) {
-              GTSHelper.sort(gts, true);
-            } else {
-              GTSHelper.sort(gts, false);
+          if (shrinkto != Long.MIN_VALUE && shrinkto != Long.MAX_VALUE) {
```
randomboolean, line 52: don't need to check that shrinkto != Long.Max_Value

warp10/src/main/java/io/warp10/script/functions/SHRINK.java
```diff
@@ -49,13 +49,15 @@ public Object applyOnElement(Object element) {
...
+            if (GTSHelper.nvalues(gts) > Math.abs(shrinkto)) {
```
randomboolean, line 53: this check should come first since since it is 99.999...% of the usages
merge this check with the former line with &&

warp10/src/main/java/io/warp10/script/functions/SHRINK.java
pi-r-p: It saves a cycle. There is no need to call shrink in this case.

pi-r-p:
PTAL

warp10/src/main/java/io/warp10/script/functions/SHRINK.java
randomboolean: Then you can also save a cycle by not calling shrink when shrinkto is equal to Min_Value + 1 (since it is equal to abs(MAX_VALUE)). Modify the condition accordingly.

pi-r-p: Code updated with Integer limits, for a safe cast line 58.

pi-r-p:
PTAL

randomboolean (approved):

PR #1344, opened 2024-02-12
Added error message and early JVM exit when using JDK 17 without labe…

hbs:
…lsid.slowimpl set

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2023  SenX S.A.S.
+//   Copyright 2018-2024  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -65,6 +65,7 @@ import com.geoxp.GeoXPLib.GeoXPShape;
 
 import io.warp10.CapacityExtractorOutputStream;
 import io.warp10.ThriftUtils;
+import io.warp10.WarpConfig;
 import io.warp10.WarpHexDecoder;
 import io.warp10.WarpURLDecoder;
 import io.warp10.WarpURLEncoder;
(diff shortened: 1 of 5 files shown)


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -107,7 +108,14 @@ public class GTSHelper {
   private static final boolean labelsIdSlowImpl;
 
   static {
-    labelsIdSlowImpl = "true".equals(System.getProperty(LABELSID_SLOWIMPL));
+    if ("17".equals(System.getProperty("java.specification.version")) && null == WarpConfig.getProperty(LABELSID_SLOWIMPL)) {
```
pi-r-p, line 111: Set labelsIdSlowImpl variable first, then test if jdk==17 and labelsIdSlowImpl == false.

Some user may set/keep labelsIdSlowImpl = false in their configuration.

pi-r-p (changes requested):

warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
hbs: That's exactly what the code is doing, it only prints the warning if the labelsid.slowimpl key is not set AND the JDK is 17.

As there is no default value set for labelsid.slowimpl in the configuration the error WILL be displayed when using JDK17 unless the configuration key has been set explicitely.

pi-r-p: OK, this was intentional.

pi-r-p (approved):

stggn (approved):

warp10/src/test/java/io/warp10/continuum/gts/SubSerieTest.java
```diff
@@ -16,13 +16,27 @@
 
 package io.warp10.continuum.gts;
 
+import java.io.StringReader;
```
stggn, line 19: Please update the copyright date

stggn (changes requested):

PR #1352, opened 2024-03-06
fix(ingress): prevent null pointer on delete

alexandrebrg:
Delete fails with null pointer exception if `selector` query param is not set.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
+++ b/warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
@@ -1622,8 +1622,8 @@ public class Ingress extends AbstractHandler implements Runnable {
 
       String selector = request.getParameter(Constants.HTTP_PARAM_SELECTOR);
 
-    if (null == selector) {
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid '" + Constants.HTTP_PARAM_SELECTOR +"' query parameter, should not be empty.");
+    if (null == selector || selector.isEmpty) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid '" + Constants.HTTP_PARAM_SELECTOR +"' query parameter, should not be missing or empty.");
       return;
     }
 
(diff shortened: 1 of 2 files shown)


warp10/src/main/java/io/warp10/continuum/ingress/Ingress.java
```diff
@@ -1622,6 +1622,10 @@ public void handleDelete(String target, Request baseRequest, HttpServletRequest
 
       String selector = request.getParameter(Constants.HTTP_PARAM_SELECTOR);
 
+      if (null == selector) {
+          throw new IOException("Invalid selector query parameter, should not be empty.");
```
hbs, line 1626: Could you rather use resp.sendError and use the constant defined in class `Constants` for the parameter name.

Also ensure the same test is performed in the standalone delete endpoint for coherency.

alexandrebrg: Sure, I made the changes.

hbs:
Thanks, your PR has been assigned reviewers and therefore entered the weekly review cycle. FYI weekly PR review happens on Wednesdays, so expect a merge next Wednesday if nothing has been noted by reviewers.

warp10/src/main/java/io/warp10/standalone/StandaloneDeleteHandler.java
```diff
@@ -214,6 +214,11 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
 
     String selector = request.getParameter(Constants.HTTP_PARAM_SELECTOR);
 
+    if (null == selector) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid '" + Constants.HTTP_PARAM_SELECTOR +"' query parameter, should not be empty.");
```
pi-r-p, line 218: Isn't missing, instead of empty ?.
Or if you change the condition to null == selector || selector.isEmpty() , you can return "missing or empty selector query parameter".

pi-r-p (changes requested):

warp10/src/main/java/io/warp10/standalone/StandaloneDeleteHandler.java
alexandrebrg: It makes sense to handle both case as this field is needed to be fullfilled. Pushed the changes.

pi-r-p (approved):

stggn (changes requested):
Please update the copyright date in `StandaloneDeleteHandler.java` (e.g. 2018-202**4**)

alexandrebrg:
> Please update the copyright date in `StandaloneDeleteHandler.java` (e.g. 2018-202**4**)

I updated the copyrights, should be good now :)

stggn (approved):

PR #1353, opened 2024-03-08
Revamping FILL with univariate support and new fillers

randomboolean:
### New Functionalities

The proposed enhancements extend the FILL operation to support univariate operations, filling missing timestamps within individual GTS based on specified gaps or empty buckets.

This approach fills existing functionality gaps: 
- the inability of bucketize fillers to utilize filler functions or to fill beyond empty buckets.
- the inability of existing filler functions and FILL framework to be used beyond cross filling use cases.

### Function Signatures

- **Existing Signature**: Maintains the original functionality, aligning two GTS by filling to match their timestamps.

```
gtsa gtsb filler FILL
```

- **New Signatures**: Introduce univariate filling, with optional parameter for specific gaps (`ticks`) or empty buckets.

```
[ gtsList filler ] FILL

[ gtsList filler ticks ] FILL

gtsList {
  'filler' filler
  // optional:
  'ticks' ticks
  'verify' True
  'invalid.value' NULL
} FILL
```

Note that:
- the optional parameter `invalid.value` is the value returned by the evaluation function if the input timestamp is not in the function's support,
- the optional parameter `verify` indicates if FILL verifies for each tick of the list `ticks` if it is a gap before filling it (else the result will contain duplicate ticks).

### Additional Fillers

- filler.linear
- filler.spline
- filler.akima
- filler.newton
- filler.lowess
- filler.rlowess

### MACROFILLER

When used in the context of the new univariate FILL mode, the macro encapsulated by MACROFILLER should be presented only with elements from the specified GTS, unlike in cross fill operations where it also receives elements from another GTS. The documentation will be revised to reflect this usage.

### Macros as Filler Parameter

We have introduced the capability to use a macro directly as a filler parameter, (similar to a lambda filler function). In this setup, the macro must be designed to expect a GTS and a tick, and to return the filler value.


warp10/src/main/java/io/warp10/continuum/gts/INTERPOLATE.java
```diff
@@ -27,23 +25,514 @@
...
+        if ((null != filled.locations) && (GeoTimeSerie.NO_LOCATION != filled.locations[i])) {
+          nLocations++;
+        }
+
+        //
+        // Move bucket passed the last tick encountered
+        //
+        while(bucket < filled.lastbucket && bucket <= filled.ticks[i-1]) {
+          bucket += filled.bucketspan;
+        }
```
pi-r-p, line 264: It cannot be the same as INTERPOLATE:

```
//# @endpoint https://sandbox.senx.io/api/v0/exec
NEWGTS
[ 0   -50.8  -122.0  42   100 ] ADDVALUE
[ 8   -50.2  -121.0  42   50 ] ADDVALUE
[ 51   -50.5  -120.5  42   50 ] ADDVALUE
[ 60   -50.9  -119.0  42   80 ] ADDVALUE
[ 80   -50.2  -120.0  42   0 ] ADDVALUE
[ 80   -50.2  -120.0  42   -10 ] ADDVALUE
[ 100   -49.5  -118.5  42   0 ] ADDVALUE
'inputData' RENAME 
'inputGts' STORE 

[ $inputGts bucketizer.last 200 10 30 ] BUCKETIZE
{
  'interpolator' 'akima'
  'interpolator.geo' 'spline'
  'invalid.tick.value' -Infinity
} INTERPOLATE  

```

Here you expect to fill with MINLONG all the empty buckets before and after the first and last tick.

warp10/src/main/java/io/warp10/continuum/gts/INTERPOLATE.java
```diff
@@ -27,23 +25,514 @@
...
+        //
+        if (bucket == filled.ticks[i]) {
+          continue;
+        }
+
+        //
+        // Fill missing values until bucket passes the current tick
+        //
+
+        while(bucket < filled.ticks[i]) {
```
pi-r-p, line 277: ditto, just loop on every buckets. There can be a shortcut only if "invalid.tick.value" is not defined.

warp10/src/main/java/io/warp10/continuum/gts/INTERPOLATE.java
```diff
@@ -27,23 +25,514 @@
...
+      for (int i = 0; i < nvalues; i++) {
+
+        if ((null != filled.elevations) && (GeoTimeSerie.NO_ELEVATION != filled.elevations[i])) {
+          nElevations++;
+        }
+
+        if ((null != filled.locations) && (GeoTimeSerie.NO_LOCATION != filled.locations[i])) {
+          nLocations++;
+        }
+      }
```
pi-r-p, line 301: If my understanding is correct, this can be done once before the occurences/bucket choice. From the result, you can build interpolation functions for location and elevation (or leave a null), then you can iterate just once on all occurrences/bucket.

warp10/src/main/java/io/warp10/continuum/gts/INTERPOLATE.java
```diff
@@ -27,23 +25,514 @@
...
+    }
+
+    //
+    // Fill elevation
+    //
+
+    if (nElevations >= 2) {
+      if (null == params.get(PARAM_INTERPOLATOR_GEO)) {
```
pi-r-p, line 369: I don't know if this is usefull, but you can add a new `interpolator.elevation` parameter (if you want linear interpolation for location, and another one for elevation). maybe `interpolator.geo` can become `interpolator.location`

warp10/src/main/java/io/warp10/continuum/gts/INTERPOLATE.java
```diff
@@ -27,23 +25,514 @@
...
+            break;
+        }
+      }
+
+      //
+      // Sort ticks
+      //
+
+      GTSHelper.sort(filled);
+
```
pi-r-p, line 397: This is could be nice, but is not in the logic of original `INTERPOLATE` function. It adds all the new datapoints after the existing ones.

pi-r-p (changes requested):
To discuss: deduplication issues. 
It can just fail, or deduplicate silently:
- with the full DEDUP logic ? 
- with a SORT ?
Failing is simpler, and dedups are easy to check after sort.

warp10/src/main/java/io/warp10/continuum/gts/INTERPOLATE.java
randomboolean: this is done only if some geo interpolation is perform (like for the original implementation)

warp10/src/main/java/io/warp10/continuum/gts/INTERPOLATE.java
```diff
@@ -27,23 +25,514 @@
...
+    //
+    // Extract initial number of values
+    //
+
+    int nvalues = filled.values;
+
+    //
+    // Compute interpolator
+    //
+
```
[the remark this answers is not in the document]
randomboolean: Is it needed ? In the original implementation an exception is raised when trying to cast. I think we can keep the same behavior.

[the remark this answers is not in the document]
pi-r-p: Clearer error is always a good thing for users. (you can check type very early, before calling the original method)

warp10/src/main/java/io/warp10/continuum/gts/INTERPOLATE.java
```diff
@@ -27,23 +25,514 @@
...
+    double xval[] = new double[nvalues];
+    double fval[] = new double[nvalues];
+    for (int i = 0; i < nvalues; i++) {
+      xval[i] = ((Number) GTSHelper.tickAtIndex(filled, i)).doubleValue();
+      fval[i] = ((Number) GTSHelper.valueAtIndex(filled, i)).doubleValue();
+    }
```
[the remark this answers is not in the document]
randomboolean: error message is improved

warp10/src/main/java/io/warp10/continuum/gts/INTERPOLATE.java
```diff
@@ -27,23 +25,514 @@
...
+        case linear:
+          if (nvalues > 1) {
+            function = (new LinearInterpolator()).interpolate(xval, fval);
+            break;
+          }
+        case noop:
+          function = null;
+          break;
+      }
+    }
```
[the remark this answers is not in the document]
randomboolean: refactored, (also prepare to accept interpolation function that are not polynomial splines), and fix noop bug

pi-r-p:
There is still a question around bucketize out of range ticks:

```
//# @endpoint https://sandbox.senx.io/api/v0/exec
NEWGTS
[ 0   -50.8  -122.0  42   100 ] ADDVALUE
[ 8   -50.2  -121.0  42   50 ] ADDVALUE
[ 51   -50.5  -120.5  42   50 ] ADDVALUE
[ 60   -50.9  -119.0  42   80 ] ADDVALUE
[ 80   -50.2  -120.0  42   0 ] ADDVALUE
[ 80   -50.2  -120.0  42   -10 ] ADDVALUE
[ 100   -49.5  -118.5  42   0 ] ADDVALUE
'inputData' RENAME 
'inputGts' STORE 

[ $inputGts bucketizer.last 200 10 30 ] BUCKETIZE
{
  'interpolator' 'akima'
  'interpolator.loc' 'spline'
  'invalid.tick.val' -Infinity
} INTERPOLATE  

```

Should be fill buckets before 0 and after 100 with the fill value as FILLVALUE does ?

pi-r-p (changes requested):
using a macro as filler works great.

but using MACROFILLER leads to unexpected errors.

```
NEWGTS
[ 0 100 ] ADDVALUE
[ 8 50 ] ADDVALUE
[ 51 50 ] ADDVALUE
[ 60 80 ] ADDVALUE
[ 80 0 ] ADDVALUE
[ 100 0 ] ADDVALUE
'inputData' RENAME 
[ SWAP NULL 0 10 20 ] BUCKETIZE
'inputGts' STORE 

//$inputGts



  $inputGts
{
  'filler' <% 
    [ 'gts' 'tick' ] STORE 
    42 
  %>  0 0 MACROFILLER
  'ticks'  [ -10 0 10 20 30 40 45 60 70 80 90 100 110 ]
} FILL SORT

```

(with or without ticks)

will lead to Cannot invoke "java.lang.Number.longValue()" because the return value of "java.util.List.get(int)" is null

randomboolean:
> using a macro as filler works great.
> 
> but using MACROFILLER leads to unexpected errors.
> 
> ```
> NEWGTS
> [ 0 100 ] ADDVALUE
> [ 8 50 ] ADDVALUE
> [ 51 50 ] ADDVALUE
> [ 60 80 ] ADDVALUE
> [ 80 0 ] ADDVALUE
> [ 100 0 ] ADDVALUE
> 'inputData' RENAME 
> [ SWAP NULL 0 10 20 ] BUCKETIZE
> 'inputGts' STORE 
> 
> //$inputGts
> 
> 
> 
>   $inputGts
> {
>   'filler' <% 
>     [ 'gts' 'tick' ] STORE 
>     42 
>   %>  0 0 MACROFILLER
>   'ticks'  [ -10 0 10 20 30 40 45 60 70 80 90 100 110 ]
> } FILL SORT
> ```
> 
> (with or without ticks)
> 
> will lead to Cannot invoke "java.lang.Number.longValue()" because the return value of "java.util.List.get(int)" is null

[01dfe14](https://github.com/senx/warp10-platform/pull/1353/commits/01dfe144f61eb240dc332c73f2f0e39f573ec8d9) fixes bug in original code of MACROFILLER and improve error message.

Note that MACROFILLER macro expects the same input arguments as in the original implementation (which differs from input arguments of a macro as a filler).

warp10/src/main/java/io/warp10/script/filler/FillerLowess.java
```diff
@@ -0,0 +1,127 @@
...
+      double a = LoessInterpolator.DEFAULT_ACCURACY;
+      Object o = stack.pop();
+      if (o instanceof Double) {
+        a = (double) o;
+        o = stack.pop();
+      }
+
+      // bandwidth
+      if (!(o instanceof Long)) {
+        throw new WarpScriptException(getName() + " expects an bandwidth as a number of data points (LONG), but instead got a " + TYPEOF.typeof(o));
```
pi-r-p, line 70: typo (an)

warp10/src/main/java/io/warp10/script/filler/FillerRlowess.java
```diff
@@ -0,0 +1,140 @@
...
+      // robustness iterations
+      if (!(o instanceof Long)) {
+        throw new WarpScriptException(getName() + " expects a number of robustifying iterations (LONG), but instead got a " + TYPEOF.typeof(o));
+      }
+      int r = ((Long) o).intValue();
+
+      // bandwidth
+      o = stack.pop();
+      if (!(o instanceof Long)) {
+        throw new WarpScriptException(getName() + " expects an bandwidth as a number of data points (LONG), but instead got a " + TYPEOF.typeof(o));
```
pi-r-p, line 82: typo (an)

warp10/src/main/java/io/warp10/script/functions/FILLTICKS.java
```diff
@@ -34,8 +33,11 @@
  *
  * FILLTICKS expects the following parameters on the stack:
  * 1: a list of five parameters: lat, lon, elevation, value, ticks
- *    where ticks is a list of ticks for which to add the data point (tick, lat, lon, elevation and value) * 
+ *    where ticks is a list of ticks for which to add the data point (tick, lat, lon, elevation and value)
+ *    
+ * This function is deprecated. Use instead 'ticks' parameter of FILL with filler.value
```
pi-r-p, line 38: ... "with 'verify' parameter set to false, to keep the same duplicate behavior"

pi-r-p (changes requested):
On a bucketized gts, all the buckets in the future are discarded, even by filler.value.

test code:

```
NEWGTS
[ 0   -50.8  -122.0  42   100 ] ADDVALUE
[ 8   -50.2  -121.0  42   50 ] ADDVALUE
[ 51   -50.5  -120.5  42   50 ] ADDVALUE
[ 60   -50.9  -119.0  42   80 ] ADDVALUE
[ 80   -50.2  -120.0  42   -10 ] ADDVALUE
[ 100   -49.5  -118.5  42   0 ] ADDVALUE
'inputData' RENAME 
'inputGts' STORE 

[
"filler.interpolate"
"filler.previous"
"filler.akima"
"filler.next"
"5 5 filler.rlowess"
"filler.linear"
"filler.trend"
"[ -48.0 -125.0 8 -100 ] filler.value"
"filler.newton"
"filler.spline"
"5 filler.lowess"
] 'fillerList' STORE
// new way

$fillerList
<%
  'filler' STORE

  [ $inputGts bucketizer.last 130 10 15 ] BUCKETIZE 0 GET 
  {
    "filler"  $filler EVAL
    //"ticks"   
    //"verify"   false // force to keep ticks duplicates.
    "invalid.value" -50
  }
  FILL 0 GET SORT 
  '+ ' $filler + RENAME


%> FOREACH
```

pi-r-p:
More remarks:
- filler.interpolate and filler.linear produce the same output, but not for geo (filler.interpolate does interpolation on geo)
You may remove filler.linear ?

- When processing a bucketized GTS with the `ticks` option, the output is bucketized. It should not be. (there could be non aligned ticks, if we ask for them in `ticks` list.

- filler.trend is obviously broken somewhere. (but it predates this PR)

randomboolean:
For your first point, maybe we should instead deprecated filler.interpolate ?
IMO, for this filler, I think it is better not to impose a linear interpolation of geo coordinates.
It is different for filler.next or filler.prev where it makes sense to copy the geo coordinates.

About the second point, I pushed a fix

pi-r-p:
My last remark is functional.
The "ticks" parameter adds ticks to FILL, but the output contains the new filled ticks and the other ones that were already in the input. 

```
// @preview gts
[
  NEWGTS
  [ -21   -50.8  -122.0  42   90 ] ADDVALUE
  [ 0   -50.8  -122.0  42   100 ] ADDVALUE
  [ 8   -50.2  -121.0  42   50 ] ADDVALUE
  [ 51   -50.5  -120.5  42   50 ] ADDVALUE
  [ 60   -50.9  -119.0  42   80 ] ADDVALUE
  [ 80   -50.2  -120.0  42   -10 ] ADDVALUE
  [ 100   -49.5  -118.5  42   0 ] ADDVALUE 'gts a' RENAME 
  [ 0 10 20 30 40 ] [] [] [] [ 10 -11 12 -13 -12 ] MAKEGTS 'gts b' RENAME
  [ 0 10 30 50 60 ] [] [] [] [ 80 88 78 48  ] MAKEGTS 'gts c' RENAME
] 'inputGtsList' STORE

[ $inputGtsList bucketizer.last 130 5 32 ] BUCKETIZE '+.bucketized' RENAME 
'bucketizedInputs' STORE

$bucketizedInputs
{
  "filler"  filler.spline // the spline cannot be computed in the futur or in the past
  "ticks" [ 11 12 13 14 15 16 ]  // overrides the bucket list
}
FILL SORT 

```

It differs from the "ticks" in MAP framework, where the output can only contains ticks in the "ticks" list.

In order to fill only a few datapoints, you need to FILL with "ticks" parameter, then use a trick to remove the extra ones.

Both behavior are interesting... So it could be nice to expose both with a new parameter, for example "ticks.keep.all", maybe false by default to be in line with MAP behavior.

randomboolean:
But we clearly don't want to be in line with MAP behavior since we fill gaps. We do want to keep original ticks every time.

pi-r-p:
There is a several use cases where you do not want the input anymore: On a list of energy counter gts, get index every end of day, with a linear interpolation. Perfect for this new fill:
In this case, we want to force the ticks (a list of every midnight tick), and ignore all the other input datapoints.

So, having an extra setting "ticks.only" true, for example, will save time. The alternative is to build a GTS with the computed ticks, and call COMMONTICKS after. Not easy for a beginner.

randomboolean:
The function's name FILL is not suitable for the use case you are describing.

pi-r-p (approved):

pi-r-p:
One remark: TYPEOF output is not clean for these new fillers.

PR #1358, opened 2024-05-15
Missing labels fix

hbs:


warp10/src/main/java/io/warp10/continuum/gts/GTSHelper.java
```diff
@@ -8954,7 +8954,7 @@ public static String buildSelector(Metadata metadata, boolean forSearch) {
           sb.append(",");
         }
         encodeName(sb, entry.getKey());
-        if (forSearch && Constants.ABSENT_LABEL_SUPPORT && "".equals(entry.getValue())) {
+        if (forSearch && !Constants.ABSENT_LABEL_SUPPORT && "".equals(entry.getValue())) {
```
pi-r-p, line 8957: I understand the change, but in which situation do we enter this code ?  There is never an empty string as label value.

pi-r-p (approved):

PR #1359, opened 2024-05-15
Issue 1356

hbs:


warp10/src/main/java/io/warp10/script/functions/RELABEL.java
```diff
@@ -47,7 +48,14 @@ public ElementStackFunction generateFunction(WarpScriptStack stack) throws WarpS
...
 
-    final Map<String, String> labels = (Map<String, String>) top;
+    final Map<String, String> labels = new LinkedHashMap<String,String>(((Map<Object, Object>) top).size());
+
+    for (Entry<Object,Object> entry: ((Map<Object,Object>) top).entrySet()) {
+      if (!(entry.getKey() instanceof String) || !(entry.getValue() instanceof String)) {
+        throw new WarpScriptException(getName() + " keys and values MUST be STRING");
+      }
+      labels.put((String) entry.getKey(), (String) entry.getValue());
+    }
```
pi-r-p, line 58: - value can be NULL to remove a single label
- key AND value can be NULL to remove all labels.

pi-r-p (changes requested):

pi-r-p (approved):

PR #1361, opened 2024-05-31
fix npe when selector or token have empty labels

pi-r-p:
for example, ".owner" key can have a null value.

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/gts/MetadataSelectorMatcher.java
+++ b/warp10/src/main/java/io/warp10/continuum/gts/MetadataSelectorMatcher.java
@@ -278,21 +278,23 @@ public class MetadataSelectorMatcher {
         Map<String,String> labels = dr.getLabelsSelectors().get(i);
         boolean first = true;
         for (Entry<String,String> entry: labels.entrySet()) {
-          if (!first) {
-            sb.append(",");
+          if (entry.getValue() != null) { // it could be null (.owner, for example)
+            if (!first) {
+              sb.append(",");
+            }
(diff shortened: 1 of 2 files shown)


warp10/src/main/java/io/warp10/continuum/gts/MetadataSelectorMatcher.java
```diff
@@ -278,21 +278,23 @@ public static MetadataIterator getIterator(final Iterator<Metadata> iter, Direct
...
-          if (entry.getValue().startsWith("=")) {
-            sb.append("=");
-            sb.append(WarpURLEncoder.encode(entry.getValue().substring(1), StandardCharsets.UTF_8));
-          } else if (entry.getValue().startsWith("~")) {
-            sb.append("~");
-            sb.append(WarpURLEncoder.encode(entry.getValue().substring(1), StandardCharsets.UTF_8));
-          } else {
-            sb.append("=");
-            sb.append(WarpURLEncoder.encode(entry.getValue(), StandardCharsets.UTF_8));
+          if (entry.getValue() != null) { // it could be null (.owner, for example)
```
hbs, line 281: null != entry.getValue()

Could you explain in more details when getValue could return null, it is not clear from the current comment

pi-r-p: I tried to understand the origin of .owner=null , but I cannot figure it out. 
- It is not in FIND.java line 742.
- There is a owner key in the read token I used to test.

pi-r-p:
bug fix in #1362 , this null check is not needed anymore.

PR #1363, opened 2024-06-06
add explicit error for FIND and FETCH execution within a plugin handler

pi-r-p:


warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -224,6 +221,10 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
 
     DirectoryClient directoryClient = stack.getDirectoryClient();
 
+    if (null == directoryClient) {
+      throw new WarpScriptException("There is no available backend. " + this.getName() + " might be executed from a plugin. To expose backend to plugins, set " + Configuration.EGRESS_CLIENTS_EXPOSE + " = true");
```
hbs, line 225: Put the function name at the beginning of the error message and do not address the user directly:

getName() + " Backend clients are not exposed unless configuration '" + Configuration.EGRESS_CLIENTS_EXPOSE +  "' is set to 'true'."

Ditto below

PR #1368, opened 2024-07-12
Added options to output GTS list as WarpScript

hbs:


pi-r-p:
A few questions about the functions:
- Is there a good reason not to build a list of GTS ? maxdepth can be reached easily.
```
[]
'xxx' UNWRAP +!
'yyy' UNWRAP +!
```
- if clearwarscript must be readable, {}= must not be escaped (just as standard FIND), and it should be a single line followed by PARSE (no SWAP nor +)

hbs:
The format is identical to the one produced in the `.info` file generated along HFiles, compatibility is key.

pi-r-p:
If so, and if needed, should it also have a fake count ?
sb.append(" # 0 0");

hbs:
Count is a comment, not ever parsed

pi-r-p (approved):

PR #1371, opened 2024-08-26
Fixed scoping when using the 'gts' parameter

hbs:


stggn (approved):

warp10/src/main/java/io/warp10/script/functions/FETCH.java
```diff
@@ -428,13 +433,62 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+            } else {
+              return modifiedMetas.get(idx++);
+            }
+          }
+
+          @Override
+          public void close() throws Exception {}
+        };
+
+        MetadataIterator mi = MetadataSelectorMatcher.getIterator(miter, dr);
```
pi-r-p, line 474: Null pointer exception here.
```
    {
      "type": "class java.lang.NullPointerException",
      "message": "Cannot invoke \"java.util.List.get(int)\" because the return value of \"io.warp10.continuum.store.thrift.data.DirectoryRequest.getLabelsSelectors()\" is null",
      "stacktrace": [
        [
          "MetadataSelectorMatcher.java",
          278,
          "io.warp10.continuum.gts.MetadataSelectorMatcher",
          "getIterator"
        ],
        [
          "FETCH.java",
          474,
          "io.warp10.script.functions.FETCH",
          "apply"
        ],
```

pi-r-p (changes requested):

hbs:
PTAL

pi-r-p (approved):

PR #1372, opened 2024-08-27
Added support for LISTs and SETs in SET-> and ->SET

hbs:


stggn (approved):

warp10/src/main/java/io/warp10/script/functions/SETTO.java
```diff
@@ -29,24 +29,26 @@
...
-      throw new WarpScriptException(getName() + " expects a set on top of the stack.");
+
+    if (top instanceof Set) {
+      List<Object> list = new ArrayList<Object>();
+
+      list.addAll((Set) top);
+
+      stack.push(list);
+    } else if (top instanceof List) {
+      stack.push(top);
```
stggn, line 48: If the list contains duplicates, they are not be removed.
This is perhaps a odd behavior for lists that pass through a SET function

hbs:
PTAL

stggn (approved):

PR #1383, opened 2024-12-02
Added support for specifying private key in multiple ways

hbs:


warp10/src/main/java/io/warp10/script/functions/ECPRIVATE.java
```diff
@@ -66,18 +66,7 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-
-    String dstr = (String) params.get(Constants.KEY_D);
-    final BigInteger d;
-
-    if (dstr.startsWith("0x") || dstr.startsWith("0X")) {
-      d = new BigInteger(dstr.substring(2), 16);
-    } else {
-      d = new BigInteger(dstr);
-    }
+    final BigInteger d = TOBD.toBigInteger(getName(), params.get(Constants.KEY_D));
```
stggn, line 69: With this function, keys beginning with "OX" are not be processed as before.
Is this intentional?

hbs: Not sure why this was ever supported. Dropping the support for that awkward syntax is confirmed.

hbs:
PTAL

warp10/src/main/java/io/warp10/script/functions/ECPRIVATE.java
pi-r-p, line 69: You do not test whether  params.get(Constants.KEY_D)  exists or not in the map. it will return an error that do not help:
`"ECPRIVATE" can only be applied to a BIGDECIMAL, STRING, BYTES, LONG or DOUBLE argument.`

Maybe replace getName() by getName() + " " + Constants.KEY_D + " parameter conversion "

pi-r-p (changes requested):

pi-r-p (approved):

PR #1387, opened 2025-01-02
Add foundationdb client busyness metric on egress

alexandrebrg:
:wave: 

This PR implements a metric that allow to understand how saturated are the network threads of the FoundationDB client, only on egresses. 

It's very convenient to have in order to have tune the number of network threads needed, scanners...

### DIFF
--- a/warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
+++ b/warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2024  SenX S.A.S.
+//   Copyright 2018-2025  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -56,6 +56,7 @@ import io.warp10.continuum.store.DirectoryClient;
 import io.warp10.continuum.store.StoreClient;
 import io.warp10.continuum.thrift.data.LoggingEvent;
 import io.warp10.crypto.KeyStore;
+import io.warp10.fdb.FDBStoreClient;
 import io.warp10.script.MemoryWarpScriptStack;
 import io.warp10.script.StackUtils;
 import io.warp10.script.WarpScriptException;
@@ -621,6 +622,15 @@ public class EgressExecHandler extends AbstractHandler {
       //
 
       Sensision.set(SensisionConstants.SENSISION_CLASS_WARPSCRIPT_JVM_FREEMEMORY, Sensision.EMPTY_LABELS, Runtime.getRuntime().freeMemory());
+
+      //
+      // Record FoundationDB network threads busyness
+      //
+
+      if (storeClient instanceof FDBStoreClient) {
+        FDBStoreClient fsc = (FDBStoreClient) storeClient;
+        fsc.getPool().updateDatabasesBusyness();
+      }
 
       LoggingEvent event = LogUtil.setLoggingEventAttribute(null, LogUtil.WARPSCRIPT_SCRIPT, scriptSB.toString());
 
--- a/warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
+++ b/warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
@@ -1,5 +1,5 @@
 //
-//   Copyright 2018-2023  SenX S.A.S.
+//   Copyright 2018-2025  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
```diff
@@ -869,6 +869,11 @@ public class SensisionConstants {
    */
   public static final String SENSISION_CLASS_WARPSCRIPT_JVM_FREEMEMORY = "warp.script.jvm.freememory";
 
+  /**
+   * Saturation of FoundationDB network threads
+   */
+  public static final String SENSISION_CLASS_WARPSCRIPT_FDB_BUSYNESS = "warp.script.fdb.busyness";
```
hbs, line 875: Please use `fdb.client.busyness` as the metric name and change the constant name to `SENSISION_CLASS_FDB_CLIENT_BUSYNESS`.

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
```diff
@@ -622,6 +624,16 @@ public void handle(String target, Request baseRequest, HttpServletRequest req, H
...
       Sensision.set(SensisionConstants.SENSISION_CLASS_WARPSCRIPT_JVM_FREEMEMORY, Sensision.EMPTY_LABELS, Runtime.getRuntime().freeMemory());
 
+      //
+      // Record FoundationDB current network thread busyness
+      //
+
+      if (storeClient instanceof FDBStoreClient) {
+        FDBStoreClient fsc = (FDBStoreClient) storeClient;
+        FDBPool pool = fsc.getPool();
+        Sensision.set(SensisionConstants.SENSISION_CLASS_WARPSCRIPT_FDB_BUSYNESS, Sensision.EMPTY_LABELS, pool.getDatabase().getMainThreadBusyness());
```
hbs, line 634: The call to getDatabase returns an instance of Database from the pool, so you will not get a global view of busyness, only for a single Database instance among the pool. This is probably not sufficient, it would be more beneficial to add a method to FDBPool to report busyness for all pooled Database instances with a `dbinstance` label with a number from 0 to the pool size - 1.

hbs (changes requested):

warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
alexandrebrg: Metric name updated

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
alexandrebrg: Indeed, I added a method into FDBPool and use labels to give data about each thread. In case you are wondering if load on this `libfdb` call can slow down the thread itself, it should'nt, the value is updated by `libfdb` itself every second. We're just reading its value.

warp10/src/main/java/io/warp10/fdb/FDBPool.java
```diff
@@ -60,6 +60,14 @@ public synchronized Database getDatabase() {
     return db;
   }
 
+  public double[] getDatabasesBusyness() {
+      double[] busyness = new double[databases.length];
```
hbs, line 64: Instead of returning a newly allocated array, it would be better to modify the method so it updates the Sensision metrics. In order to speed things up (the call to Sensision.set being synchronized), you could keep track of the last update and only update the metrics if the last update was more than 1s ago, since the underlying FDB metrics are not updated more frequently than that it would be sufficient to have an always up to date value.

alexandrebrg: Updated accordingly.

warp10/src/main/java/io/warp10/continuum/sensision/SensisionConstants.java
hbs, line 1: Update copyright to 2018-2025

warp10/src/main/java/io/warp10/fdb/FDBPool.java
hbs, line 1: Update copyright to 2022-2025

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
hbs, line 1: Update copyright to 2018-2025

hbs (changes requested):
Just a few minor updates still needed

warp10/src/main/java/io/warp10/continuum/egress/EgressExecHandler.java
alexandrebrg: Updated all three files

hbs:
Thanks for the PR.

PR #1389, opened 2025-01-09
KVStore support

hbs:
This PR adds a Key/Value store feature to Warp 10, enabling the storage of arbitrary Key/Value pairs for the following storage engines:

- LevelDB
- in memory
- FoundationDB for both standalone+ and distributed versions

Additionally, stored K/V pairs can be replicated via the Datalog mechanism.

WarpLib functions `KVSTORE` and `KVLOAD` allow to store and retrieved stored K/V pairs.

Retrieval is either range or key based, allowing to scan the key space or access specific keys explicitely.

Tenant prefixes are supported for FoundationDB based versions.

Optional custom prefix can be added to keys prior to storage to allow for user segregation by key spaces.

### DIFF
--- a/etc/conf.templates/standalone/00-warp.conf.template
+++ b/etc/conf.templates/standalone/00-warp.conf.template
@@ -1,5 +1,5 @@
 //
-//   Copyright 2019-2023  SenX S.A.S.
+//   Copyright 2019-2025  SenX S.A.S.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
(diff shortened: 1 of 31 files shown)


pi-r-p:
Is it possible to support cast of key to string in KVLOAD ?

Usage :
```
{  'key' $bytes } $wt KVSTORE
{
  'token' $rt
  'keys' [ 'key' ]
  'keycast' true           // default to false ?
} KVLOAD 'key' GET 
```

pi-r-p:
BYTES as MAP keys are not usable (at all). Here is a test:

```
{
  'drink' 'brawndo'
} $wt KVSTORE 


{
  'token' $rt
  'keys' [  'drink' ]
}
KVLOAD 'r' STORE

$r
<% 
  [ 'k' 'v' ] STORE 
  $k 'ascii' BYTES->
  $v 'ascii' BYTES->
%> F MMAP    /// ok, great, see the drink key

$r 'drink' 'ascii' ->BYTES  GET   // null


```

hbs:
PTAL

pi-r-p (approved):

hbs:
Fixed config files. PTAL

pi-r-p (approved):

PR #1391, opened 2025-01-23
Fixed feeder ID retrieval

hbs:


pi-r-p (approved):
Cannot test. 
Merges correctly with  #1392

PR #1392, opened 2025-01-24
Added keepalive messages between feeders and consumers

hbs:


warp10/src/main/java/io/warp10/standalone/datalog/TCPDatalogConsumer.java
```diff
@@ -127,6 +129,10 @@ public class TCPDatalogConsumer extends Thread implements DatalogConsumer {
 
   private String suffix;
 
+  static {
+    KEEPALIVE_DELAY = Long.parseLong(WarpConfig.getProperty(FileBasedDatalogManager.CONFIG_DATALOG_CONSUMER_KEEPALIVE, TCPDatalogFeederWorker.DEFAULT_KEEPALIVE));
+  }
+
```
pi-r-p, line 135: Can't this class be instantiated before the configuration is set ?  
Remember such a problem in the past...

warp10/src/main/java/io/warp10/standalone/datalog/TCPDatalogConsumer.java
```diff
@@ -480,6 +515,7 @@ public void run() {
                 DatalogHelper.writeLong(out, bytes.length, 4);
                 out.write(bytes);
                 out.flush();
+                lastMessage = System.currentTimeMillis();
```
pi-r-p, line 518: this one can be removed

pi-r-p (commented):
Cannot really test it.

warp10/src/main/java/io/warp10/standalone/datalog/TCPDatalogConsumer.java
hbs: Nope, consumers are created because they are defined in the configuration, so config is always set when this static block is called.

hbs:
PTAL

PR #1395, opened 2025-02-11
Selective MSTORE

pi-r-p:
Yes, it is a breaking change.
But this signature was never ever documented.

When combined with MCSTORE, this allows a easy way to store parameters, then default parameters :
```
<%
  [ 'a' 'b' ] MSTORE
  {
    'a' 0
    'b' 0
  } MCSTORE
  $a $b +
%> 'm' STORE

{ 'b' 4 } @m
```


pi-r-p:
-> need @randomboolean review, he is the only one that may use this signature.

randomboolean:
I agree, it looks more like a bugfix and it is coherent with LSTORE that does not store null values into variable names in excess of the first list size

randomboolean (approved):

PR #1396, opened 2025-02-13
Modified duplicate ticks handling to improve performance

hbs:


randomboolean:
The ticks parameter is not expected to be sorted. Also, this optimization should be a fonction called also by the List signature (applyFromList)

hbs:
> The ticks parameter is not expected to be sorted. Also, this optimization should be a function called also by the List signature (applyFromList)

Are you suggesting the fix doesn't work as is?

pi-r-p:
This PR is included in #1397 , so it will be easier to close it and move the discussion to #1397 .

PR #1397, opened 2025-02-14
Speed fill

pi-r-p:


warp10/src/main/java/io/warp10/script/WarpScriptSingleValueFillerFunction.java
```diff
@@ -23,10 +23,10 @@ public interface WarpScriptSingleValueFillerFunction {
...
     public WarpScriptSingleValueFillerFunction compute(GeoTimeSerie gts) throws WarpScriptException;
 
-    default public Object evaluate(long tick) throws WarpScriptException {
+    default public void fillTick(long tick, GeoTimeSerie gts, Object invalidValue) throws WarpScriptException {
       throw new WarpScriptException("Invalid Filler Error: evaluator function has not been precomputed yet.");
     }
   }
 
-  public Object evaluate(long tick) throws WarpScriptException;
+  public void fillTick(long tick, GeoTimeSerie gts, Object invalidValue) throws WarpScriptException;
```
hbs, line 31: This is a breaking change since you are radically changing the interface. Instead mark evaluate as deprecated, add default implementations to evaluate and fillTick which throw an exception and you should be good to go except for backward compatibility across 3.x of extensions making use of WarpScriptSingleValueFillerFunction as defined in this PR.

warp10/src/main/java/io/warp10/script/filler/FillerNext.java
```diff
@@ -62,4 +65,31 @@ public int getPostWindow() {
   public int getPreWindow() {
     return 0;
   }
+
+  @Override
+  public WarpScriptSingleValueFillerFunction compute(GeoTimeSerie gts) throws WarpScriptException {
+
+    // fill always sort input before.
```
hbs, line 72: You cannot assume a single call path for a public method, so call gts.sort() which will return immediately if the GTS is already sorted.

warp10/src/main/java/io/warp10/script/filler/FillerNext.java
```diff
@@ -62,4 +65,31 @@ public int getPostWindow() {
...
+    final int nTicks = gts.size();
+    final GeoTimeSerie original = gts;
+    final long nextTick = GTSHelper.tickAtIndex(original, 0);
```
hbs, line 75: This line and the following will fail if the GTS is empty

warp10/src/main/java/io/warp10/script/filler/FillerNext.java
```diff
@@ -62,4 +65,31 @@ public int getPostWindow() {
...
+    final long lastTick = GTSHelper.tickAtIndex(original, nTicks - 1);
+    final int[] currentIndex = {0};
```
hbs, line 77: Why not make currentIndex a field of the instance you create below instead? This would make the code way easier to understand.

warp10/src/main/java/io/warp10/script/filler/FillerPrevious.java
```diff
@@ -63,4 +66,33 @@ public int getPreWindow() {
     return 1;
   }
 
+  @Override
+  public WarpScriptSingleValueFillerFunction compute(GeoTimeSerie gts) throws WarpScriptException {
+
+    // fill always sort input before.
```
hbs, line 72: Ditto, cf ante

hbs (changes requested):

warp10/src/main/java/io/warp10/script/filler/FillerNext.java
pi-r-p: no, it will return long min value.

pi-r-p:
PTAL

pi-r-p:
removed the last binary search in the ticks loop. 
PTAL

pi-r-p:
- filler.interpolate is now as fast as filler.linear (and supports GEO).
- there was a bug on 3.4.1 with empty GTS (fixed)

PTAL

warp10/src/main/java/io/warp10/script/filler/FillerInterpolate.java
```diff
@@ -109,4 +115,111 @@ public int getPreWindow() {
...
+    final PolynomialSplineFunction elevationFunction;
+
+    if (gts.size() == 0) {
+      valFunction = null;
+      latFunction = null;
+      lonFunction = null;
+      elevationFunction = null;
+    } else {
+      if (GeoTimeSerie.TYPE.DOUBLE != gts.getType() && GeoTimeSerie.TYPE.LONG != gts.getType() && gts.size() > 0) {
+        throw new WarpScriptException(getName() + " expects a GTS of type DOUBLE or LONG, but instead got a GTS of type " + gts.getType().name());
```
hbs, line 133: The message will be weird if an empty GTS whose type was forced is passed as parameter.

warp10/src/main/java/io/warp10/script/filler/FillerInterpolate.java
```diff
@@ -109,4 +115,111 @@ public int getPreWindow() {
...
+        latFunction = null;
+        lonFunction = null;
+      }
+
+
+      // elevations
+      // how many ticks do have an elevation ?
+      int nElev = 0;
+      for (int i = 0; i < gts.size(); i++) {
+        if (GeoTimeSerie.NO_LOCATION != GTSHelper.elevationAtIndex(gts, i)) {
```
hbs, line 182: NO_ELEVATION

warp10/src/main/java/io/warp10/script/filler/FillerInterpolate.java
```diff
@@ -109,4 +115,111 @@ public int getPreWindow() {
...
+          nElev++;
+        }
+      }
+      if (nElev > 0) {
+        double[] xElevTicks = new double[nElev];
+        double[] fElevVal = new double[nElev];
+        int idx = 0;
+        for (int i = 0; i < gts.size(); i++) {
+          long e = GTSHelper.elevationAtIndex(gts, i);
+          if (e != GeoTimeSerie.NO_LOCATION) {
```
hbs, line 192: NO_ELEVATION

warp10/src/main/java/io/warp10/script/filler/FillerLinear.java
```diff
@@ -31,28 +31,33 @@ public FillerLinear(String name) {
...
-    int size = gts.size();
     final PolynomialSplineFunction function;
-    if (size > 1) {
-      function = (new LinearInterpolator()).interpolate(xval, fval);
-    } else {
+    if (gts.size()==0) {
       function = null;
+    } else {
+      if (GeoTimeSerie.TYPE.DOUBLE != gts.getType() && GeoTimeSerie.TYPE.LONG != gts.getType() && gts.size() > 0) {
+        throw new WarpScriptException(getName() + " expects a GTS of type DOUBLE or LONG, but instead got a GTS of type " + gts.getType().name());
```
hbs, line 39: ditto above

warp10/src/main/java/io/warp10/script/filler/FillerLowess.java
```diff
@@ -78,33 +78,39 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-      double bandwidthRatio = Math.min(1.0, (double) bandwidth / size + 1e-12);
-      function = (new LoessInterpolator(bandwidthRatio,0, accuracy)).interpolate(xval, fval);
-    } else if (size > 1) {
-      function = (new LinearInterpolator()).interpolate(xval, fval);
-    } else {
+    if (gts.size() == 0) {
       function = null;
+    } else {
+      if (GeoTimeSerie.TYPE.DOUBLE != gts.getType() && GeoTimeSerie.TYPE.LONG != gts.getType()) {
+        throw new WarpScriptException(getName() + " expects a GTS of type DOUBLE or LONG, but instead got a GTS of type " + gts.getType().name());
```
hbs, line 86: ditto

warp10/src/main/java/io/warp10/script/filler/FillerNewton.java
```diff
@@ -34,30 +34,35 @@ public FillerNewton(String name) {
...
-    int size = gts.size();
     final PolynomialFunctionNewtonForm function;
-    if (size > 1) {
-      function = (new DividedDifferenceInterpolator()).interpolate(xval, fval);
-    } else {
+    if (gts.size() == 0) {
       function = null;
+    } else {
+      if (GeoTimeSerie.TYPE.DOUBLE != gts.getType() && GeoTimeSerie.TYPE.LONG != gts.getType()) {
+        throw new WarpScriptException(getName() + " expects a GTS of type DOUBLE or LONG, but instead got a GTS of type " + gts.getType().name());
```
hbs, line 42: ditto

warp10/src/main/java/io/warp10/script/filler/FillerNext.java
```diff
@@ -62,4 +65,30 @@ public int getPostWindow() {
...
+    GTSHelper.sort(gts);
+    final int nTicks = gts.size();
+    final GeoTimeSerie original = gts;
+    final long lastTick = GTSHelper.tickAtIndex(original, nTicks - 1);
+
+    return new WarpScriptSingleValueFillerFunction() {
+      private int currentIndex = 0;
+      @Override
+      public void fillTick(long tick, GeoTimeSerie gts, Object invalidValue) throws WarpScriptException {
+        if (nTicks == 0 || tick >= lastTick) {
```
hbs, line 81: 0 == nTicks

warp10/src/main/java/io/warp10/script/filler/FillerPrevious.java
```diff
@@ -63,4 +66,34 @@ public int getPreWindow() {
...
+    final GeoTimeSerie original = gts;
+    final long firstTick = GTSHelper.tickAtIndex(original, 0);
+
+    return new WarpScriptSingleValueFillerFunction() {
+      private int currentIndex = 0;
+
+      @Override
+      public void fillTick(long tick, GeoTimeSerie gts, Object invalidValue) throws WarpScriptException {
+
+        if (nTicks == 0 || tick <= firstTick) {
```
hbs, line 83: 0 == nTicks

warp10/src/main/java/io/warp10/script/filler/FillerPrevious.java
```diff
@@ -63,4 +66,34 @@ public int getPreWindow() {
...
+          if (null != invalidValue && tick != firstTick) {
+            GTSHelper.setValue(gts, tick, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, invalidValue, false);
+          }
+          return;
+        }
+        while (tick > GTSHelper.tickAtIndex(original, currentIndex) && currentIndex < nTicks) {
+          currentIndex++;
+        }
+        int previousIndex = currentIndex == 0 ? 0 : currentIndex - 1;
```
hbs, line 92: 0 == currentIndex

warp10/src/main/java/io/warp10/script/filler/FillerRlowess.java
```diff
@@ -90,33 +90,37 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-      function = (new LoessInterpolator(bandwidthRatio,robustness, accuracy)).interpolate(xval, fval);
-    } else if (size > 1) {
-      function = (new LinearInterpolator()).interpolate(xval, fval);
-    } else {
+    if (gts.size() == 0) {
       function = null;
-    }
+    } else {
+      if (GeoTimeSerie.TYPE.DOUBLE != gts.getType() && GeoTimeSerie.TYPE.LONG != gts.getType()) {
+        throw new WarpScriptException(getName() + " expects a GTS of type DOUBLE or LONG, but instead got a GTS of type " + gts.getType().name());
```
hbs, line 98: ditto

hbs (changes requested):

warp10/src/main/java/io/warp10/script/filler/FillerInterpolate.java
pi-r-p: The empty gts is managed a few lines earlier, I remove the condition.

pi-r-p:
fixed linked to https://github.com/senx/warp10-platform/pull/1396#issuecomment-2669128164

warp10/src/main/java/io/warp10/script/WarpScriptSingleValueFillerFunction.java
randomboolean: I believe fillTick should default to evaluate instead of throwing an error so existing extensions are not broken.

warp10/src/main/java/io/warp10/script/WarpScriptSingleValueFillerFunction.java
```diff
@@ -23,10 +23,20 @@ public interface WarpScriptSingleValueFillerFunction {
...
     }
+
+    @Deprecated
+    default public Object evaluate(long tick) throws WarpScriptException {
+      throw new WarpScriptException("Invalid Filler Definition Error: evaluate function is deprecated.");
+    }
   }
 
-  public Object evaluate(long tick) throws WarpScriptException;
+  public void fillTick(long tick, GeoTimeSerie gts, Object invalidValue) throws WarpScriptException;
```
randomboolean, line 36: Do you mean to modify the gts inplace ?
This is also a breaking change

randomboolean (changes requested):
Tests return errors

`[ NEWGTS [ 10  1.0 ] ADDVALUE [ 30  2.0 ] ADDVALUE filler.akima [ 20 ] ]  FILL`
with this PR : =>FILL<=' in section [TOP] (/ by zero)
previously : [[{"c":"","l":{},"a":{},"la":0,"v":[[10,1.0],[30,2.0],[20,1.5]]}]]

warp10/src/main/java/io/warp10/script/WarpScriptSingleValueFillerFunction.java
pi-r-p: Existing extensions, if any, implements evaluate(), so the default method will never be executed, will it ?

randomboolean: Existing extensions implements evaluate(), but your changes to fill() method in GTSHelper makes them use their default fillTick(), which then should default to their implementation of evaluate in order not to break their actual behaviour

warp10/src/main/java/io/warp10/script/WarpScriptSingleValueFillerFunction.java
randomboolean: Nvm, you pass a clone of the input gts when you use fillTick in fill.
But then I dont understand the point. Why do you need fillTick rather than evaluate ?
Why implementing the new precomputable interface with evaluate isn't enough to make older filler faster ?

pi-r-p:
> Tests return errors
> 
> `[ NEWGTS [ 10 1.0 ] ADDVALUE [ 30 2.0 ] ADDVALUE filler.akima [ 20 ] ] FILL` with this PR : =>FILL<=' in section [TOP] (/ by zero) previously : [[{"c":"","l":{},"a":{},"la":0,"v":[[10,1.0],[30,2.0],[20,1.5]]}]]

Indeed, Same for empty ticks list with the previous signature too.
Fixed.

warp10/src/main/java/io/warp10/script/WarpScriptSingleValueFillerFunction.java
pi-r-p: > Do you mean to modify the gts inplace ? This is also a breaking change

Well, no... gts is cloned very early in the process. Please provide an example of malfunction.

pi-r-p: I kept evaluate for retro compatibility, and for filler.trend (the last slow one, and broken imho).

Main reasons:
- support geo
- the gts argument allow to call setvalue directly in the filler, instead of returning values to the caller (including geo).

pi-r-p:
Question is: how many extensions contains a filler ?
To be retrocompatible, fillTick can have a default implementation "evaluate then setvalue".

It depends if we accept a breaking change on these new single value fillers or not... To discuss.

randomboolean:
> Question is: how many extensions contains a filler ? To be retrocompatible, fillTick can have a default implementation "evaluate then setvalue".
> 
> It depends if we accept a breaking change on these new single value fillers or not... To discuss.

Also, you need not change the implementation of the new fillers with the new interface, since their would have inherited a fillTick which default to their evaluate (+ setValue).

pi-r-p:
@randomboolean, PTAL. This is a bit weird now, the interface has default methods everywhere, but it should work.

warp10/src/main/java/io/warp10/script/filler/FillerInterpolate.java
```diff
@@ -109,4 +115,111 @@ public int getPreWindow() {
...
+          long l = GTSHelper.locationAtIndex(gts, i);
+          if (l != GeoTimeSerie.NO_LOCATION) {
+            double[] latlon = GeoXPLib.fromGeoXPPoint(l);
+            xLocTicks[idx] = GTSHelper.tickAtIndex(gts, i);
+            fLatVal[idx] = latlon[0];
+            fLonVal[idx] = latlon[1];
+            idx++;
+          }
+        }
+        latFunction = (new LinearInterpolator()).interpolate(xLocTicks, fLatVal);
```
randomboolean, line 170: This is a breaking change. Old filler.interpolate does not handle geo filling based on the previous and next existing geo coordinate.
See the apply method of the same class used in cross FILL.
If the previous data point does not have a geo coordinate, then the actual data point would not get a geo coordinate. In your implementation, it can if the datapoint before the previous has geo coordinates.

randomboolean (changes requested):

warp10/src/main/java/io/warp10/script/filler/FillerInterpolate.java
pi-r-p: Indeed. You will have more datapoints filled with geo information than before with this new implementation.

pi-r-p: As discussed IRL, if a customer find this edge case with filler.interpolate, it can switch to the map signature, and add `'force.old.filler' true`  flag. Execution time is then multiplied by 10.

warp10/src/main/java/io/warp10/script/filler/FillerLinear.java
```diff
@@ -31,28 +31,33 @@ public FillerLinear(String name) {
...
+      double[] fval = GTSHelper.getValuesAsDouble(gts);
+
+      int size = gts.size();
+      if (size > 1) {
+        function = (new LinearInterpolator()).interpolate(xval, fval);
+      } else {
+        function = null;
+      }
     }
+
```
randomboolean, line 52: remove one blank line of the two

warp10/src/main/java/io/warp10/script/filler/FillerLowess.java
```diff
@@ -78,33 +78,39 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
+      if (size > 2) {
+        double bandwidthRatio = Math.min(1.0, (double) bandwidth / size + 1e-12);
+        function = (new LoessInterpolator(bandwidthRatio, 0, accuracy)).interpolate(xval, fval);
+      } else if (size > 1) {
+        function = (new LinearInterpolator()).interpolate(xval, fval);
+      } else {
+        function = null;
+      }
     }
+
```
randomboolean, line 102: remove one blank line of the two

warp10/src/main/java/io/warp10/script/filler/FillerLowess.java
```diff
@@ -78,33 +78,39 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-        } else {
-          return function.value(tick);
+      public void fillTick(long tick, GeoTimeSerie filled, Object invalidValue) throws WarpScriptException {
+        if (null != function && function.isValidPoint(tick)) {
+          GTSHelper.setValue(filled, tick, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, function.value(tick), false);
+        } else if (null != invalidValue) {
+          GTSHelper.setValue(filled, tick, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, invalidValue, false);
         }
       }
+
```
randomboolean, line 113: remove unecessary blank line

warp10/src/main/java/io/warp10/script/filler/FillerNewton.java
```diff
@@ -34,30 +34,35 @@ public FillerNewton(String name) {
...
-        } else {
-          return function.value(tick);
+      public void fillTick(long tick, GeoTimeSerie filled, Object invalidValue) throws WarpScriptException {
+        if (null != function) {
+          GTSHelper.setValue(filled, tick, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, function.value(tick), false);
+        } else if (null != invalidValue) {
+          GTSHelper.setValue(filled, tick, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, invalidValue, false);
         }
       }
+
```
randomboolean, line 65: unnecessary blank line

warp10/src/main/java/io/warp10/script/filler/FillerRlowess.java
```diff
@@ -90,33 +90,37 @@ public Object apply(WarpScriptStack stack) throws WarpScriptException {
...
-        } else {
-          return function.value(tick);
+      public void fillTick(long tick, GeoTimeSerie filled, Object invalidValue) throws WarpScriptException {
+        if (null != function && function.isValidPoint(tick)) {
+          GTSHelper.setValue(filled, tick, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, function.value(tick), false);
+        } else if (null != invalidValue) {
+          GTSHelper.setValue(filled, tick, GeoTimeSerie.NO_LOCATION, GeoTimeSerie.NO_ELEVATION, invalidValue, false);
         }
       }
+
```
randomboolean, line 123: unnecessary blank line

randomboolean (approved):
Also, a few unnecessary blank lines can be removed

PR #1398, opened 2025-02-19
Macro repo unit tests issue

pi-r-p:
When wf/macroA calls wf/macroB in its unit tests, macroB cannot be found, and macroA cannot be loaded.


hbs:
The implications of that small change need to be thoroughly studied as they may have security impacts which MUST be fully understood prior to merging.

hbs:
Closing this PR as it breaks the determinism of macro loading.

PR #1402, opened 2025-08-25
Update fdb-java dependency version to 7.3.68

alexandrebrg:
:wave: 

This PR allows to define FoundationDB `api_version` to `730`. The `fdb-java` dep is backward compatible, meaning `710` will still be a valid api version.


pi-r-p:
@alexandrebrg  : can we update to an even more recent fdb-java ?

PR #1407, opened 2026-02-13
Fixed directory walking to ensure we are robust to errors such as dis…

hbs:
…appearing files. Reverted previous broken fix.


pi-r-p (changes requested):
This new approach do not follow symbolic links anymore

pi-r-p (approved):

